Issue #5, opened 2017-08-16
Mongodb connections are not securely persisted across workspaces


EricJizbaMSFT:
Rather than using a workspace specific storage path to store connection strings, we should use the vscode secure storage that works across workspaces

EricJizbaMSFT:
@fiveisprime @chrisdias Matt and I discussed no longer persisting connection strings at all. The Cosmos DB extension will list connections from Azure and we don't want to persist connection strings unless we can store them securely. If you confirm this behavior, I'll change this issue to something like "Don't persist connection strings"

Ideally we have secure storage, but per offline discussion we are not sure if/when that will be publicly available.

fiveisprime:
Let's just leave this for now. We can revisit this if users have a need to persist raw connection strings.. I have a feeling this will come back up.

chrisdias:
@fiveisprime not clear what you mean by "Let's just leave this for now"... what is "this"? :)

The challenge here is that it would be good to let people move between their non CosmosDB connections and their CosmosDB connections.

Can we simply continue to what we do with connection strings today (persist per workspace) and revisit the issue once we have the secure storage API?  

Or, maybe that is what @fiveisprime is suggesting? :)

fiveisprime:
I don't like the idea of storing these in plain text since there may be credentials in the connection strings.

"Leave this for now" was meant as: avoid persisting potentially sensitive information until we have a secure solution or until we find that we have a need and the security concern is invalid (or acceptable by the community's standards).

Sorry that wasn't super clear, Eric asked me for the same clarification 🙍

Issue #3, opened 2017-08-16
MongoDB connection is not persisted


EricJizbaMSFT:
We should persist the latest connected MongoDB instance so that the user doesn't have to click 'connect' when they re-open VS Code

PR #6, opened 2017-08-17
Refactor persistence of connection strings

### LINKED ISSUE #2: Mongo extension is not activated if VS Code is opened without a workspace
Repro steps:
1. Open VS Code without a default workspace (if necessary, open VS Code, go to 'File -> Close Folder', close Code, and re-open)
2. Click the mongo 'New Scrapbook' button

Expected: A new scrapbook opens
Actual: 'Error command mongo.newScrapbook not found'

This error occurs because we depend on ExtensionContext.storagePath, which is undefined if no workspace is open.

EricJizbaMSFT:
Rather than persisting connection strings in normal storage on a per workspace basis, we will now persist strings in secure storage globally for VS Code.

We will also persist the latest connected DB and reconnect if possible when VS Code opens.

Fixes #2 #3 #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 
-# MongoDB NoSQL Support for VS Code
+# Cosmos DB NoSQL Support for VS Code
 
-The MongoDB extension makes it easy to work with MongoDB NoSQL databases, collections, and documents. With this extension, you can:
+The Cosmos DB extension makes it easy to work with MongoDB NoSQL databases, collections, and documents. With this extension, you can:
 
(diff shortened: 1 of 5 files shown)


EricJizbaMSFT:
@whiskeyjay PTAL - it seems I can't add you as a reviewer yet :/

EricJizbaMSFT:
Closing this PR for now. The plans around secure storage are in flux and in discussion offline it sounds like we might want to err on the side of caution and just not persist these connection strings at all for now.

Issue #15, opened 2017-08-17
Display Cosmos DB instances in VS Code (specifically MongoDB)


EricJizbaMSFT:
This is the generic issue for displaying MongoDB instances from Azure. This issue only covers instances that use MongoDB API.

We plan to leverage the AzureLogin extension and let that extension manage subscriptions/resource groups. We will display the resource name and resource group instead of the connection string.

We will also have to distinguish Azure connections from connections the user adds. (Because the actions will be different and the display name will have to be different)

Issue #10, opened 2017-08-17
Azure instances in the Cosmos DB Explorer should not have 'Remove Server' action


EricJizbaMSFT:
This action is used to remove a connection string from the explorer, but that doesn't make sense when we load the server info from Azure. It is also confusing because it sounds like we're deleting the Cosmos DB instance

PR #17, opened 2017-08-21
List Azure Resources in Mongo Explorer

EricJizbaMSFT:
For now, this _only_ lists Mongo resources, and no other CosmosDB resources. I also removed the persistence of manually added connection strings because they were not using secure storage. I will add that support back in a future commit using secure storage.

Related issues: #10 #15

### DIFF
--- a/src/mongo/mongo.ts
+++ b/src/mongo/mongo.ts
@@ -120,7 +120,10 @@ export class Model implements IMongoResource {
 			servers = servers.concat(...serverResult);
 		}));
 
-		return servers;
+		return servers.sort((a, b) => {
+			const n = a.resourceGroupName.localeCompare(b.resourceGroupName);
+			return n !== 0 ? n : a.name.localeCompare(b.name);
+		});
 	}
 }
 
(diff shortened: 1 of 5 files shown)


src/mongo/mongo.ts
```diff
@@ -30,129 +33,138 @@ export interface IMongoResource extends vscode.TreeItem {
...
-}
 
-export class NoConnectionServer implements IMongoResource {
+	private async getAzureMongoResources(): Promise<IMongoResource[]> {
+		let servers: Server[] = [];
+
+		await Promise.all(this.azureAccount.filters.map(async (filter) => {
+			const docDBClient = new DocumentdbManagementClient(filter.session.credentials, filter.subscription.subscriptionId);
+
+			const serverResult = await Promise.all(filter.resourceGroups.map(async group => {
```
whiskeyjay: If no resource group level filter is applied, would it be faster to list all instead of sending multiple requests by RGs?

src/mongo/mongo.ts
```diff
@@ -30,129 +33,138 @@ export interface IMongoResource extends vscode.TreeItem {
...
+				const dbs = (await docDBClient.databaseAccounts.listByResourceGroup(group.name)).filter(db => db.kind === "MongoDB");
+				return Promise.all(dbs.map(async db => {
```
whiskeyjay: Are the results sorted on UI? By name or by RG?

src/mongo/mongo.ts
EricJizbaMSFT: We would have to do extensive testing to find out. On one hand it might be more requests this way, but it can also do more asynchronously.

In addition, the 'DatabaseAccount' objects returned by the 'list' operation do not have the resource group name as a part of the metadata. It's almost easier to do it this way so that we know the group name associated with each resource.

src/mongo/mongo.ts
EricJizbaMSFT: Modified to use the same sorting as the appservice extension

PR #20, opened 2017-08-22
Persist attached mongo connection strings to secure storage

EricJizbaMSFT:
The location of the secure storage api might be moving around a little bit going forward, but the API itself shouldn't change much.

NOTE: We will not persist connection strings for resources from Azure. Those will be loaded on the fly each time. 

Related to #5

### DIFF
--- a/src/nodes.ts
+++ b/src/nodes.ts
@@ -114,9 +114,14 @@ export class AttachedServersNode implements INode {
 	readonly id: string = 'cosmosDBAttachedServers';
 	readonly label: string = 'Attached Mongo Servers';
 
+	private readonly _serviceName = "ms-azuretools.vscode-cosmosdb.connectionStrings";
 	private _attachedServers: INode[] = [];
 
 	readonly collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+
+	constructor(private readonly _azureAccount: AzureAccount, private readonly _globalState: vscode.Memento) {
+		this.loadPersistedServers();
+	}
 
 	get iconPath(): any {
 		return {
@@ -129,7 +134,7 @@ export class AttachedServersNode implements INode {
 		return this._attachedServers.length > 0 ? this._attachedServers : [new AttachMongoServerNode()];
 	}
 
-	async attach(connectionString: string): Promise<INode> {
+	async attach(connectionString: string, isInitializing: boolean = false): Promise<INode> {
 		try {
 			const db = await MongoClient.connect(connectionString);
 			const node = new MongoServerNode(connectionString, db.serverConfig);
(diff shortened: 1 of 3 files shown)


src/mongo/mongo.ts
```diff
@@ -125,6 +141,11 @@ export class Model implements IMongoResource {
 			return n !== 0 ? n : a.name.localeCompare(b.name);
 		});
 	}
+
+	private async persistServers() {
+		const secret = JSON.stringify(this._servers.map(s => s.connectionString));
+		await this.azureAccount.credentials.writeSecret(this._credentialsService, this._credentialsAccount, secret);
```
whiskeyjay: I know that Keytar uses [CredWrite](https://msdn.microsoft.com/en-us/library/windows/desktop/aa375187.aspx) function on Windows to store the secret, the function limits the size of the secret to 512 bytes. What if you have many connection strings so the generated JSON size exceeds the limit?

src/mongo/mongo.ts
```diff
@@ -40,18 +40,28 @@ export class Model implements IMongoResource {
...
 	
 	private _onChange: EventEmitter<void> = new EventEmitter<void>();
 	readonly onChange: Event<void> = this._onChange.event;
 
+	private readonly _credentialsService = "ms-azuretools.vscode-cosmosdb";
+	private readonly _credentialsAccount = "mongoConnectionStrings";
+
 	constructor(private azureAccount: AzureAccount) {
+		this.azureAccount.credentials.readSecret(this._credentialsService, this._credentialsAccount).then(async value => {
+			const connectionStrings: string[] = value ? JSON.parse(value) : null;
```
whiskeyjay: What happens if _value_ is not a valid JSON string? Although unlikely, the secret can be manually modified from the Control Panel on Windows (not sure about other platforms) or by any other apps using the same underlying Win32 functions.

src/mongo/mongo.ts
whiskeyjay: We should probably communicate with vscode about this.

src/mongo/mongo.ts
EricJizbaMSFT: Wrapped in a try/catch with an error message

src/mongo/mongo.ts
EricJizbaMSFT: Refactored so that we only store one connection string per secret. I tested this on Windows and it was working fine. If you're worried about longer connection strings (not sure if that's possible) I'd prefer to deal with that in a separate issue. I think this will solve it for most customers.

src/nodes.ts
```diff
@@ -129,27 +134,55 @@ export class AttachedServersNode implements INode {
...
+	async attach(connectionString: string, isInitializing: boolean = false): Promise<INode> {
 		try {
 			const db = await MongoClient.connect(connectionString);
 			const node = new MongoServerNode(connectionString, db.serverConfig);
 			if (this._attachedServers.find(s => s.id === node.id)) {
 				vscode.window.showWarningMessage(`Mongo server '${node.id}' is already attached.`)
 			} else {
 				this._attachedServers.push(node);
+				if (!isInitializing) {
+					await this._azureAccount.credentials.writeSecret(this._connectionStringsId, node.id, connectionString);
```
whiskeyjay: "_serviceName" or "_appName" makes more sense than _connectionStringsId, given the first parameter is called "service".

PR #26, opened 2017-08-24
Ignore resource group filters

EricJizbaMSFT:
We only want to filter based on subscription. Also change the label when no resources are found. It's weird to say 'Add a _filter_' to see more resources

#23


EricJizbaMSFT:
Filter might not be the best word, but the 'Azure Account' extension uses filter and we want to be consistent

Issue #7, opened 2017-08-17
Support non-MongoDB instances of CosmosDB


EricJizbaMSFT:
General issue to discuss how we want to handle the other supported APIs in CosmosDB (as of today - DocumentDB, Graph, and Table).

Do we display them with basic actions like 'Open in Portal'? Do we ignore them?

EricJizbaMSFT:
Closing this as we now have support for creating non-Mongo instances and displaying them in the explorer (with just the 'Open in Portal' action). We can open new issues for any more functionality

PR #27, opened 2017-08-24
Split Explorer code into 'Cosmos DB' and 'MongoDB'

EricJizbaMSFT:
I split this PR up into a few commits to make it easier to view. The first two commits are just moving files and copying code from one file to another. If you just want to see the actual code changes, look at the last commit: https://github.com/Microsoft/vscode-cosmosdb/commit/f68e5003e4999e624a98a564dd58fa6efc79dfa8

This is part of the work to enable non-MongoDB resources as in #7. It has no functional change and the only difference the user will notice is a few renames:
1. I modified the command categories to be either Cosmos DB or MongoDB (if the command entry point is in the top level explorer for Cosmos DB, then it has 'Mongo' in the name and Cosmos DB as the category. If the command entry point is mongo specific, like on a Mongo node, then it does not have 'Mongo' in the name and instead has 'MongoDB' as the category)
2. 'Add Server' is now 'Attach Mongo Server' ('add' is too similar to 'create', so I chose 'attach')


EricJizbaMSFT:
Renamed to INode and rebased on top of master

Issue #28, opened 2017-08-28
Intellisense in Mongo scrapbook is broken


EricJizbaMSFT:
Connect to a db, open a scrapbook, type 'db'

Expected: A list of collections and functions

Actual: Only the functions listed and:
```
[Error - 11:14:02] (node:15299) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: invalid schema, expected mongodb
[Error - 11:14:02] (node:15299) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

PR #29, opened 2017-08-28
Fix Mongo Scrapbook intellisense

EricJizbaMSFT:
Fix for #28 

In the Server class, `id` used to always be the connection string. However, when I added the Azure resources to the explorer, I refactored `id` and added a new `connectionString` field. Apparently I missed updating one of the fields. See here for the previous commit: https://github.com/Microsoft/vscode-cosmosdb/commit/3f5396480c5d2c06cc0ab2436ef8997dabdcceca#diff-566bee68995619b9ddea388ded5443caR162

### DIFF
--- a/src/mongo/languageClient.ts
+++ b/src/mongo/languageClient.ts
@@ -51,7 +51,7 @@ export default class MongoDBLanguageClient {
 	}
 
 	connect(database: Database): void {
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
@chrisdias FYI - a fix for the issue described in today's sync-up

PR #41, opened 2017-09-03
move to external azure-account api

chrisdias:
moved to the external azure-account extension and moved to using keytar node module.

### DIFF
--- a/src/nodes.ts
+++ b/src/nodes.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import * as vscode from 'vscode';
 import * as path from 'path';
+import * as keytarType from 'keytar';
 
 import { MongoClient } from 'mongodb';
 import { EventEmitter, Event, Command } from 'vscode';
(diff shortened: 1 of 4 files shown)


src/nodes.ts
```diff
@@ -157,9 +168,11 @@ export class AttachedServersNode implements INode {
 		const index = this._attachedServers.findIndex((value) => value.id === server.id);
 		if (index !== -1) {
 			const deletedNodes = this._attachedServers.splice(index, 1);
-			await this._azureAccount.credentials.deleteSecret(this._serviceName, server.id);
-			await this.persistIds();
-			return deletedNodes;
+			if (this._keytar) {
+				await this._keytar.deletePassword(this._serviceName, server.id);
+				await this.persistIds();
+				return deletedNodes;
```
chrmarti: I don't have much context here, but maybe the last two lines should be run even when keytar could not be loaded? (similar above)

src/nodes.ts
```diff
@@ -168,9 +181,8 @@ export class AttachedServersNode implements INode {
 		if (value) {
```
chrmarti: Also check if keytar was loaded?

PR #44, opened 2017-09-11
Update command id

chrmarti:
Small clean up on the account extension's side.


EricJizbaMSFT:
@chrmarti when will the new account extension be published with this change?

chrmarti:
@EricJizbaMSFT: @chrisdias will coordinate with you to roll this out simultaneously (early next week I believe).

EricJizbaMSFT:
I would actually prefer the account extension be updated ASAP so that we can test on the latest version before our release. We haven't published the Cosmos DB or App Service extensions yet, so there's no worry about breaking anyone (not sure if any other extensions have a dependency)

chrisdias:
Good point about not breaking anyone, we will do this today.

chrisdias:
@ericjizbamsft we've published a new version of azure account to the marketplace.

PR #46, opened 2017-09-12
Disp a warning message if working folder is absent.

PrashanthCorp:
1. Using a warning message for now as ShowErrorMessage isn't working. 
2. Will look into allowing scrapbooks and prompting a file name only during save time.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,19 +69,21 @@ function createScrapbook(): Thenable<void> {
 		let count = 1;
 		const max = 99999;
 		if (!vscode.workspace.workspaceFolders){
-			vscode.window.showWarningMessage("No open workspace!");
+			vscode.window.showWarningMessage("No open workspace. We shall create a temporary scrapbook at this moment. Be sure to save this later.");
+			uri = vscode.Uri.file(`Scrapbook-temp.mongo`);
 		}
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -68,6 +68,9 @@ function createScrapbook(): Thenable<void> {
 		let uri: vscode.Uri = null;
 		let count = 1;
 		const max = 99999;
+		if (vscode.workspace.workspaceFolders === undefined){
```
EricJizbaMSFT: You could just do `if (!vscode.workspace.workspaceFolders)`

src/extension.ts
```diff
@@ -69,19 +69,21 @@ function createScrapbook(): Thenable<void> {
...
+				if (!vscode.workspace.textDocuments.find(doc => doc.uri.fsPath === uri.fsPath) && !fs.existsSync(uri.fsPath)) {
+					break;
+				}
+				count++;
+			}
+			if (count === max) {
+				vscode.window.showErrorMessage('Could not create new scrapbook.');
+				return;
 			}
-			count++;
```
PrashanthCorp: The diff tool is mistaken here. Essentially wrapped the previous code in the else clause (if a workspace does exist).

PR #57, opened 2017-09-14
Display mongo server connection error in explorer

### LINKED ISSUE #54: move error message when cannot connect to mongo server
- connect to a local mongodb server
- shut down mongod on your computer
- reload vs code

result:

image

I'm wondering if we should not show a top level error in this case, but localize it to the explorer.  Something like this:

image

EricJizbaMSFT:
Rather than popping up the vscode error message, display it right in the explorer.

Also, delay showing error messages for attached nodes until the user expands the node (rather than when the extension is activated)

Fixes #54

### DIFF
--- a/src/mongo/nodes.ts
+++ b/src/mongo/nodes.ts
@@ -8,11 +8,11 @@ import * as cp from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
 
-import { MongoClient, Db, ReadPreference, Code, Server, Collection, Cursor, ObjectID, MongoError, ReplSet } from 'mongodb';
+import { MongoClient, Db, ReadPreference, Code, Server, Collection, Cursor, ObjectID, MongoError } from 'mongodb';
 import { Shell } from './shell';
 import { EventEmitter, Event, Command } from 'vscode';
 import { AzureAccount } from '../azure-account.api';
-import { INode } from '../nodes';
+import { INode, ErrorNode } from '../nodes';
 import { ResourceManagementClient } from 'azure-arm-resource';
 import docDBModels = require("azure-arm-documentdb/lib/models");
 import DocumentdbManagementClient = require("azure-arm-documentdb");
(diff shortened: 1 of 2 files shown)


src/mongo/nodes.ts
```diff
@@ -74,13 +54,18 @@ export class MongoServerNode implements IMongoServer {
...
-	static async getMongoDatabaseNodes(connectionString: string, parentNode: IMongoServer): Promise<MongoDatabaseNode[]> {
-		const db = await MongoClient.connect(connectionString);
+	static async getMongoDatabaseNodes(connectionString: string, parentNode: IMongoServer): Promise<INode[]> {
+		let db: Db;
 		try {
+			db = await MongoClient.connect(connectionString);
 			const value: { databases: { name }[] } = await db.admin().listDatabases();
 			return value.databases.map(database => new MongoDatabaseNode(database.name, parentNode));
+		} catch (error) {
+			return [new ErrorNode(error.message)];
```
whiskeyjay: Can the message property be a very long string? Might consider limiting the length of the text by trimming it.

EricJizbaMSFT: It can be a long string, but you can always expand the explorer view to see more of the message. Ideally we would be able to wrap the message, but I don't think that's possible.

src/mongo/nodes.ts
```diff
@@ -74,13 +54,18 @@ export class MongoServerNode implements IMongoServer {
 		return MongoServerNode.getMongoDatabaseNodes(this._connectionString, this);
 	}
 
-	static async getMongoDatabaseNodes(connectionString: string, parentNode: IMongoServer): Promise<MongoDatabaseNode[]> {
-		const db = await MongoClient.connect(connectionString);
+	static async getMongoDatabaseNodes(connectionString: string, parentNode: IMongoServer): Promise<INode[]> {
+		let db: Db;
 		try {
+			db = await MongoClient.connect(connectionString);
```
PrashanthCorp: It looks like you moved this line into the try. Was this line causing an issue? 
I'm trying to understand which of the 3 lines in the try block is expected to raise an error. 
And does error.name have any useful information (like a URIError, ReferenceError, etc.) ?

EricJizbaMSFT:
@nturinski we _can't_ create an error node when attaching a db errors. We have to connect to the db to get the host and port, which we use as the label for the main node.

Any comments you put on a review are appreciated and not viewed as revenge :) You did make me realize that I should propagate the error in the attach scenario so that it's caught by our telemetry

src/mongo/nodes.ts
EricJizbaMSFT: The catch block is most relevant to the first line and the finally block is most relevant to the second and third line.

In other words, I moved the 'MongoClient.connect' line into the try block because it is the line that will throw an error if we can't connect. The other two lines _shouldn't_ throw errors, but I want to be sure `db` is closed even if an error does occur (hence the finally block)

Issue #62, opened 2017-09-15
Clicking on collections with no workspace loaded seemingly does nothing


PrashanthCorp:
Repro : 
Log in to Azure, set up a Mongo collection. The resources appear on the side bar. Drill down to a collection and single click – it does nothing.

There is an unhandled promise rejection that manifests as unresponsive UI. 

Expected : A suggestion of the root cause/fix/reason for beahvior
Actual : Nothing happens

A first-time user has no indication of : 
1.  what a left click on a collection is expected to do (since we don’t have a context menu for a collection) 
2.  what failed

Root cause : No workspace
Source : See [the use of workspace.rootpath](https://github.com/Microsoft/vscode-cosmosdb/blob/3a7feeee3f775c4a37ad6ea4fced1f07a6236758/src/mongo/commands.ts#L45)
Following some breakpoints led me there. 

When a workspace exists, a result.json opens as expected.

Issue #13, opened 2017-08-17
Add ability to view multiple results per workspace


EricJizbaMSFT:
We always write results to the same 'result.json' file and we always overwrite the existing content

PR #63, opened 2017-09-15
Collection click - handle when no workspace selected

PrashanthCorp:
Resolve Issue #62

### DIFF
--- a/src/mongo/commands.ts
+++ b/src/mongo/commands.ts
@@ -43,7 +43,6 @@ export class MongoCommands {
 
 	public static showResult(result: string, column?: vscode.ViewColumn): Thenable<void> {
 		let uri : vscode.Uri= null;
-		let openTextDocumentOptions : Object = {language : 'json'};
 		if(vscode.workspace.rootPath){
 			uri = vscode.Uri.file(path.join(vscode.workspace.rootPath, 'result.json'));
 			if (!fs.existsSync(uri.fsPath)) {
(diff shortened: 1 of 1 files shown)


src/mongo/commands.ts
```diff
@@ -42,11 +42,18 @@ export class MongoCommands {
 	}
 
 	public static showResult(result: string, column?: vscode.ViewColumn): Thenable<void> {
-		let uri = vscode.Uri.file(path.join(vscode.workspace.rootPath, 'result.json'));
-		if (!fs.existsSync(uri.fsPath)) {
-			uri = uri.with({ scheme: 'untitled' });
+		let uri : vscode.Uri= null;
+		let openTextDocumentOptions : Object = {language : 'json'};
```
EricJizbaMSFT: 'language' as 'json' doesn't help us in this case. We only register the 'Update' command when a file is named [result.json](https://github.com/Microsoft/vscode-cosmosdb/blob/master/package.json#L173) and the language doesn't matter.

For now I think we should just pop up an error message when a workspace isn't open. We can fix this case when we address this issue: https://github.com/Microsoft/vscode-cosmosdb/issues/13 (I think we should define a 'mongoresult' langauge type, but I don't want to do that before our first release)

EricJizbaMSFT:
@PrashanthCorp FYI if you put `Fixes #62` instead of `Resolve Issue #62` in the description of the PR, then Github will automatically close the issue when you merge the PR

PR #65, opened 2017-09-15
Update README.md

PrashanthCorp:
Fix the shortcut to run a scrapbook

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ With this extension, you can:
 - Click on any DB to open the Mongo shell playground editor
 - Enter your scripts, eg: `db.<collectionName>.find()`
 - IntelliSense (completions) will be provided as you write your scripts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ With this extension, you can:
 - Click on any DB to open the Mongo shell playground editor
 - Enter your scripts, eg: `db.<collectionName>.find()`
 - IntelliSense (completions) will be provided as you write your scripts
-- Select the script and press `CMD+'` (`CTRL+'` on Windows and Linux) to see the results
+- Select the script and press `CMD+'` (`CTRL+SHIFT+'` on Windows and Linux) to see the results
```
EricJizbaMSFT: The context menu lists it as a double quote instead of shift single quote. Can we match that? Also make sure to update the mac one
screen shot 2017-09-15 at 4 15 34 pm

PR #73, opened 2017-09-19
Allow user to change page limit

PrashanthCorp:
Redoing [PR#66](https://github.com/Microsoft/vscode-cosmosdb/pull/66) - mixed up some commits there due to a rebase. 

package.json : expose max docs as global setting; 
nodes.ts: import that setting when using maxdocs

### DIFF
--- a/package.json
+++ b/package.json
@@ -275,6 +275,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show or hide the Cosmos DB Explorer"
+				},
+				"cosmosDB.mongo.maxDocs" : {
+					"type":"integer",
+					"default":20,
+					"description": "Set a page limit for number of results(documents) displayed"
 				}
 			}
 		}
--- a/package.json
+++ b/package.json
@@ -276,9 +276,9 @@
 					"default": true,
 					"description": "Show or hide the Cosmos DB Explorer"
 				},
-				"cosmosDB.mongo.maxDocs" : {
-					"type":"integer",
-					"default":20,
+				"cosmosDB.mongo.maxDocs": {
+					"type": "integer",
+					"default": 20,
 					"description": "Set a page limit for number of results(documents) displayed"
 				}
 			}
--- a/src/mongo/nodes.ts
+++ b/src/mongo/nodes.ts
@@ -268,7 +268,7 @@ export class MongoCollectionNode implements INode {
 	}
 
 	private find(args?: any): Thenable<string> {
-		const maxDocs = 20;
+		const maxDocs = <number>vscode.workspace.getConfiguration().get('cosmosDB.mongo.maxDocs');
 		return this.collection.find(args).limit(maxDocs)
 			.toArray().then(docs => this.stringify(docs));
 	}
--- a/src/mongo/nodes.ts
+++ b/src/mongo/nodes.ts
@@ -268,7 +268,10 @@ export class MongoCollectionNode implements INode {
 	}
 
 	private find(args?: any): Thenable<string> {
-		const maxDocs = <number>vscode.workspace.getConfiguration().get('cosmosDB.mongo.maxDocs');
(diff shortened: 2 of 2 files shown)


PrashanthCorp:
@Eric : The reasoning was that when CosmosDB gets more API's on board, then this distinction would be required. I put the mongo as a qualifier if this setting is unique, and cosmosdb to mark the extension using it. 

Your thoughts?

EricJizbaMSFT:
On one hand it is the CosmosDB extension, but on the other hand you can attach Mongo servers that aren't hosted in Azure and have nothing to do with CosmosDB. I would prefer to keep it consistent with the `mongo.shell.path` setting above that just prefixes it with mongo.

EricJizbaMSFT:
But I also still think you could just do `cosmosDB.maxDocs`. The other apis should have similar behavior and we could use the same setting for each

package.json
```diff
@@ -275,6 +275,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show or hide the Cosmos DB Explorer"
+				},
+				"cosmosDB.mongo.maxDocs" : {
+					"type":"integer",
+					"default":20,
```
whiskeyjay: space after colon

package.json
```diff
@@ -275,6 +275,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show or hide the Cosmos DB Explorer"
+				},
+				"cosmosDB.mongo.maxDocs" : {
+					"type":"integer",
```
whiskeyjay: space after colon

package.json
```diff
@@ -275,6 +275,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Show or hide the Cosmos DB Explorer"
+				},
+				"cosmosDB.mongo.maxDocs" : {
```
whiskeyjay: no space before colon

src/mongo/nodes.ts
```diff
@@ -268,7 +268,10 @@ export class MongoCollectionNode implements INode {
 	}
 
 	private find(args?: any): Thenable<string> {
-		const maxDocs = 20;
+		let maxDocs = <number>vscode.workspace.getConfiguration().get('cosmosDB.mongo.maxDocs');
```
whiskeyjay: Is getConfiguration method guaranteed not to return null? Otherwise this might be a potential null ref error.

whiskeyjay: The get method accepts a type parameter:

```
get<number>('cosmosDB.mongo.maxDocs');
```

src/mongo/nodes.ts
```diff
@@ -268,7 +268,10 @@ export class MongoCollectionNode implements INode {
...
+		if(maxDocs <= 0){
```
whiskeyjay: missing spaces:
```
if (maxDocs <= 0) {
```

whiskeyjay: Could be one line:
```
maxDocs = maxDocs > 0 ? maxDocs : 20;
```

src/mongo/nodes.ts
```diff
@@ -268,7 +268,15 @@ export class MongoCollectionNode implements INode {
 	}
 
 	private find(args?: any): Thenable<string> {
-		const maxDocs = 20;
+		let maxDocs: number = 0; 
+		try{
+		maxDocs = vscode.workspace.getConfiguration().get<number>('cosmosDB.mongo.maxDocs');
```
whiskeyjay: indent

src/mongo/nodes.ts
```diff
@@ -268,7 +268,15 @@ export class MongoCollectionNode implements INode {
...
+		}
+		catch(error){
+		}
+		finally{
+		maxDocs = maxDocs > 0 ? maxDocs : 20;
```
whiskeyjay: indent

PR #87, opened 2017-09-29
Adding DocumentDB support - databases, collections

### LINKED ISSUE #75: View data for DocumentDB Accounts in Explorer
This includes a Database node and a Collection node in the explorer

PrashanthCorp:
Fixes #75


src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+		arguments: [this],
+		title: ''
+	};
+
+
+	async getDocuments(): Promise<any> {
+		let dbLink: string = this.db.getDbLink();
+		let client = new DocumentDBConnectionClient(this.db.server.getEndpoint(), {masterKey: this.db.server.getPrimaryMasterKey() });
+		let docs;
+		try{
```
EricJizbaMSFT: Remove the try/catch

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+			let docs = await this.readOneCollection(collSelfLink, client);
+			console.log(docs);
+			return await docs; 
+		} catch (error){
+			console.log(error.message);
+			return;
+		}
+	}
+
+	async readOneCollection(selfLink, client): Promise<any>{
```
EricJizbaMSFT: This is only used once, so just move the logic into getDocuments.

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+	async listDatabases(client): Promise<any[]> {
+		let databases = await client.readDatabases();
+		return await new Promise<any[]>((resolve, reject) => {
+			databases.toArray( (err, dbs: Array<Object>) => err ? reject(err) : resolve(dbs) );
+		});
+	}
+
+	async getDocDBDatabaseNodes(client, DocDBServerNodeInstance): Promise<INode[]> {
+		let databases;
+		try {
```
EricJizbaMSFT: Remove the try/catch

As dicussed offline - we want errors to be consistenly displayed by VS Code and we want the error to bubble up so we track it in telemetry

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+import * as path from 'path';
+import { EventEmitter, Event, Command } from 'vscode';
+import { AzureAccount } from '../azure-account.api';
+import { INode, ErrorNode } from '../nodes';
+import { ResourceManagementClient } from 'azure-arm-resource';
+import docDBModels = require("azure-arm-documentdb/lib/models");
+import DocumentdbManagementClient = require("azure-arm-documentdb");
+import { MongoDatabaseNode } from '../mongo/nodes';
+import {MongoCommands} from '../mongo/commands'
+var DocumentDBConnectionClient = require("documentdb").DocumentClient;
```
EricJizbaMSFT: I looked into this more and if you add `"@types/documentdb": "^1.10.2",` to devDependencies in package.json, you should be able to do the normal import like this:
```
import { DocumentClient } from 'documentdb';
```

Plus then you get all the helpful typings and stuff :)

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+		return this._primaryMasterKey;
+	}
+
+	getEndpoint(): string {
+		return this._endpoint;
+	}
+
+	getChildren(): Promise<INode[]> {
+		//return new Promise<DocDBServerNode[]>;
+		let client = new DocumentDBConnectionClient(this.getEndpoint(), this.getPrimaryMasterKey());
```
EricJizbaMSFT: Looks like this need to be an object `{ masterKey: this.getPrimaryMasterKey() }`

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+		} catch (error){
+			console.log(error.message);
+			return;
+		}
+	}
+
+	async readOneCollection(selfLink, client): Promise<any>{
+		let documents = await client.readDocuments(selfLink);
+		return await new Promise<any[]>((resolve, reject) => {
+			documents.toArray((err, docs: Array<Object>) => err ? reject(err) : resolve(docs));
```
EricJizbaMSFT: As discussed in standup, we need to limit the max number of documents. Maybe we should use the 'executeNext' function instead of 'toArray' (Also when you call `readDocuments`, you can pass in `FeedOptions` that has a `maxItemCount`)

src/docdb/nodes.ts
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	async listDatabases(client): Promise<any[]> {
+		let databases = await client.readDatabases();
+		return await new Promise<any[]>((resolve, reject) => {
+			databases.toArray( (err, dbs: Array<Object>) => err ? reject(err) : resolve(dbs) );
+		});
+	}
+
+	async getDocDBDatabaseNodes(client, DocDBServerNodeInstance): Promise<INode[]> {
```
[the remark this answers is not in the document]
PrashanthCorp: I realized the single-use here makes the code more spaghetti like. 
My reasoning was that `getChildren` is a specification in vscode.treeitem [here](https://github.com/Microsoft/vscode/blob/b64aa9b199874814c72940f1e4b5d995bbd3d0f7/src/vs/workbench/common/views.ts#L48). It was not clear to me if that was async, so I used `getChildren` merely to cal the async functions l `listDatabases`, `getDocDBDatabaseNodes`

src/docdb/nodes.ts
```diff
@@ -0,0 +1,151 @@
...
+	async getDocuments(): Promise<any> {
+		let dbLink: string = this.db.getDbLink();
+		let client = new DocumentClient(this.db.server.getEndpoint(), {masterKey: this.db.server.getPrimaryMasterKey() });
+		let collSelfLink = dbLink + "/colls/" + this.id;
+		let docs = await this.readOneCollection(collSelfLink, client);
+		return await docs; 
+	}
+
+	async readOneCollection(selfLink, client): Promise<any>{
+		let documents = await client.readDocuments(selfLink, {maxItemCount : 20});
```
PrashanthCorp: Looks like the maxDocs setting got checked in after I started pushing this in. Give it a configurable setting after this check-in.

src/nodes.ts
```diff
@@ -13,6 +13,8 @@ import { AzureAccount, AzureResourceFilter } from './azure-account.api';
 import { ResourceManagementClient } from 'azure-arm-resource';
 import docDBModels = require("azure-arm-documentdb/lib/models");
 import DocumentdbManagementClient = require("azure-arm-documentdb");
+import { DocDBServerNode } from './docdb/nodes';
+var DocumentDBConnectionClient = require("documentdb").DocumentClient;
```
EricJizbaMSFT: import { DocumentClient } from 'documentdb';

src/nodes.ts
```diff
@@ -102,11 +106,23 @@ export class CosmosDBResourceNode implements IMongoServer {
...
+
 	async getChildren(): Promise<INode[]> {
 		if (this._isMongo) {
 			const connectionString = await this.getConnectionString();
 			return MongoServerNode.getMongoDatabaseNodes(connectionString, this);
 		}
+		if (this._isDocDB) {
+			const masterKey = await this.getMasterKey();
+			let client = new DocumentDBConnectionClient(this._databaseAccount.documentEndpoint, {masterKey: masterKey});
+			let DocDBServer = new DocDBServerNode(<string>masterKey, client.id, this._databaseAccount.documentEndpoint);
```
EricJizbaMSFT: Since you changed the signature of `getMasterKey()` to `string`, now you can change `<string>masterKey` to `masterKey`

EricJizbaMSFT: There's no need to create a node just to call getChildren. I would prefer if we called a _static_ method on DocDBServerNode and just pass in the masterKey and endpoint

(Right now there's no need to _ever_ call the constructor for DocDBServerNode. The constructor for the MongoServerNode is only called in the 'Attach' scenario, but we don't support that yet for DocDB servers)

src/docdb/nodes.ts
EricJizbaMSFT: Sounds good - thanks for creating an issue on it (I assume that's what this is for? https://github.com/Microsoft/vscode-cosmosdb/issues/88)

src/docdb/nodes.ts
```diff
@@ -0,0 +1,160 @@
...
+		return this._primaryMasterKey;
+	}
+
+	getEndpoint(): string {
+		return this._endpoint;
+	}
+
+	async getChildren(): Promise<INode[]> {
+		let client = new DocumentClient(this.getEndpoint(), { masterKey: this.getPrimaryMasterKey() });
+		let DocDBServerNodeInstance = this;
```
EricJizbaMSFT: DocDBServerNodeInstance is never used - remove it.

(sidenote: variables should always start with lowercase letters. It helps differentiate from class names, which are always uppercase)

PR #93, opened 2017-10-05
DocDB : Allow creation of databases and collections

PrashanthCorp:
- Add Icons for DocumentDB resources. 
- Create collections and databases.


src/extension.ts
```diff
@@ -149,6 +154,89 @@ async function createMongoDatabase(server: IMongoServer) {
...
+}
+
+async function createDocDBCollection(db: DocDBDatabaseNode) {
+	const collectionName = await vscode.window.showInputBox({ placeHolder: 'Collection Name' });
+	if (collectionName) {
+		let masterKey = db.getPrimaryMasterKey();
+		let endpoint = db.getEndpoint();
+		let options = {};
+		let partitionKey: string = await vscode.window.showInputBox({
+			prompt: 'Partition Key: Choose a JSON property name that will possibly have a wide range of values.',
```
PrashanthCorp: I need to remove the trailing period for consistency.

package.json
```diff
@@ -34,6 +34,8 @@
 		"onCommand:cosmosDB.createAccount",
 		"onCommand:cosmosDB.removeMongoServer",
 		"onCommand:cosmosDB.createMongoDatabase",
+		"onCommand:cosmosDB.createDocDBDatabase",
+		"onCommand:cosmosDB.createDocDBCollection",
```
PrashanthCorp: Currently placed these commands related to similar scenarios (creation of mongo and DocDB entities). 
Should these be made alphabetical?

package.json
```diff
@@ -212,6 +224,14 @@
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServer"
 				},
 				{
+					"command": "cosmosDB.createDocDBCollection",
+					"when": "view == cosmosDBExplorer && viewItem == DocDbDatabase"
```
PrashanthCorp: Need to change context value to DocDBDatabase to adhere to CamelCase standard.

src/docdb/nodes.ts
```diff
@@ -85,8 +85,8 @@ export class DocDBDatabaseNode implements INode {
 
 	get iconPath(): any {
 		return {
-			light: path.join(__filename, '..', '..', '..', '..', 'resources', 'icons', 'light', 'Database.svg'),
-			dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'icons', 'dark', 'Database.svg')
+			light: path.join(__filename, '..', '..', '..', '..', 'resources', 'icons', 'theme-agnostic', 'Azure DocumentDB - database LARGE.svg'),
```
PrashanthCorp: Should these files be renamed? 
I've left the names as is from the [public icon set](https://www.microsoft.com/en-us/download/details.aspx?id=41937).

package.json
EricJizbaMSFT: I don't have a strong preference

src/docdb/nodes.ts
EricJizbaMSFT: I don't have a strong preference - but how do these icons help solve the main issue? I thought the main goal was to get different icons for graph/mongo/table/docdb

src/extension.ts
```diff
@@ -149,6 +154,89 @@ async function createMongoDatabase(server: IMongoServer) {
...
+	const databaseName = await vscode.window.showInputBox({ placeHolder: 'Database Name' });
+	if (databaseName) {
+		const masterKey = await server.getPrimaryMasterKey();
+		const endpoint = await server.getEndpoint();
+		let client = new DocumentClient(endpoint, { masterKey: masterKey });
+		client.createDatabase({ id: databaseName }, function (err, created) {
+			if (!err) {
+				vscode.window.showInformationMessage("Created a db with name " + databaseName);
+			} else {
+				vscode.window.showErrorMessage(err.body);
```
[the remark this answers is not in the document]
PrashanthCorp: err.body is from the API, whenever there's a 400 or 401 error. 
I'll place each of them in a try-catch though.

src/extension.ts
```diff
@@ -149,6 +154,89 @@ async function createMongoDatabase(server: IMongoServer) {
...
+				}
+				);
+			}
+		}
+	}
+}
+
+function validatePartitionKey(key: string): string | undefined | null {
+	if (key[0] != '/') {
+		return "Need a leading / in the partitionKey";
```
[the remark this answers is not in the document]
PrashanthCorp: IIRC, the portal displays the leading / when it corrects the name. We don't have such control (overwriting what the user typed in), so I preferred suggesting to the user.

package.json
```diff
@@ -111,6 +113,16 @@
...
 			},
 			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.createDocDBDatabase",
+				"title": "Create Database"
+			},
+			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.createDocDBCollection",
+				"title": "Create Collection"
```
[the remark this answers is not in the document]
PrashanthCorp: I only address DocumentDB atm. 
The others might need other API's for wrangling with the graphs and tables.

src/extension.ts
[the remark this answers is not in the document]
StephenWeatherford: Can you prefill the textbox with "/"?

src/extension.ts
StephenWeatherford: too long

src/extension.ts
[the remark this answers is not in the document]
PrashanthCorp: Just ran tsc on the file, the err thrown here is of type QueryError. It is definite to have a err.body, not err.message.

src/extension.ts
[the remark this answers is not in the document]
PrashanthCorp: It also looks like having '/folder//file' will be also problem. I'll get back to the double-slash later. 
@StephenWeatherford I don't know of a way. I can add a default slash, but that gets overwritten once the user starts typing.

package.json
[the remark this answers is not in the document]
PrashanthCorp: The commands are shown only for DocumentDB account dbs/colls. 
I've made changes to that effect [here](https://github.com/Microsoft/vscode-cosmosdb/pull/93/files#diff-55f44dc2e4ecace970a4bac3d92a80ccR21)
I didn't do it for collection because we don't *yet* have any commands which ought to behave differently based on them

src/commands.ts
```diff
@@ -201,6 +206,104 @@ export class CosmosDBCommands {
...
+            validateInput: CosmosDBCommands.validateDatabaseName,
+            ignoreFocusOut: true
+        });
+        if (databaseName) {
+            const masterKey = await server.getPrimaryMasterKey();
+            const endpoint = await server.getEndpoint();
+            let client = new DocumentClient(endpoint, { masterKey: masterKey });
+            client.createDatabase({ id: databaseName }, async function (err, created) {
+                if (err) {
+                    vscode.window.showErrorMessage(err.body);
```
EricJizbaMSFT: As discussed offline - don't ever use showErrorMessage. Let the error bubble up so that telemetry catches it

(I thought VS Code always showed messages, but they don't. So you will have to modify the code here to show a message instead of throwing the error:
https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L91 )

src/docdb/nodes.ts
```diff
@@ -3,78 +3,38 @@
...
 
 export class DocDBDatabaseNode implements INode {
-	readonly contextValue: string = 'DocDbDatabase';
-
-	constructor(readonly id: string, readonly _primaryMasterKey: string, readonly _endPoint: string) {
+	readonly contextValue: string;
+	constructor(readonly id: string, readonly _primaryMasterKey: string, readonly _endPoint: string, readonly defaultExperience: string) {
+		switch (defaultExperience) {
+			case "DocumentDB":
+				this.contextValue = "DocDBDatabase"
```
EricJizbaMSFT: Ignore table and graph for now. Just use DocDB all the time

src/extension.ts
```diff
@@ -88,7 +90,7 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
 		} catch (err) {
 			result = 'Failed';
 			errorData = util.errToString(err);
-			throw err;
+			vscode.window.showErrorMessage(err);
```
EricJizbaMSFT: Usually we want to display `err.message`
```
if (err instanceof Error) {
                    vscode.window.showErrorMessage(err.message);
                }
```

src/commands.ts
```diff
@@ -201,6 +206,116 @@ export class CosmosDBCommands {
...
+                        client.createCollection(db.getDbLink(), collectionDef, options, (err, result) => {
+                            if (err) {
+                                reject(err.body);
+                            }
+                            else {
+                                resolve(result);
+                            }
+                        });
+                    });
+                    /*
```
EricJizbaMSFT: Remove commented code

src/commands.ts
```diff
@@ -201,6 +206,104 @@ export class CosmosDBCommands {
...
+        const databaseName = await vscode.window.showInputBox({
+            placeHolder: 'Database Name',
+            validateInput: CosmosDBCommands.validateDatabaseName,
+            ignoreFocusOut: true
+        });
+        if (databaseName) {
+            const masterKey = await server.getPrimaryMasterKey();
+            const endpoint = await server.getEndpoint();
+            let client = new DocumentClient(endpoint, { masterKey: masterKey });
+            client.createDatabase({ id: databaseName }, async function (err, created) {
```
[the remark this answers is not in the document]
PrashanthCorp: I'm going to make the resolve-reject block a single-line command, to make it consistent to [this](https://github.com/Microsoft/vscode-cosmosdb/pull/87/files#diff-55f44dc2e4ecace970a4bac3d92a80ccR62), based off an offline discussion we had around that time.

Issue #35, opened 2017-08-29
Several commands fail when run from the Command Pallete


EricJizbaMSFT:
There are several commands that expect a node from the explorer as input and fail if not run from the explorer. For example the 'Drop MongoDB' command fails with an error "Running the contributed command:'cosmosDB.dropMongoDB' failed" when run from the command pallete

Other commands silently fail without an error message, like 'Copy Connection String'

Ideally these commands will open a QuickPick dialog so that the user can select a node

PR #97, opened 2017-10-11
Implement drop DB and Drop Collection 

### LINKED ISSUE #79: Create/Delete new Collection (for DocumentDB accounts)
The Mongo command creates a Database and collection at the same time. We might want to do the same here or we might want to have separate commands for 'Create Database' and 'Create Collection'

PrashanthCorp:
Fixes #79

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,9 +37,9 @@
 		"onCommand:cosmosDB.createDocDBDatabase",
 		"onCommand:cosmosDB.createDocDBCollection",
 		"onCommand:cosmosDB.connectMongoDB",
-		"onCommand:cosmosDB.dropMongoDB",
-		"onCommand:cosmosDB.dropDocDBDatabase",
-		"onCommand:cosmosDB.dropDocDBCollection",
+		"onCommand:cosmosDB.deleteMongoDB",
+		"onCommand:cosmosDB.deleteDocDBDatabase",
+		"onCommand:cosmosDB.deleteDocDBCollection",
 		"onCommand:cosmosDB.newMongoScrapbook",
 		"onCommand:cosmosDB.updateMongoDocuments",
 		"onCommand:cosmosDB.openMongoCollection",
@@ -136,18 +136,18 @@
 			},
 			{
 				"category": "MongoDB",
-				"command": "cosmosDB.dropMongoDB",
-				"title": "Drop"
+				"command": "cosmosDB.deleteMongoDB",
+				"title": "Delete"
 			},
 			{
 				"category": "DocumentDB",
-				"command": "cosmosDB.dropDocDBDatabase",
-				"title": "Drop Database"
+				"command": "cosmosDB.deleteDocDBDatabase",
+				"title": "Delete Database"
 			},
 			{
 				"category": "DocumentDB",
-				"command": "cosmosDB.dropDocDBCollection",
-				"title": "Drop Collection"
+				"command": "cosmosDB.deleteDocDBCollection",
+				"title": "Delete Collection"
 			},
 			{
 				"category": "Cosmos DB",
@@ -252,15 +252,15 @@
 					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
 				},
 				{
-					"command": "cosmosDB.dropMongoDB",
+					"command": "cosmosDB.deleteMongoDB",
 					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
 				},
 				{
-					"command": "cosmosDB.dropDocDBCollection",
+					"command": "cosmosDB.deleteDocDBCollection",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentCollection"
 				},
 				{
-					"command": "cosmosDB.dropDocDBDatabase",
+					"command": "cosmosDB.deleteDocDBDatabase",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentDatabase"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -257,11 +257,11 @@
 				},
 				{
 					"command": "cosmosDB.dropDocDBCollection",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentDatabase"
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentCollection"
 				},
 				{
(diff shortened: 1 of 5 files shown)


src/commands.ts
```diff
@@ -317,7 +309,36 @@ export class CosmosDBCommands {
...
+    public static async dropDocDBDatabase(db: DocDBDatabaseNode, explorer: CosmosDBExplorer): Promise<void> {
+        const confirmed = await vscode.window.showWarningMessage("Are you sure to delete database: " + db.label + " and its collections?",
+            "Yes", "No");
+        if (confirmed === "Yes") {
+            const masterKey = await db.getPrimaryMasterKey();
+            const endpoint = await db.getEndpoint();
+            const client = new DocumentClient(endpoint, { masterKey: masterKey });
+            await new Promise((resolve, reject) => {
+                client.deleteDatabase(db.getDbLink(), function (err, result) {
+                    err ? reject(err.body) : resolve(result);
```
EricJizbaMSFT: This 'reject' won't ever be displayed to the user. You could do this:
```
reject(new Error(err.body))
```
But I would actually modify extension.ts so that you don't have to fix other cases:
```
else if (typeof err === "string") {
    vscode.window.showErrorMessage(err);
}
```

(Or you could do both)

src/commands.ts
```diff
@@ -317,7 +309,36 @@ export class CosmosDBCommands {
...
+    public static async dropDocDBCollection(coll: DocDBCollectionNode, explorer: CosmosDBExplorer): Promise<void> {
+        const confirmed = await vscode.window.showWarningMessage("Are you sure to delete collection: " + coll.label + "?", "Yes", "No");
+        if (confirmed === "Yes") {
+            const masterKey = await coll.db.getPrimaryMasterKey();
+            const endpoint = await coll.db.getEndpoint();
+            const client = new DocumentClient(endpoint, { masterKey: masterKey });
+            const collLink = coll.db.getDbLink() + '/colls/' + coll.id;
+            await new Promise((resolve, reject) => {
+                client.deleteCollection(collLink, (err, result) => {
+                    err ? reject(err.body) : resolve(result);
```
EricJizbaMSFT: remove result and just do `resolve()` (This should fix the CI errors)

src/commands.ts
```diff
@@ -317,7 +309,36 @@ export class CosmosDBCommands {
         }
         return null;
     }
-
+    public static async dropDocDBDatabase(db: DocDBDatabaseNode, explorer: CosmosDBExplorer): Promise<void> {
+        const confirmed = await vscode.window.showWarningMessage("Are you sure to delete database: " + db.label + " and its collections?",
```
EricJizbaMSFT: Add 'you want' to the messages and instead of using a colon, wrap the label in quotes. Something like this:

`"Are you sure you want to delete database '" + db.label + "' and its collections?"`

src/commands.ts
```diff
@@ -317,7 +309,36 @@ export class CosmosDBCommands {
...
+            explorer.refresh(db.server);
+        }
+    }
+    public static async dropDocDBCollection(coll: DocDBCollectionNode, explorer: CosmosDBExplorer): Promise<void> {
+        const confirmed = await vscode.window.showWarningMessage("Are you sure to delete collection: " + coll.label + "?", "Yes", "No");
+        if (confirmed === "Yes") {
+            const masterKey = await coll.db.getPrimaryMasterKey();
+            const endpoint = await coll.db.getEndpoint();
+            const client = new DocumentClient(endpoint, { masterKey: masterKey });
+            const collLink = coll.db.getDbLink() + '/colls/' + coll.id;
```
EricJizbaMSFT: Since you have a 'getDbLink' function, I would prefer you moved this logic into a 'getCollLink' function on the collection node

package.json
```diff
@@ -138,6 +140,16 @@
...
 			},
 			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.dropDocDBDatabase",
+				"title": "Drop Database"
+			},
+			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.dropDocDBCollection",
+				"title": "Drop Collection"
```
EricJizbaMSFT: The portal uses 'Delete' instead of 'Drop'. Can you change all occurrences to Delete?

(Including Mongo - I'm not sure why it was 'Drop' in the first place)

PrashanthCorp: I think Drop comes from the sequel terminology of dropping a table.

src/extension.ts
```diff
@@ -177,8 +184,8 @@ async function removeMongoServer(node: INode) {
 	}
 }
 
-function dropDatabase(database: MongoDatabaseNode): void {
-	vscode.window.showInformationMessage('Are you sure you want to drop the database \'' + database.id + '\' and its collections?', { modal: true }, 'Drop')
+function deleteDatabase(database: MongoDatabaseNode): void {
+	vscode.window.showInformationMessage('Are you sure you want to delete database \'' + database.id + '\' and its collections?', { modal: true }, 'Drop')
```
EricJizbaMSFT: Can you change this message to be the same as your other messages?
1. warning instead of info
1. not modal
1. 'yes' and 'no' instead of 'drop'

src/commands.ts
```diff
@@ -317,7 +309,36 @@ export class CosmosDBCommands {
         }
         return null;
     }
-
+    public static async deleteDocDBDatabase(db: DocDBDatabaseNode, explorer: CosmosDBExplorer): Promise<void> {
+        const confirmed = await vscode.window.showWarningMessage("Are you sure you want to delete database '" + db.label + "' and its collections?",
```
EricJizbaMSFT: The user will get a null ref if they try to run this command from the command palette. Wrap this in 'if (db)' for now

We should probably leverage [this](https://code.visualstudio.com/docs/extensionAPI/extension-points#_context-specific-visibility-of-command-palette-menu-items) to hide commands from the palette until they work, but that should be a separate PR to fix https://github.com/Microsoft/vscode-cosmosdb/issues/35

danieltskv:
[Add iPhone accessibility features status to telemetry logs](https://trello.com/c/bna15IiN/278-add-iphone-accessibility-features-status-to-telemetry-logs)

PrashanthCorp:
Hi Daniel, I'm unable to see the trello card you've shared. Is there a chance you can change the permissions?

PR #99, opened 2017-10-13
Persist documents to Azure, create&delete documents

### LINKED ISSUE #83: Edit and Persist DocumentDb docs

PrashanthCorp:
Fixes #83 

Also: 
Consolidate the update commands to cause less confusion
Remove OpenCollection - the UI wouldn't make sense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,10 +36,12 @@
 		"onCommand:cosmosDB.createMongoDatabase",
 		"onCommand:cosmosDB.createDocDBDatabase",
 		"onCommand:cosmosDB.createDocDBCollection",
+		"onCommand:cosmosDB.createDocDBDocument",
 		"onCommand:cosmosDB.connectMongoDB",
 		"onCommand:cosmosDB.deleteMongoDB",
 		"onCommand:cosmosDB.deleteDocDBDatabase",
 		"onCommand:cosmosDB.deleteDocDBCollection",
+		"onCommand:cosmosDB.deleteDocDBDocument",
 		"onCommand:cosmosDB.newMongoScrapbook",
 		"onCommand:cosmosDB.updateDocuments",
 		"onCommand:cosmosDB.openMongoCollection",
@@ -125,6 +127,11 @@
 				"title": "Create Collection"
 			},
 			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.createDocDBDocument",
+				"title": "Create Document"
+			},
+			{
 				"category": "MongoDB",
 				"command": "cosmosDB.removeMongoServer",
 				"title": "Remove Server"
@@ -148,6 +155,11 @@
 				"category": "DocumentDB",
 				"command": "cosmosDB.deleteDocDBCollection",
 				"title": "Delete Collection"
+			},
+			{
+				"category": "DocumentDB",
+				"command": "cosmosDB.deleteDocDBDocument",
+				"title": "Delete Document"
 			},
 			{
 				"category": "Cosmos DB",
@@ -236,6 +248,10 @@
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServer"
 				},
 				{
+					"command": "cosmosDB.createDocDBDocument",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentCollection"
+				},
+				{
 					"command": "cosmosDB.createDocDBCollection",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentDatabase"
 				},
@@ -258,6 +274,10 @@
 				{
 					"command": "cosmosDB.deleteDocDBCollection",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentCollection"
+				},
+				{
+					"command": "cosmosDB.deleteDocDBDocument",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocument"
 				},
 				{
 					"command": "cosmosDB.deleteDocDBDatabase",
@@ -290,6 +310,10 @@
 				{
 					"command": "cosmosDB.refresh",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServer"
+				},
+				{
+					"command": "cosmosDB.refresh",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentCollection"
 				},
 				{
 					"command": "cosmosDB.refresh",
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
 			{
 				"category": "CosmosDB",
 				"command": "cosmosDB.update",
-				"title": "Update"
+				"title": "Update to Azure"
 			},
 			{
 				"category": "MongoDB",
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
 				"title": "Copy Connection String"
 			},
 			{
-				"category": "MongoDB",
+				"category": "CosmosDB",
 				"command": "cosmosDB.updateMongoDocuments",
 				"title": "Update"
 			},
@@ -198,11 +198,6 @@
 				"category": "DocumentDB",
 				"command": "cosmosDB.openDocDBDocument",
 				"title": "Open Document"
-			},
-			{
-				"category": "DocumentDB",
-				"command": "cosmosDB.updateDocDBDocument",
-				"title": "Update Document"
 			}
 		],
 		"menus": {
@@ -213,10 +208,6 @@
 				},
 				{
 					"command": "cosmosDB.updateMongoDocuments",
-					"when": "resourceFilename==result.json"
-				},
-				{
-					"command": "cosmosDB.updateDocDBDocument",
 					"when": "resourceFilename==result.json"
 				}
 			],
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
 		"onCommand:cosmosDB.newMongoScrapbook",
 		"onCommand:cosmosDB.updateDocuments",
 		"onCommand:cosmosDB.openMongoCollection",
-		"onCommand:cosmosDB.openDocDBCollection",
 		"onCommand:cosmosDB.openDocDBDocument",
 		"onCommand:cosmosDB.refresh",
 		"onCommand:cosmosDB.openInPortal",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -36,14 +36,16 @@
...
 		"onCommand:cosmosDB.createDocDBCollection",
+		"onCommand:cosmosDB.createDocDBDocument",
 		"onCommand:cosmosDB.connectMongoDB",
 		"onCommand:cosmosDB.deleteMongoDB",
 		"onCommand:cosmosDB.deleteDocDBDatabase",
 		"onCommand:cosmosDB.deleteDocDBCollection",
+		"onCommand:cosmosDB.deleteDocDBDocument",
 		"onCommand:cosmosDB.newMongoScrapbook",
-		"onCommand:cosmosDB.updateMongoDocuments",
+		"onCommand:cosmosDB.updateDocuments",
```
PrashanthCorp: Couldn't come up with a clear name for this. 
We are trying to consolidate two slightly different operations: 
1. Mongo - Update **collection**
2. DocDB - Update **document**

Any feedback?

package.json
```diff
@@ -36,14 +36,16 @@
...
 		"onCommand:cosmosDB.openMongoCollection",
-		"onCommand:cosmosDB.openDocDBCollection",
```
PrashanthCorp: Removed this because a left-click on the collection would give you the documents. Also, we aren't allowing any operations on the collection anyway.

src/commands.ts
```diff
@@ -235,6 +236,24 @@ export class CosmosDBCommands {
...
+            client.createDocument(coll.getCollLink(), { 'id': 'temporary' }, (err, result) => {
+                if (err) {
+                    reject(err.body);
+                }
+                else {
+                    resolve(result);
+                }
+            });
+        });
+        explorer.refresh(coll);
```
PrashanthCorp: I only do a refresh of the collection node and not open the temporary document. 
The reason was that if the user has any un-persisted changes on the document they are working on, we wouldn't want to overwrite that.

src/commands.ts
```diff
@@ -235,6 +236,24 @@ export class CosmosDBCommands {
         }
     }
 
+    public static async createDocDBDocument(coll: DocDBCollectionNode, explorer: CosmosDBExplorer) {
+        const masterKey = coll.db.getPrimaryMasterKey();
+        const endpoint = coll.db.getEndpoint();
+        const client = new DocumentClient(endpoint, { masterKey: masterKey });
+        await new Promise((resolve, reject) => {
+            client.createDocument(coll.getCollLink(), { 'id': 'temporary' }, (err, result) => {
```
PrashanthCorp: Using a catch-all 'temporary' word here. This would fail if the user creates a doc twice without changing the 1st doc's ID. 

Any thoughts on how to get around this?

src/commands.ts
```diff
@@ -343,6 +362,48 @@ export class CosmosDBCommands {
...
         }
     }
+
+    public static async deleteDocDBDocument(doc: DocDBDocumentNode, explorer: CosmosDBExplorer): Promise<void> {
+        if (doc) {
+            const confirmed = await vscode.window.showWarningMessage("Are you sure you want to delete document '" + doc.label + "'?", "Yes", "No");
+            if (confirmed === "Yes") {
+                const masterKey = await doc.coll.db.getPrimaryMasterKey();
+                const endpoint = await doc.coll.db.getEndpoint();
+                const client = new DocumentClient(endpoint, { masterKey: masterKey });
```
PrashanthCorp: I notice I'm writing these 3 lines a LOT of the time. And I only use the endpoint and masterKey for one purpose - to create a client. 
Low-priority work item to refactor every use into one GenerateClient function (that handles whether a document, collection, db or server is passed)?

src/commands.ts
```diff
@@ -343,6 +362,48 @@ export class CosmosDBCommands {
...
+                const docLink = doc.getDocLink();
+                await new Promise((resolve, reject) => {
+                    client.deleteDocument(docLink, (err) => {
+                        err ? reject(new Error(err.body)) : resolve();
+                    });
+                });
+                explorer.refresh(doc.coll);
```
PrashanthCorp: Edge case - what if the user proposes to delete the document that is open? Should I check for that here and then close the result.json if that's the case?

src/extension.ts
```diff
@@ -17,15 +17,18 @@ import { CosmosDBCommands } from './commands';
...
+import { DocumentClient } from 'documentdb';
 import MongoDBLanguageClient from './mongo/languageClient';
 import { Reporter } from './telemetry';
 
 let connectedDb: MongoDatabaseNode = null;
 let languageClient: MongoDBLanguageClient = null;
 let explorer: CosmosDBExplorer;
 let lastCommand: MongoCommand;
+let lastOpenedDocument: DocDBDocumentNode;
+let lastOpenedDocumentType: string;
```
PrashanthCorp: Using a flag here. 
Doesn't look clean, but storing different variables for last Mongo db opened and last DocDB Document opened seemed typesafe to typescript.

src/extension.ts
```diff
@@ -61,18 +65,32 @@ export function activate(context: vscode.ExtensionContext) {
...
 
 	initCommand(context, 'cosmosDB.launchMongoShell', () => launchMongoShell());
-	initAsyncCommand(context, 'cosmosDB.openDocDBCollection', async (collection: DocDBCollectionNode) => {
-		util.showResult(JSON.stringify(await collection.getDocuments(), null, 2));
+	initAsyncCommand(context, 'cosmosDB.openDocDBDocument', async (document: DocDBDocumentNode) => {
+		const masterKey = await document.coll.db.getPrimaryMasterKey();
+		const endpoint = await document.coll.db.getEndpoint();
+		lastOpenedDocument = document;
+		util.showResult(JSON.stringify(document.data, null, 2));
+		lastOpenedDocumentType = "DocDB";
```
PrashanthCorp: By design this flag has to modify if and only if `util.showResult` is called. Is there a clean way to automatically enforce this?

src/commands.ts
```diff
@@ -343,6 +362,48 @@ export class CosmosDBCommands {
...
+        const client = new DocumentClient(endpoint, { masterKey: masterKey });
+        const editor = vscode.window.activeTextEditor;
+        const newdocument = JSON.parse(editor.document.getText());
+        const docLink = newdocument._self;
+        await new Promise((resolve, reject) => {
+            client.replaceDocument(docLink, newdocument,
+                { accessCondition: { type: 'IfMatch', condition: newdocument._etag } },
+                (err, updated) => {
+                    if (err) {
+                        reject(new Error(err.body));
```
EricJizbaMSFT: I assume this will show a readable message if the accessCondition fails?

src/extension.ts
EricJizbaMSFT: This is fine for now

src/extension.ts
EricJizbaMSFT: Use an enum instead. I don't like strings because typing errors can occur more easily

src/docdb/nodes.ts
```diff
@@ -101,11 +99,57 @@ export class DocDBCollectionNode implements INode {
...
 		let documents = await client.readDocuments(selfLink, { maxItemCount: 20 });
 		return await new Promise<any[]>((resolve, reject) => {
 			documents.toArray((err, docs: Array<Object>) => err ? reject(err) : resolve(docs));
 		});
 	}
 
+}
+
+export class DocDBDocumentNode implements INode {
+	data: IDocDBDocumentSpec;
```
EricJizbaMSFT: I don't like storing the data like this. We should dynamically retrieve the data every time the user clicks the collection. In addition, we shouldn't display '_rid', '_self', '_etag', etc. in result.json because we don't want the user to edit it

You don't have to fix in this PR, but please file an issue and fix this as highest priority after you merge the PR

src/extension.ts
```diff
@@ -61,18 +65,32 @@ export function activate(context: vscode.ExtensionContext) {
 	initAsyncCommand(context, 'cosmosDB.deleteMongoDB', (element: MongoDatabaseNode) => deleteDatabase(element));
 	initAsyncCommand(context, 'cosmosDB.deleteDocDBDatabase', (element: DocDBDatabaseNode) => CosmosDBCommands.deleteDocDBDatabase(element, explorer));
 	initAsyncCommand(context, 'cosmosDB.deleteDocDBCollection', (element: DocDBCollectionNode) => CosmosDBCommands.deleteDocDBCollection(element, explorer));
+	initAsyncCommand(context, 'cosmosDB.deleteDocDBDocument', (element: DocDBDocumentNode) => CosmosDBCommands.deleteDocDBDocument(element, explorer));
 	initCommand(context, 'cosmosDB.newMongoScrapbook', () => createScrapbook());
 	initCommand(context, 'cosmosDB.executeMongoCommand', () => lastCommand = MongoCommands.executeCommandFromActiveEditor(connectedDb));
-	initCommand(context, 'cosmosDB.updateMongoDocuments', () => MongoCommands.updateDocuments(connectedDb, lastCommand));
+	initCommand(context, 'cosmosDB.updateDocuments', () => {
```
EricJizbaMSFT: This should be initAsyncCommand and make sure to await updateDocDBDocument

src/commands.ts
PrashanthCorp: It says 'Conflict', followed by some JSON like message. 
Talked to Jing offline about formatting this error message, but that work item seems low priority.

src/commands.ts
PrashanthCorp: I see a '404, not found' error at a breakpoint ( `reject(err.body)` ) , but it doesn't bubble up. 
I'll discuss this offline.

EricJizbaMSFT: That's because you're not awaiting 'updateDocDBDocument' in extension.ts

src/commands.ts
```diff
@@ -235,6 +236,28 @@ export class CosmosDBCommands {
         }
     }
 
+    public static async createDocDBDocument(coll: DocDBCollectionNode, explorer: CosmosDBExplorer) {
+        const masterKey = coll.db.getPrimaryMasterKey();
+        const endpoint = coll.db.getEndpoint();
+        const client = new DocumentClient(endpoint, { masterKey: masterKey });
+        const docid = await vscode.window.showInputBox({
+            placeHolder: "Enter an unique id",
```
EricJizbaMSFT: This should be 'Enter _a_ unique id"

PR #105, opened 2017-10-20
Confirm and persist document when switching docs with un-persisted changes

PrashanthCorp:
Scenario : User switches documents (sometime accidentally). The hitherto current document they have been working has changes not persisted to azure. User would not want to lost work. 

Approach: Check if the current document state is different from the persisted state. Prompt for confirmation, and persist the changes if they so wish.


src/commands.ts
```diff
@@ -10,9 +10,10 @@ import { ResourceModels, ResourceManagementClient, SubscriptionClient, Subscript
 import DocumentdbManagementClient = require("azure-arm-documentdb");
 import docDBModels = require("azure-arm-documentdb/lib/models");
 import { DocumentClient } from 'documentdb';
+import { parse } from 'jsonc-parser';
```
PrashanthCorp: Currently, the only use case of this is to ignore comments in the JSON. JSON.parse does not work well with comments.

src/util.ts
```diff
@@ -49,10 +49,13 @@ export function errToString(error: any): string {
 	return error.toString();
 }
 
-export function showResult(result: string, column?: vscode.ViewColumn): Thenable<void> {
+export function showResult(result: string, filename: string, column?: vscode.ViewColumn): Thenable<void> {
```
PrashanthCorp: Change signature since we now show documents as document.json, and scrapbook results as result.json.

src/mongo/nodes.ts
```diff
@@ -211,11 +214,17 @@ export class MongoCollectionNode implements INode {
...
-		command: 'cosmosDB.openMongoCollection',
-		arguments: [this],
-		title: ''
-	};
+	readonly collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+
+	async getChildren(): Promise<INode[]> {
+		const command = MongoCommands.getCommand(`db.${this.label}.find()`);
+		const parentNode = this;
+		let result = await MongoCommands.executeCommand(command, this.db).then(documents => JSON.parse(documents));
```
EricJizbaMSFT: The first three lines in this function can be simplified to:
```
let result = JSON.parse(await this.find());
```

src/mongo/nodes.ts
```diff
@@ -211,11 +214,17 @@ export class MongoCollectionNode implements INode {
...
+		if (!Array.isArray(result)) {
+			result = [result];
+		}
+		return result.map(document => new MongoDocumentNode(document._id, parentNode, document));
```
EricJizbaMSFT: nit: I prefer not define single-use variables. In this case you could just use `this` instead of `parentNode`

src/util.ts
```diff
@@ -49,10 +49,13 @@ export function errToString(error: any): string {
...
 	let uri: vscode.Uri = null;
+	if (!filename.endsWith('.json')) {
```
EricJizbaMSFT: Remove this logic - we should always pass in a filename with '.json' if the file is supposed to end in 'json'

src/mongo/nodes.ts
PrashanthCorp: The functions find, insertOne, insertMany, etc. are all private. The way those methods and executeCommand are written seem to suggest that all the operations be done through the executeCommand method, which calls these private methods.

src/commands.ts
PrashanthCorp: Agreed. 
Should we be adding this behavior in case the user adds comments themselves? 
Currently, we would be throwing an error of the form : 
 'Unexpected token / in JSON at position ...'

EricJizbaMSFT: Personally, I don't think the user will add comments to the file and I don't think the added complexity of using a third party tool (and testing to make sure it works) is worth it.

But I see your point so it's up to you

src/mongo/nodes.ts
EricJizbaMSFT: As long as we're inside the class implementing the private functions, I think it's perfectly reasonable to use them

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+		return;
+	}
+	try {
+		recentDoc = (lastOpenedDocumentType === DocumentType.DocDB) ? lastOpenedDocDBDocument : lastOpenedMongoDocument;
+		oldDocument = recentDoc.data;
+		const editor = vscode.window.activeTextEditor;
+		newDocument = JSON.parse(editor.document.getText());
+	}
+	catch {
+		return;
```
EricJizbaMSFT: Why are we ignoring errors here?

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+	if (!_.isMatch(oldDocument, newDocument)) {
+		const confirmed = await vscode.window.showWarningMessage(`Your changes to  ${recentDoc.label}  will be lost. Update to Azure?`, "Yes", "No");
+		if (!confirmed) {
+			throw UserCancelledError;
+		}
+		else if (confirmed === "Yes") {
+			await update();
+			return;
+		}
+		else if (confirmed === "No") {
```
EricJizbaMSFT: Remove this last case - it doesn't do anything

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+			return;
+		}
+	}
+	return;
```
EricJizbaMSFT: Remove the extra `return;` calls - they don't do anything

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+	try {
+		recentDoc = (lastOpenedDocumentType === DocumentType.DocDB) ? lastOpenedDocDBDocument : lastOpenedMongoDocument;
+		oldDocument = recentDoc.data;
+		const editor = vscode.window.activeTextEditor;
+		newDocument = JSON.parse(editor.document.getText());
+	}
+	catch {
+		return;
+	}
+	if (!_.isMatch(oldDocument, newDocument)) {
```
EricJizbaMSFT: This code should be in the try block.

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+	if (lastOpenedDocumentType === DocumentType.Mongo) {
+		await MongoCommands.updateDocuments(connectedDb, lastCommand, lastOpenedMongoDocument);
+	}
+	else if (lastOpenedDocumentType === DocumentType.DocDB) {
+		await CosmosDBCommands.updateDocDBDocument(lastOpenedDocDBDocument);
+	}
+}
+
+async function updateOpenDocumentIfChanged(): Promise<void> {
+	let oldDocument, newDocument, recentDoc;
```
EricJizbaMSFT: I don't like declaring variables like this. If you move everything into the try block then you can see what the variable is and declare it inline:

```
const oldDocument = recentDoc.data;
```

src/extension.ts
```diff
@@ -127,6 +125,44 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+	if (lastOpenedDocumentType === undefined) {
+		return;
+	}
+	try {
+		recentDoc = (lastOpenedDocumentType === DocumentType.DocDB) ? lastOpenedDocDBDocument : lastOpenedMongoDocument;
```
EricJizbaMSFT: Can we rename `recentDoc` to `oldDocumentNode`

PR #111, opened 2017-10-24
Refactor DocDB commands into separate file

PrashanthCorp:
The main commands file was getting too large, and it made sense to compartmentalize the code at least wrt the DocDB code.


src/docdb/commands.ts
```diff
@@ -0,0 +1,208 @@
+import * as vscode from 'vscode';
```
EricJizbaMSFT: Add the copyright header

PR #112, opened 2017-10-25
Mongo : Create and Delete collections and documents

PrashanthCorp:
Bringing Mongo's implementation to parity with DocDB currently in master.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -3,11 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-/*
-Starting off with a small set of constants used in multiple places. This should hopefully become something akin to a resource file for localization.
-*/
-
 export class DialogBoxResponses {
-    static readonly inputYes: string = "Yes";
-    static readonly inputNo: string = "No";
+    static readonly Yes: string = "Yes";
+    static readonly No: string = "No";
 }
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/*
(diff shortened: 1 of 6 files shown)


src/mongo/commands.ts
```diff
@@ -78,6 +80,44 @@ export class MongoCommands {
 		}
 		return lastCommandOnSameLine || lastCommandBeforePosition || commands[commands.length - 1];
 	}
+
+	public static async createMongoCollection(db: MongoDatabaseNode, explorer: CosmosDBExplorer) {
+		const collectionName = await vscode.window.showInputBox({
+			placeHolder: "Enter name of collection",
```
EricJizbaMSFT: The placeholder between mongo and docdb should be consistent. I like this version better

PrashanthCorp:
Question : [Mongo]
It makes sense to place `createMongoDatabase` and `createMongoCollection` under `mongo/commands`

Do you agree?

EricJizbaMSFT:
Sounds fine to me

src/constants.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/*
+Starting off with a small set of constants used in multiple places. This should hopefully become something akin to a resource file for localization.
```
EricJizbaMSFT: I would remove this comment. Ideally we don't use the same strings over and over again, so we won't need 'something akin to a resource file'

Typescript localization looks like this: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/extension.ts#L30

It's actually better than a resource file because the english string is in-line with the code (and thus easier to read/understand)

src/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+*/
+
+export class DialogBoxResponses {
+    static readonly inputYes: string = "Yes";
```
EricJizbaMSFT: nit: Can this just be 'Yes'? I think the name of the class makes 'input' redundant

I would also make the name of the file 'DialogBoxResponses' to match the class name (The cosmosdb repo doesn't match the two very often, but it probably should)

PR #114, opened 2017-10-27
Load More functionality for DocDB documents

PrashanthCorp:


src/docdb/nodes.ts
```diff
@@ -152,4 +183,36 @@ export class DocDBDocumentNode implements INode {
...
+	readonly contextValue = 'LoadMoreButton'
+
+	readonly command: Command = {
+		command: 'cosmosDB.loadMore',
+		arguments: [this],
+		title: ''
+	};
+
+	static async loadNextKElements(iterator: QueryIterator<any>, k: number): Promise<any> {
+		let elements = [], i: number = 0, hasMoreItems: boolean = false;
```
EricJizbaMSFT: The pattern throughout our repos is to define each variable on a separate line

src/docdb/nodes.ts
```diff
@@ -152,4 +183,36 @@ export class DocDBDocumentNode implements INode {
...
+		let current = await new Promise<any>((resolve, reject) => iterator.nextItem((err, result) => err ? reject(err) : resolve(result)));
+		while (current !== undefined && i < k) {
+			elements.push(current);
+			i++;
+			current = await new Promise<any>((resolve, reject) => iterator.nextItem((err, result) => err ? reject(err) : resolve(result)));
+		}
+		if (current !== undefined) {
+			hasMoreItems = true;
```
EricJizbaMSFT: Remove `hasMoreItems`. Never used

src/docdb/nodes.ts
```diff
@@ -152,4 +183,36 @@ export class DocDBDocumentNode implements INode {
...
+			elements.push(iterator);
```
EricJizbaMSFT: That's weird to push 'iterator' onto elements (it's a different type, correct?). Hopefully when you move this to `DocDBCollectionNode` you won't have to do that (you could just store `iterator` as a private variable in the class)

src/docdb/nodes.ts
```diff
@@ -100,18 +106,43 @@ export class DocDBCollectionNode implements INode {
 	}
 
 	async getChildren(): Promise<INode[]> {
```
EricJizbaMSFT: You have too many functions (`getChildren`, `addChildrenFromDocuments`, `addMoreChildren` and `listDocuments`) that are too similar in functionality with too much copied code. Let's try to combine it into one or two functions. We can discuss offline if you need more details

src/docdb/nodes.ts
```diff
@@ -91,6 +93,10 @@ export class DocDBCollectionNode implements INode {
 		return this.db.getDbLink() + '/colls/' + this.id;
 	}
 
+	clearCache(): void {
+		this.children = null;
```
EricJizbaMSFT: I think your logic would be simpler if this set children to an empty array `[]` instead of null

src/docdb/nodes.ts
```diff
@@ -100,18 +106,43 @@ export class DocDBCollectionNode implements INode {
...
-		const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
-		let documents = await this.listDocuments(collLink, client);
-		return documents.map(document => new DocDBDocumentNode(document.id, parentNode, document));
+		if (!this.children) {
+			const collLink: string = this.getCollLink();
+			const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
+			let documents = await this.listDocuments(collLink, client);
+			this.addChildrenFromDocuments(documents);
+		}
+		return this.children ? this.children : [];
```
EricJizbaMSFT: Rather than pushing and popping the loadMore node all the time, never include it in children and instead just add it here:
```
return this.children.concat([this._loadMoreNode])
```

(Obviously you would have to save the loadMore node as a private variable called `_loadMoreNode` for that to work)

src/docdb/nodes.ts
```diff
@@ -91,34 +98,39 @@ export class DocDBCollectionNode implements INode {
...
-	async readOneCollection(selfLink, client): Promise<any> {
-		let documents = await client.readDocuments(selfLink, { maxItemCount: 20 });
-		return await new Promise<any[]>((resolve, reject) => {
-			documents.toArray((err, docs: Array<Object>) => err ? reject(err) : resolve(docs));
-		});
+		if (!this._hasFetched) {
+			const collLink: string = this.getCollLink();
+			const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
+			let docIterator = await client.readDocuments(collLink);
+			const elements = await LoadMoreNode.loadMore(docIterator, this._batchSize);
```
EricJizbaMSFT: Most of this is duplicated from 'addMoreChildren'

Just call 'addMoreChildren'

src/docdb/nodes.ts
```diff
@@ -67,13 +71,16 @@ export class DocDBDatabaseNode implements INode {
...
 export class DocDBCollectionNode implements INode {
 
 	constructor(readonly id: string, readonly db: DocDBDatabaseNode) {
 	}
 
 	readonly contextValue: string = "cosmosDBDocumentCollection";
+	private _children = [];
+	private _hasFetched: boolean = false;
+	private _loadMoreNode: LoadMoreNode = null;
+	private _batchSize: number = 20;
```
EricJizbaMSFT: Remove _batchSize. It's the same as the default in LoadMoreNode

src/docdb/nodes.ts
```diff
@@ -91,34 +98,39 @@ export class DocDBCollectionNode implements INode {
...
 	}
 
-	async getDocuments(): Promise<any> {
-		const dbLink: string = this.db.getDbLink();
-		const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
-		const collSelfLink = this.getCollLink();
-		const docs = await this.readOneCollection(collSelfLink, client);
-		return await docs;
+	clearCache(): void {
+		this._children = null;
```
EricJizbaMSFT: Assign to empty array instead

src/docdb/nodes.ts
```diff
@@ -152,4 +164,37 @@ export class DocDBDocumentNode implements INode {
...
+
+	readonly contextValue = 'LoadMoreButton'
+
+	readonly command: Command = {
+		command: 'cosmosDB.loadMore',
+		arguments: [this],
+		title: ''
+	};
+
+	static async loadMore(iterator: QueryIterator<any>, batchSize: number = 20): Promise<IResults> {
```
EricJizbaMSFT: I still feel this is a weird place to put this logic. I think the 'Load More' node should be a very simple wrapper for the 'cosmosDB.loadMore' command just like the SignInToAzureNode. And all state/functions should be moved to the DocDBCollectionNode node.

That being said, we're going to refactor anyways when we add LoadMore in other places - so you don't have to change for this PR if you don't want to.

src/docdb/nodes.ts
```diff
@@ -91,34 +98,39 @@ export class DocDBCollectionNode implements INode {
...
+			}
+			this._children = this._children.concat(documents.map(document => new DocDBDocumentNode(document.id, this, document)));
+			this._hasFetched = true;
+		}
+		return this._children.concat([this._loadMoreNode]);
+
+	}
+
+	async addMoreChildren(): Promise<void> {
+		const elements = await LoadMoreNode.loadMore(this._loadMoreNode.iterator, this._batchSize);
```
EricJizbaMSFT: It doesn't make sense to call a static method on LoadMoreNode with state from an instance of a LoadMoreNode. Either store the state somewhere else or make the method non-static and use the state directly (I prefer storing the iterator in DocDBCollectionNode)

src/docdb/nodes.ts
```diff
@@ -91,34 +98,39 @@ export class DocDBCollectionNode implements INode {
...
+			if (elements.hasMore) {
+				this._loadMoreNode = new LoadMoreNode(docIterator, this);
+			}
+			else {
+				this._loadMoreNode = null;
+			}
+			this._children = this._children.concat(documents.map(document => new DocDBDocumentNode(document.id, this, document)));
+			this._hasFetched = true;
+		}
+		return this._children.concat([this._loadMoreNode]);
```
EricJizbaMSFT: We don't want to concat a null loadMoreNode

src/docdb/nodes.ts
```diff
@@ -91,34 +99,38 @@ export class DocDBCollectionNode implements INode {
...
-		});
+		if (!this._hasFetched) {
+			const collLink: string = this.getCollLink();
+			const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
+			this.iterator = await client.readDocuments(collLink);
+			this._loadMoreNode = new LoadMoreNode(this);
+			await this.addMoreChildren(this.iterator);
+			this._hasFetched = true;
+		}
+		if (!this._hasMore) {
```
EricJizbaMSFT: nit:
```
return this._hasMore ? this._children.concat([this._loadMoreNode]) : this._children;
```

src/docdb/nodes.ts
```diff
@@ -91,34 +99,38 @@ export class DocDBCollectionNode implements INode {
...
-	async readOneCollection(selfLink, client): Promise<any> {
-		let documents = await client.readDocuments(selfLink, { maxItemCount: 20 });
-		return await new Promise<any[]>((resolve, reject) => {
-			documents.toArray((err, docs: Array<Object>) => err ? reject(err) : resolve(docs));
-		});
+		if (!this._hasFetched) {
+			const collLink: string = this.getCollLink();
+			const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
+			this.iterator = await client.readDocuments(collLink);
+			this._loadMoreNode = new LoadMoreNode(this);
```
EricJizbaMSFT: Don't assign to loadmore here

src/docdb/nodes.ts
```diff
@@ -91,34 +99,38 @@ export class DocDBCollectionNode implements INode {
...
-	async getDocuments(): Promise<any> {
-		const dbLink: string = this.db.getDbLink();
-		const client = new DocumentClient(this.db.getEndpoint(), { masterKey: this.db.getPrimaryMasterKey() });
-		const collSelfLink = this.getCollLink();
-		const docs = await this.readOneCollection(collSelfLink, client);
-		return await docs;
+	clearCache(): void {
+		this._children = [];
+		this._hasFetched = false;
+		this._loadMoreNode = null;
```
EricJizbaMSFT: Don't assign to loadmore here

src/docdb/nodes.ts
```diff
@@ -67,13 +71,17 @@ export class DocDBDatabaseNode implements INode {
...
 
 	constructor(readonly id: string, readonly db: DocDBDatabaseNode) {
 	}
 
 	readonly contextValue: string = "cosmosDBDocumentCollection";
+	private _children = [];
+	private _hasFetched: boolean = false;
+	private _loadMoreNode: LoadMoreNode = null;
+	private _hasMore: boolean;
+	private iterator = null;
```
EricJizbaMSFT: Don't set these values to null and make sure to define a type. For example, this should be:
```
private iterator: QueryIterator<any>;
```

src/docdb/nodes.ts
```diff
@@ -67,13 +71,17 @@ export class DocDBDatabaseNode implements INode {
 
 }
 
-
 export class DocDBCollectionNode implements INode {
 
 	constructor(readonly id: string, readonly db: DocDBDatabaseNode) {
 	}
```
EricJizbaMSFT: Just create the loadmore once in the constructor. Also you only need to pass in the id, not the whole node

```
this._loadMoreNode = new LoadMoreNode(id)
```

src/docdb/nodes.ts
```diff
@@ -91,34 +99,38 @@ export class DocDBCollectionNode implements INode {
...
+		if (!this._hasMore) {
+			return this._children;
+		}
+		else {
+			return this._children.concat([this._loadMoreNode]);
+		}
+
+	}
+
+	async addMoreChildren(iterator: any = this.iterator): Promise<void> {
```
EricJizbaMSFT: Why do you even have iterator as a parameter? Won't it always be this.iterator?

src/docdb/nodes.ts
```diff
@@ -91,34 +99,38 @@ export class DocDBCollectionNode implements INode {
...
+		const elements = await LoadMoreNode.loadMore(iterator);
+		const loadMoreDocuments = elements.results;
+		this._hasMore = elements.hasMore;
+		if (!this._hasMore) {
+			this._loadMoreNode = null;
```
EricJizbaMSFT: Don't assign to loadmore here

src/extension.ts
```diff
@@ -1,4 +1,4 @@
-/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT: Looks like a bad merge - add the comment back

src/docdb/nodes.ts
PrashanthCorp: The ID in itself may not be enough. I use a parentNode reference in the LoadMoreNode constructor. I'll still create it just once, just after constructor.

PR #119, opened 2017-10-31
DocumentDB creation/deletion errors - show full message in output window

### LINKED ISSUE #96: Parse Query Errors from DocDB node SDK
The current error object has a format requiring multiple JSON.parse operations on the object. Not high priority.

PrashanthCorp:
Fixes #96 
The activity ID is useful too, hence we are showing the whole error message in the output window.

### DIFF
--- a/src/ErrorData.ts
+++ b/src/ErrorData.ts
@@ -1,0 +1,32 @@
+import { error } from "util";
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+export class ErrorData {
+    public readonly message: string;
+    public readonly errorType: string;
+
+    constructor(error: any) {
+        if (error instanceof Error) {
+            try {
+                const parsed = JSON.parse(error.message);
+                this.errorType = parsed.Code || parsed.code;
+                this.message = parsed.Message || parsed.message;
+            } catch (err) {
+                this.errorType = error.constructor.name;
+                this.message = error.message;
+            }
+        } else if (typeof (error) === 'object' && error !== null) {
+            this.errorType = (<object>error).constructor.name;
+            this.message = JSON.stringify(error);
+        } else if (error !== undefined && error !== null && error.toString && error.toString().trim() !== '') {
+            this.errorType = typeof (error);
+            this.message = error.toString();
+        } else {
+            this.errorType = typeof (error);
+            this.message = 'Unknown Error';
+        }
+    }
+}
--- a/src/ErrorData.ts
+++ b/src/ErrorData.ts
@@ -1,5 +1,3 @@
-import { error } from "util";
-
 /*---------------------------------------------------------------------------------------------
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -112,6 +112,7 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
 		const start = Date.now();
 		let result = 'Succeeded';
 		let errorData: string = '';
+		const output = util.getOutputChannel();
 
 		try {
 			await callback(...args);
@@ -122,9 +123,17 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
 				result = 'Canceled';
 			}
 			else if (err instanceof Error) {
-				vscode.window.showErrorMessage(err.message);
+				output.appendLine(err.message);
+				if (JSON.parse(err.message).message.includes("\n") || JSON.parse(err.message).message.includes("\r")) {
+					output.show();
+					vscode.window.showErrorMessage('An error has occured. See output window for more details.');
+				}
+				else {
+					vscode.window.showErrorMessage(err.message);
+				}
 			}
 			else if (typeof err === "string") {
+				output.appendLine(err);
 				vscode.window.showErrorMessage(err);
 			}
 		} finally {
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -122,9 +123,17 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
 				result = 'Canceled';
 			}
 			else if (err instanceof Error) {
-				vscode.window.showErrorMessage(err.message);
+				output.appendLine(err.message);
+				if (JSON.parse(err.message).message.includes("\n") || JSON.parse(err.message).message.includes("\r")) {
```
EricJizbaMSFT: `\n` is used on all OS's, but `\r` isn't. For that reason, just check for `\n`

src/extension.ts
```diff
@@ -122,9 +123,17 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
...
+					output.show();
+					vscode.window.showErrorMessage('An error has occured. See output window for more details.');
+				}
+				else {
+					vscode.window.showErrorMessage(err.message);
+				}
 			}
 			else if (typeof err === "string") {
+				output.appendLine(err);
```
EricJizbaMSFT: This could be multiline as well

src/util.ts
```diff
@@ -49,6 +49,12 @@ export function errToString(error: any): string {
 	return error.toString();
 }
 
+const outputChannel = vscode.window.createOutputChannel("Azure CosmosDB");
```
EricJizbaMSFT: outputChannnel is a 'disposable' and should be added to context.subscriptions in extension.ts

src/extension.ts
EricJizbaMSFT: That's why we have the `ErrorData` class in the other repos (instead of 'util.errToString()') so that you can use the ErrorData.message property

src/util.ts
```diff
@@ -49,6 +49,12 @@ export function errToString(error: any): string {
 	return error.toString();
```
EricJizbaMSFT: Remove errToString - we should use ErrorData in all cases

src/extension.ts
```diff
@@ -110,26 +112,43 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
-			errorData = util.errToString(err);
+			properties.result = 'Failed';
 			if (err instanceof UserCancelledError) {
-				result = 'Canceled';
+				properties.result = 'Canceled';
 			}
 			else if (err instanceof Error) {
-				vscode.window.showErrorMessage(err.message);
+				errorData = new ErrorData(err);
+				output.appendLine(errorData.errorType);
```
EricJizbaMSFT: No need to output the errorType?

src/extension.ts
```diff
@@ -110,26 +112,43 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
 			await callback(...args);
 		} catch (err) {
-			result = 'Failed';
-			errorData = util.errToString(err);
+			properties.result = 'Failed';
 			if (err instanceof UserCancelledError) {
-				result = 'Canceled';
+				properties.result = 'Canceled';
 			}
 			else if (err instanceof Error) {
```
EricJizbaMSFT: This should be an 'else' case. All errors or strings or whatever should go through ErrorData

src/util.ts
EricJizbaMSFT: FYI - this comment is not high priority. App service does it this way, too, even though it's not technically correct

src/extension.ts
```diff
@@ -25,7 +26,8 @@ import MongoDBLanguageClient from './mongo/languageClient';
 import { Reporter } from './telemetry';
 import { UserCancelledError } from './errors';
 import { DocDBCommands } from './docdb/commands';
-import { DialogBoxResponses } from './constants'
+import { DialogBoxResponses } from './constants';
+import { error } from 'util';
```
EricJizbaMSFT: Why do you import error? I don't think that type exists?

src/ErrorData.ts
```diff
@@ -0,0 +1,32 @@
+import { error } from "util";
```
EricJizbaMSFT: Remove this import if it's not used, or move it below the copyright notice

src/extension.ts
PrashanthCorp: The DocDBErrorTypes seemed useful one word summaries, for eg., 'Conflict' . I can remove them if they are stylistically inconsistent.

src/extension.ts
PrashanthCorp: It seems to be a VSCode auto-completion thing. If I've used error as a variable anywhere and later renamed it, it imports the symbol it can't find in the file. 
I noticed it then, but didn't figure out the cause.

src/extension.ts
EricJizbaMSFT: It's not _always_ useful, though. For example, sometimes errorType is just 'string'

I have a slight preference not to print it. However, if you disagree I would at least do this instead:
```
output.appendLine(`Error type: ${errorData.errorType}`);
```

src/extension.ts
```diff
@@ -110,26 +111,42 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
+				output.appendLine(errorData.message);
+				if (errorData.message.includes("\n")) {
+					output.show();
+					vscode.window.showErrorMessage('An error has occured. See output window for more details.');
+				}
+				else {
+					vscode.window.showErrorMessage(errorData.message);
+				}
 			}
 			else if (typeof err === "string") {
```
EricJizbaMSFT: Remove this case - it never happens

src/extension.ts
```diff
@@ -110,26 +111,42 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
+		let errorData: ErrorData | undefined = null;
+		const output = util.getOutputChannel();
+		context.subscriptions.push(output);
 
 		try {
 			await callback(...args);
 		} catch (err) {
-			result = 'Failed';
-			errorData = util.errToString(err);
+			properties.result = 'Failed';
```
EricJizbaMSFT: Move `properties.result = 'Failed';` into the else case

Issue #68, opened 2017-09-18
Cannot see db results when no workspace is open


whiskeyjay:

PrashanthCorp:
Do we need to address this? We have a warning message for this reason...

EricJizbaMSFT:
Should be fixed with other save bugs

PR #120, opened 2017-11-01
Tie 'Save' action to 'Update to Azure' action

### LINKED ISSUE #109: CosmosDB update/persist : Tie 'Persist to Azure' to the save Action
Save file is instinctive to the user.
We would want to tie the persist to this action.

Approach: Listen to save file action and update document.

### LINKED ISSUE #118: updateOpenDocumentIfChanged fails if the current editor is not cosmos-document.json
Repro Steps:
1. Open a cosmos db document
1. Edit the document
1. Open any other file
1. Select a different cosmos db document in the tree
1. Select 'Yes' when it says "Your changes to {0} will be lost..."

Expected: It updates my doc to azure
Actual: It tries to upload the most recent file, which is _not_ a cosmos db document

EricJizbaMSFT:
1. Always open cosmos-document.json in an extension-specific directory
1. Dirty dot on file should match dirty state in Azure
1. Remove some unused code

Fixes #109 
Fixes #118

### DIFF
--- a/src/DocumentEditor.ts
+++ b/src/DocumentEditor.ts
@@ -1,0 +1,160 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { DialogBoxResponses } from './constants';
+import { UserCancelledError } from './errors';
+import { IDocumentNode } from './nodes';
+import * as util from './util';
+
+export class DocumentEditor implements vscode.Disposable {
+    private lastOpenedDocNode: IDocumentNode | undefined;
+    private localDocPath: string;
+    private localDocEditor: vscode.TextEditor | undefined;
+
+    private ignoreSave: boolean = false;
+    private isLocalDocOpen: boolean = false;
+
+    private recoveredDocsFolder: string;
+    private recoveredFileName: string | undefined;
+
+    private readonly dontShowKey: string = 'cosmosDB.dontShow.SaveEqualsUpdateToAzure';
+
+    constructor(context: vscode.ExtensionContext) {
+        // Use a workspace-specific path unique to our extension if possible. Otherwise, just use the main folder for our extension
+        this.localDocPath = path.resolve(path.join(context.storagePath || context.extensionPath, 'cosmos-document.json'));
+        this.recoveredDocsFolder = path.join(context.extensionPath, 'recoveredDocs');
+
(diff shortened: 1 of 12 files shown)


PrashanthCorp:
Does this address #68? If yes, I'll abandon my PR.

EricJizbaMSFT:
@PrashanthCorp  No - this does not fix 68. See here: https://github.com/Microsoft/vscode-cosmosdb/pull/120/files#diff-8cd4968b81985f0efc2053eabc37db6dR60

Notice the error is still being thrown

src/DocumentEditor.ts
```diff
@@ -0,0 +1,160 @@
...
+    private async onDidSaveTextDocument(globalState: vscode.Memento, doc: vscode.TextDocument): Promise<void> {
+        if (!this.ignoreSave && this.isLocalDocPath(doc) && this.lastOpenedDocNode) {
+            // soft-copy the node to avoid race conditions
+            const node: IDocumentNode = this.lastOpenedDocNode;
+
+            // If the user saved the file as a part of closing VS Code, we might not be able to 'Update to Azure' in time
+            // However, we can copy the file to this path and prompt user about it on next activation
+            this.recoveredFileName = `${this.lastOpenedDocNode.label}${Date.now().toString()}.json`;
+            try {
+                const dontShow: boolean | undefined = globalState.get(this.dontShowKey);
```
PrashanthCorp: Earlier in the file, dontShowKey seems to be a string [here](https://github.com/Microsoft/vscode-cosmosdb/pull/120/files#diff-0ed5637d7d8938c933072b40619ca513R25)
It feels a bit type-unsafe.

src/DocumentEditor.ts
```diff
@@ -0,0 +1,160 @@
...
+                if (dontShow !== true) {
```
PrashanthCorp: Is this different from if(!dontShow) ? I'm assuming you're checking for dontShow being false or undefined. 
Also, how is having dontShow be false different from undefined?

EricJizbaMSFT: There's not a big difference. I could do it either way, but I wanted to be explicit since we're getting the setting from globalState.

For example if the globalState somehow outputs a string "false", then `if(!dontShow)` would evaluate to true

src/DocumentEditor.ts
EricJizbaMSFT: That is correct. There are two separate variables. One is `dontShowKey` which is the key used in globalState

The other is `dontShow` which is the boolean value associated with the key

thirdpartynotices.txt
```diff
@@ -9,6 +9,7 @@ are grateful to these developers for their contribution to open source.
 2. MongoDB (https://github.com/mongodb/node-mongodb-native)
 3. copy-paste (https://github.com/xavi-/node-copy-paste)
 4. opn (https://github.com/sindresorhus/opn)
+5. fs-extra (https://github.com/jprichardson/node-fs-extra)
```
StephenWeatherford: Ask Matt to register...

EricJizbaMSFT: I've been doing the registrations myself. It's a super easy form

(I usually complete it after i merge the PR)

PR #121, opened 2017-11-01
Open result file in extension folder if workspace is undefined.

### LINKED ISSUE #68: Cannot see db results when no workspace is open

PrashanthCorp:
Fixes #68. 
Uses the extension's folder path as a stand-in if no workspace has been opened yet.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -51,14 +51,13 @@ export function errToString(error: any): string {
 
 export function showResult(result: string, filename: string, column?: vscode.ViewColumn): Thenable<void> {
 	let uri: vscode.Uri = null;
-	if (vscode.workspace.rootPath) {
-		uri = vscode.Uri.file(path.join(vscode.workspace.rootPath, filename));
+	const currExtensionPath = vscode.extensions.getExtension("ms-azuretools.vscode-cosmosdb").extensionPath;
+	const filepath = vscode.workspace.rootPath || currExtensionPath;
+	if (filepath) {
+		uri = vscode.Uri.file(path.join(filepath, filename));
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -51,14 +51,13 @@ export function errToString(error: any): string {
 
 export function showResult(result: string, filename: string, column?: vscode.ViewColumn): Thenable<void> {
 	let uri: vscode.Uri = null;
-	if (vscode.workspace.rootPath) {
-		uri = vscode.Uri.file(path.join(vscode.workspace.rootPath, filename));
+	const currExtensionPath = vscode.extensions.getExtension("ms-azuretools.vscode-cosmosdb").extensionPath;
```
PrashanthCorp: I'm not too comfortable about having the extension ID `ms-azuretools.vscode-cosmosdb` as a string. Ideally, I should be pulling this from the package.json file (publisher and name fields).

EricJizbaMSFT: You can pass context.extensionPath all the way from the activate() method in extension.ts. That way you don't even have to 'getExtension'

Something like: 
```
MongoCommands.executeCommandFromActiveEditor(connectedDb, context.extensionPath)
```
```
await util.showResult(result, 'result.json', activeEditor.viewColumn + 1, extensionPath);
```

src/util.ts
```diff
@@ -51,14 +51,13 @@ export function errToString(error: any): string {
...
+	const filepath = vscode.workspace.rootPath || currExtensionPath;
+	if (filepath) {
```
EricJizbaMSFT: This check shouldn't be necessary? The extensionPath should always exist

PR #122, opened 2017-11-01
Show error when copying connection string on linux without xclip

### LINKED ISSUE #64: 'Copy Connection String' doesn't work on linux
Same root problem as https://github.com/Microsoft/vscode-azure-account/issues/9

The copy-paste npm package we use requires xclip, but it's not installed on most linux OS's by default

EricJizbaMSFT:
We could potentially output the connection string to the output window so the user can manually copy it, but I would prefer not to do that with a sensitive string.

Plus, it's easy to install xclip

Fixes #64

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -12,6 +12,7 @@ import * as path from 'path';
 import * as copypaste from 'copy-paste';
 import * as opn from 'opn';
 import * as util from "./util";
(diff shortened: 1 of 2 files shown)


floatdrop:
Maybe it worth checking out [clipboardy](https://github.com/sindresorhus/clipboardy) - it contains xsel in package.

PR #130, opened 2017-11-02
Let Mongo auto-generate _id on new document

### LINKED ISSUE #124: id when creating new Mongo doc must be string of 12 bytes or 24 hex characters
Repro steps:
1. Open explorer on a mongo account
1. Create a new doc
1. Enter 'jing' as the id
1. Open the doc
1. Edit the doc
1. Update to Azure

Expected: Update succeeds
Actual: "Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters"

EricJizbaMSFT:
There's no need to ask the user for id. Mongo will generate the it for us.

Fixes #124

### DIFF
--- a/src/mongo/commands.ts
+++ b/src/mongo/commands.ts
@@ -94,8 +94,11 @@ export class MongoCommands {
 			placeHolder: "Enter a unique id for the document.",
 			ignoreFocusOut: true
 		});
-		await collectionNode.collection.insertOne({ "_id": docId });
-		explorer.refresh(collectionNode);
+
+		if (docId !== undefined) {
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
Mongo _has_ to autogenerate the '_id' field otherwise you get bug #124

The portal asks for an id, but it's not the field `_id`. If I specify 'test' as the id, I get this (note there's two separate fields):
screen shot 2017-11-01 at 5 36 55 pm

EricJizbaMSFT:
Okay now it should match portal behavior

PR #134, opened 2017-11-02
Add new documents to the explorer

### LINKED ISSUE #123: Newly created document does not show up in the explorer
Repro steps:
1. Expand a "DocumentDB" account in the explorer
1. Create a new Document

Expected:
1. New document is added to tree view

Actual:
1. New document is not added. However, it _does_ show after a refresh

EricJizbaMSFT:
These documents aren't in order, but I think it makes sense to be at the top since it's a new document. The next time the user does a hard refresh, it will show up in order.

Fixes #123

### DIFF
--- a/src/docdb/nodes.ts
+++ b/src/docdb/nodes.ts
@@ -119,6 +119,9 @@ export class DocDBCollectionNode implements INode {
 		this._children = this._children.concat(loadMoreDocuments.map(document => new DocDBDocumentNode(document.id, this, document)));
 	}
 
+	addNewDocToCache(document: any): void {
+		this._children.unshift(new DocDBDocumentNode(document.id, this, document))
+	}
 }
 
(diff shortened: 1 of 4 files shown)


src/docdb/nodes.ts
```diff
@@ -119,6 +119,9 @@ export class DocDBCollectionNode implements INode {
 		this._children = this._children.concat(loadMoreDocuments.map(document => new DocDBDocumentNode(document.id, this, document)));
 	}
 
+	addNewDocToCache(document: any): void {
```
PrashanthCorp: `addNewDocToCache` is just a line and used just once for each definition (in docdb and mongo). Would it make sense to just have the unshift operation just be inline?

EricJizbaMSFT: @PrashanthCorp I would prefer to keep this._children as private. That's the only reason I created this one-line public method

Issue #139, opened 2017-11-02
"Create New Collection" doesn't validate input?


whiskeyjay:
1. Right click on a (Mongo) database and then select "Create Collection".
2. In the InputBox type in an invalid name, like "!@)*#&!@#", then hit enter.

Expected:
InputBox validation will catch invalid names.

Actual:
Nothing happens (no error message, no new collection)

nturinski:
This applies to creating a new Document as well.  If you put invalid characters, the output pane appends a new blank line, but no error and no new document is created.

nturinski:
Nit: Apparently, Partition keys cannot start with numbers or any other symbol other than a alphabet character.  Not a big deal, but it'd be nice to see that before trying to create the collection

PrashanthCorp:
![image](https://user-images.githubusercontent.com/30420973/32353937-b44489d6-bfe4-11e7-8e18-ab2e3c0a785d.png)
Seems to fail a little gracefully in my use case.

PR #143, opened 2017-11-02
Validating inputs for Mongo Collection Name DocDB DocumentID

PrashanthCorp:
Modify the partition key check to search for illegal symbols anywhere in the key. 
Addresses #139

### DIFF
--- a/src/docdb/commands.ts
+++ b/src/docdb/commands.ts
@@ -119,7 +119,7 @@ export class DocDBCommands {
     }
 
     private static validatePartitionKey(key: string): string | undefined | null {
-        if (/^[#?\\]*$/.test(key)) {
+        if (/[#?\\]+/.test(key)) {
             return "Cannot contain these characters - ?,#,\\, etc."
         }
         return null;
--- a/src/docdb/commands.ts
+++ b/src/docdb/commands.ts
@@ -45,6 +45,7 @@ export class DocDBCommands {
         const client = new DocumentClient(endpoint, { masterKey: masterKey });
         const docid = await vscode.window.showInputBox({
             placeHolder: "Enter a unique id",
+            validateInput: DocDBCommands.validateDocumentID,
             ignoreFocusOut: true
         });
         const newDoc = await new Promise((resolve, reject) => {
@@ -116,6 +117,16 @@ export class DocDBCommands {
             return "Name has to be between 1 and 255 chars long";
         }
         return undefined;
+    }
+
+    private static validateDocumentID(name: string): string | undefined | null {
+        if (/^[-a-z0-9]+$/ig.test(name)) {
+            return "ID should contain only the following : digits, alphabets and -";
+        }
+        if (name.length > 40) {
+            return "The ID is too long";
+        }
+        return null;
     }
 
     private static validatePartitionKey(key: string): string | undefined | null {
--- a/src/mongo/commands.ts
+++ b/src/mongo/commands.ts
@@ -66,12 +66,20 @@ export class MongoCommands {
 	public static async createMongoCollection(db: MongoDatabaseNode, explorer: CosmosDBExplorer) {
 		const collectionName = await vscode.window.showInputBox({
 			placeHolder: "Enter name of collection",
+			validateInput: MongoCommands.validateCollectionName,
 			ignoreFocusOut: true
 		});
 		if (collectionName) {
 			await db.createCollection(collectionName);
 			explorer.refresh(db);
 		}
+	}
+
+	private static validateCollectionName(key: string): string | undefined | null {
+		if (/[#?\\\/\$]+/.test(key)) {
+			return "Cannot contain these characters - ?,#,\\,/, $ etc."
+		}
+		return null;
 	}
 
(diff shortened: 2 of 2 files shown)


src/docdb/commands.ts
```diff
@@ -118,8 +119,18 @@ export class DocDBCommands {
         return undefined;
     }
 
+    private static validateDocumentID(name: string): string | undefined | null {
+        if (/^[-a-z0-9]+$/ig.test(name)) {
+            return "ID should contain only the following : digits, alphabets and -";
+        }
+        if (name.length > 40) {
+            return "The ID is too long";
```
whiskeyjay: It would be better to explicitly say how many is too long.

src/docdb/commands.ts
```diff
@@ -118,8 +119,18 @@ export class DocDBCommands {
         return undefined;
     }
 
+    private static validateDocumentID(name: string): string | undefined | null {
+        if (/^[-a-z0-9]+$/ig.test(name)) {
+            return "ID should contain only the following : digits, alphabets and -";
+        }
+        if (name.length > 40) {
```
whiskeyjay: You are not checking lower boundary, does it mean empty strings are allowed as document IDs?

src/mongo/commands.ts
```diff
@@ -74,6 +75,13 @@ export class MongoCommands {
 		}
 	}
 
+	private static validateCollectionName(key: string): string | undefined | null {
+		if (/[#?\\\/\$]+/.test(key)) {
+			return "Cannot contain these characters - ?,#,\\,/, $ etc."
```
whiskeyjay: Does "etc." imply there are more invalid characters?

EricJizbaMSFT: '$' is a valid character. These are the only disallowed characters I get in the portal:
screen shot 2017-11-06 at 9 44 52 am 2

src/mongo/commands.ts
```diff
@@ -74,6 +75,13 @@ export class MongoCommands {
 		}
 	}
 
+	private static validateCollectionName(key: string): string | undefined | null {
```
EricJizbaMSFT: This exact same logic should apply to DocDB

src/docdb/commands.ts
EricJizbaMSFT: How did you get a limit of 40? I successfully created a docId of length 100

src/docdb/commands.ts
```diff
@@ -118,8 +119,18 @@ export class DocDBCommands {
         return undefined;
     }
 
+    private static validateDocumentID(name: string): string | undefined | null {
+        if (/^[-a-z0-9]+$/ig.test(name)) {
+            return "ID should contain only the following : digits, alphabets and -";
```
EricJizbaMSFT: This is not true. I successfully created a docId '$$$'

src/docdb/commands.ts
```diff
@@ -118,8 +119,18 @@ export class DocDBCommands {
...
+        if (name.length > 40) {
+            return "The ID is too long";
+        }
+        return null;
+    }
+
     private static validatePartitionKey(key: string): string | undefined | null {
-        if (/^[#?\\]*$/.test(key)) {
+        if (/[#?\\]+/.test(key)) {
             return "Cannot contain these characters - ?,#,\\, etc."
```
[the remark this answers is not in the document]
PrashanthCorp: I have only tested for the 3 characters. 
Trying to exhaustively find more rules was considered lower priority, and hence I moved on to other work items.

src/docdb/commands.ts
PrashanthCorp: Yes. It auto-generates an ID.

src/mongo/commands.ts
PrashanthCorp: [The mongo website](https://docs.mongodb.com/manual/reference/limits/#naming-restrictions) suggests otherwise

src/mongo/commands.ts
PrashanthCorp: There are mongo specified restrictions on names ([specified here for eg](https://docs.mongodb.com/manual/reference/limits/#naming-restrictions)), which could be different from Cosmos/DocumentDB's naming ([this page has some rules](https://docs.microsoft.com/en-us/azure/cosmos-db/create-documentdb-dotnet#add-a-collection)). I don't yet see why they would have the same logic on our side.

src/docdb/commands.ts
PrashanthCorp: I'll admit 40 was an estimate on the length of the hash it generated (including hyphens). I didn't test it completely. I'll remove the length check if I can't quickly determine a tight upper bound.

StephenWeatherford:
@PrashanthCorp Are you going to abandon this?

PrashanthCorp:
@StephenWeatherford  this was deemed lower priority. The current bugs in for 0.3.0 would touch on some of this code hopefully.

StephenWeatherford:
@PrashanthCorp If you're not merging this, please abandon it for now so it gets off our list.  You can always re-open later.

EricJizbaMSFT:
Closing this since it seems like a dupe of https://github.com/Microsoft/vscode-cosmosdb/pull/293

PR #147, opened 2017-11-03
Change versioning in package-lock

PrashanthCorp:
0.1.1 -> 0.2.0


whiskeyjay:
I thought this file can be excluded from check-in?

PR #148, opened 2017-11-03
DocDB createdocument : disallow empty docID

### LINKED ISSUE #142: Canceling "Create Document" creates a new document anyway
If you click "Create Document" from the context menu, it will create a new document with a random id anyway.  Same behavior as just hitting enter when you have not entered input yet.

PrashanthCorp:
Fixes #142

### DIFF
--- a/src/docdb/commands.ts
+++ b/src/docdb/commands.ts
@@ -43,21 +43,23 @@ export class DocDBCommands {
         const masterKey = coll.db.getPrimaryMasterKey();
         const endpoint = coll.db.getEndpoint();
         const client = new DocumentClient(endpoint, { masterKey: masterKey });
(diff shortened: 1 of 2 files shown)


src/docdb/commands.ts
```diff
@@ -43,21 +43,23 @@ export class DocDBCommands {
         const masterKey = coll.db.getPrimaryMasterKey();
         const endpoint = coll.db.getEndpoint();
         const client = new DocumentClient(endpoint, { masterKey: masterKey });
-        const docid = await vscode.window.showInputBox({
+        const docID = await vscode.window.showInputBox({
```
whiskeyjay: Use validateInput to check for empty string and specifically say something like "Empty string is not allowed".

PR #150, opened 2017-11-06
Allow empty strings for docID

PrashanthCorp:
Trim because ID's ending in space aren't allowed.

### DIFF
--- a/src/docdb/commands.ts
+++ b/src/docdb/commands.ts
@@ -43,21 +43,23 @@ export class DocDBCommands {
         const masterKey = coll.db.getPrimaryMasterKey();
         const endpoint = coll.db.getEndpoint();
         const client = new DocumentClient(endpoint, { masterKey: masterKey });
-        const docid = await vscode.window.showInputBox({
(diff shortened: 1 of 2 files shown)


src/docdb/commands.ts
```diff
@@ -43,16 +43,16 @@ export class DocDBCommands {
...
         const endpoint = coll.db.getEndpoint();
         const client = new DocumentClient(endpoint, { masterKey: masterKey });
-        const docID = await vscode.window.showInputBox({
+        let docID = await vscode.window.showInputBox({
             placeHolder: "Enter a unique id",
-            validateInput: DocDBCommands.validateDocumentName,
             ignoreFocusOut: true
         });
-        if (docID) {
+        if (docID || docID === "") {
```
EricJizbaMSFT: nit: `docID !== undefined` seems more readable to me

PrashanthCorp: I decided to be more explicit about the fact that I _am_ allowing "" too. 
I felt just leaving `docID !== undefined` would make me want to test what it does for null and other false-y values.

PR #151, opened 2017-11-06
Show output message on success

### LINKED ISSUE #146: 'Update to Azure' should show message on success
Right now the user has no way to know if it succeeded or failed

PrashanthCorp:
Fixes #146 .
The error cases get bubbled up from the scenarios I tested

### DIFF
--- a/src/DocumentEditor.ts
+++ b/src/DocumentEditor.ts
@@ -100,7 +100,12 @@ export class DocumentEditor implements vscode.Disposable {
     }
 
     private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
-        const updatedDoc: {} = await node.update(JSON.parse(doc.getText()));
+        const docFromEditor = JSON.parse(doc.getText());
+        const updatedDoc: {} = await node.update(docFromEditor);
+        const output = util.getOutputChannel();
+        output.appendLine(JSON.stringify(docFromEditor));
+        output.appendLine("Document update performed successfully. Document now persisted");
+        output.show();
         await this.updateEditor(updatedDoc);
     }
 
--- a/src/DocumentEditor.ts
+++ b/src/DocumentEditor.ts
@@ -100,11 +100,11 @@ export class DocumentEditor implements vscode.Disposable {
     }
 
     private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
-        const docFromEditor = JSON.parse(doc.getText());
-        const updatedDoc: {} = await node.update(docFromEditor);
+        const updatedDoc: {} = await node.update(JSON.parse(doc.getText()));
         const output = util.getOutputChannel();
-        output.appendLine(JSON.stringify(docFromEditor));
-        output.appendLine("Document update performed successfully. Document now persisted");
+        const timestamp = Math.floor(Date.now());
+        const docLink = node.getDocLink();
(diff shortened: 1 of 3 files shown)


src/DocumentEditor.ts
```diff
@@ -100,7 +100,12 @@ export class DocumentEditor implements vscode.Disposable {
     }
 
     private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
-        const updatedDoc: {} = await node.update(JSON.parse(doc.getText()));
+        const docFromEditor = JSON.parse(doc.getText());
+        const updatedDoc: {} = await node.update(docFromEditor);
+        const output = util.getOutputChannel();
+        output.appendLine(JSON.stringify(docFromEditor));
```
EricJizbaMSFT: Talked with @jinglouMSFT We don't want to display the document since it could be a really large file.

src/DocumentEditor.ts
```diff
@@ -100,7 +100,12 @@ export class DocumentEditor implements vscode.Disposable {
...
+        output.appendLine("Document update performed successfully. Document now persisted");
```
EricJizbaMSFT: Fix this message to include more info about the document and cosmos db account.

Here's my suggestion:
```
`${timestamp}: Updated document "${documentId}" to collection "${accountName}/${databaseName}/${collectionName}"`
```

PrashanthCorp: Addressed this in the next iteration : 
I'm using the docLink in place of the path. That seems more intuitive. 
Mongo : Using . instead of /  - makes it easier for the user to copy and paste into a shell for them to see what's happening (this.db.collection.document) 

Makes sense?

src/DocumentEditor.ts
```diff
@@ -101,6 +101,11 @@ export class DocumentEditor implements vscode.Disposable {
 
     private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
         const updatedDoc: {} = await node.update(JSON.parse(doc.getText()));
+        const output = util.getOutputChannel();
+        const timestamp = Math.floor(Date.now());
+        const docLink = node.getDocLink();
+        output.appendLine(`${timestamp}: Updated document "${node.id}" to collection "${docLink}"`);
```
EricJizbaMSFT: I'm fine with you using docLink, but then `node.id` is redundant information. Just do:
```
output.appendLine(`${timestamp}: Updated document "${docLink}"`);
```

src/DocumentEditor.ts
```diff
@@ -101,6 +101,11 @@ export class DocumentEditor implements vscode.Disposable {
 
     private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
         const updatedDoc: {} = await node.update(JSON.parse(doc.getText()));
+        const output = util.getOutputChannel();
+        const timestamp = Math.floor(Date.now());
```
EricJizbaMSFT: You should format the timestamp:
```
const timestamp = (new Date()).toLocaleTimeString();
```

EricJizbaMSFT: I would prefer if this was the timestamp from the document itself, but it's not a big deal

PrashanthCorp: Mongo documents don't typically have the time-stamp field. This code is in the DocumentEditor file, which is agnostic to DocDB/Mongo. So I don't think using time-stamp from the doc would be clean code.

PR #159, opened 2017-11-08
Preserve graph state

whiskeyjay:


src/graph/GraphViewServer.ts
```diff
@@ -22,6 +22,10 @@ export class GraphViewServer extends EventEmitter {
   private _httpServer: http.Server;
   private _port: number | undefined;
   private _socket: SocketIO.Socket;
+  private _lastQuery: string | undefined;
```
StephenWeatherford: Can we turn these into a single object with members?  Something like
interface ClientViewState {
  query?: string;
  data?: any[];
  errorMessage?: string
}

src/graph/client/graphClient.ts
```diff
@@ -131,14 +156,24 @@ export class GraphClient {
...
+  }
+
+  public setQuery(query: string) {
+    this.emitToHost('setQuery', query);
+  }
+
+  private setView() {
+    htmlElements.graphRadio.checked = this._graphView;
+    htmlElements.jsonRadio.checked = !this._graphView;
+    d3.select(htmlElements.graphSection).classed("active", this._graphView);
```
StephenWeatherford: nit: !! would make it more understandable

src/graph/GraphViewServer.ts
```diff
@@ -153,6 +162,24 @@ export class GraphViewServer extends EventEmitter {
...
+    console.log('getPageState');
+
+    if (this._lastQuery) {
+      this._socket.emit('setPageState', this._lastQuery, this._lastErrorMsg, this._lastResults, this._lastView);
+    }
+  }
+
+  private handleSetQuery(query: string) {
+    console.log('setQuery');
+    this._lastQuery = query;
```
StephenWeatherford: This means that if you query for results, wait for the results to show, then type a new query (or the start of a new query), but don't hit Execute, then tab away and back, the new query will show, and the old results will animate in, making it look like they're the new results.  Correct?

Perhaps you should throw away the last results (in the server, not the client) when the query changes?

src/graph/client/graphClient.ts
```diff
@@ -96,6 +102,21 @@ export class GraphClient {
...
+      if (!errorMsg) {
+        this.showResults(results);
+
+        if (view === 'json') {
+          this.selectJsonView();
+        } else {
+          this.selectGraphView();
+        }
+      } else {
+        this.setStateError(errorMsg);
```
StephenWeatherford: You should still set the view back to what the user had last selected, even if there's an error.

src/graph/client/graphClient.ts
```diff
@@ -122,6 +143,10 @@ export class GraphClient {
     });
   }
 
+  public getPageState() {
+    this.emitToHost('state');
```
StephenWeatherford: 'getPageState'

src/graph/client/graphClient.ts
```diff
@@ -131,14 +156,24 @@ export class GraphClient {
...
-    d3.select(htmlElements.jsonSection).classed("active", true);
+    this.setView();
+  }
+
+  public setQuery(query: string) {
+    this.emitToHost('setQuery', query);
+  }
+
+  private setView() {
+    htmlElements.graphRadio.checked = this._graphView;
```
StephenWeatherford: You should call selectGraphView or selectJsonView() instead.

whiskeyjay: Both select view functions call this.

resources/graphClient/graphClient.css
```diff
@@ -172,6 +172,11 @@ textarea {
     padding-top: 18%;
 }
 
+#stats {
```
StephenWeatherford: Careful on merge, this is not the current version of the code.

resources/graphClient/graphClient.html
```diff
@@ -63,6 +63,7 @@ <h1 id="title">
             <div id="resultsSection">
                 <div id="graphSection" class="active">
                     <!-- SVG will be inserted here -->
+                    <span id="stats"></span>
```
StephenWeatherford: Careful on merge, this is not the current version of the code.

src/graph/GraphViewServer.ts
```diff
@@ -22,9 +22,22 @@ export class GraphViewServer extends EventEmitter {
   private _httpServer: http.Server;
   private _port: number | undefined;
   private _socket: SocketIO.Socket;
+  private _previousPageState: {
+    lastQuery: string | undefined,
```
StephenWeatherford: nit: you could remove the "last" from all these

src/graph/client/graphClient.ts
```diff
@@ -15,6 +17,8 @@ declare let d3: any;
 const animationStepMs = 50; // TODO: optimize.  Slow down ticks?
 const graphWidth = 1200, graphHeight = 500; //TODO: be resizable or adapt to editor size
 const defaultQuery = "g.V()";
+const maxNodes = 1000;
```
StephenWeatherford: Careful on merge, this is not the current version of the code.

src/graph/client/graphClient.ts
```diff
@@ -122,6 +122,11 @@ export class GraphClient {
     this._socket.on('setPageState', (previousState) => {
       htmlElements.queryInput.value = previousState.lastQuery;
 
+      if (previousState.isQueryRunning) {
```
StephenWeatherford: Still need to set the correct tab being active

PR #160, opened 2017-11-08
Modify the update and delete operations to be partitionKey-aware

### LINKED ISSUE #141: Not able to delete documents
I am unable to delete documents from the extension.  I created a document from the extension and from the portal and both are giving me the same error (attached).
image
image

PrashanthCorp:
Fixes #141.


src/docdb/nodes.ts
```diff
@@ -133,6 +133,8 @@ export class DocDBDocumentNode implements IDocumentNode {
 
 	readonly contextValue: string = "cosmosDBDocument";
 
+	readonly partitionKeyValue = this.getPartitionKeyValue();
```
PrashanthCorp: Made it an accessible readonly variable because I'd ideally only want to compute the partitionKeyValue once, but access multiple times.

src/docdb/nodes.ts
```diff
@@ -166,7 +168,7 @@ export class DocDBDocumentNode implements IDocumentNode {
 		const _self: string = this.data._self;
 		this._data = await new Promise<IDocDBDocumentSpec>((resolve, reject) => {
 			client.replaceDocument(_self, newData,
-				{ accessCondition: { type: 'IfMatch', condition: newData._etag } },
+				{ accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this.partitionKeyValue || Object() },
```
PrashanthCorp: I realize that every time I access this.partitionKeyValue I'm also doing a falsey-check. Not placing the ( || Object() ) in the method since I don't know if _every_ API call needs an empty Object if partitionKeyValue is undefined/null

src/docdb/nodes.ts
```diff
@@ -178,4 +180,24 @@ export class DocDBDocumentNode implements IDocumentNode {
...
+			return null;
+		}
+		const fields = partitionKey.paths[0].split('/');
+		if (fields[0] === '') {
+			fields.shift();
+		}
+		let value;
+		for (let field of fields) {
+			value = value ? value[field] : this.data[field];
+			if (!value) {
```
PrashanthCorp: Looping is similar to the LoadMore() method. 
Break if at any point, the field is null/undefined (implying that the partitionKey path doesn't exist in that document).

PR #161, opened 2017-11-09
After deleting a node successfully, remove the node from the cache.

### LINKED ISSUE #136: Deleted document not automatically removed from DB Explorer
1. Create a new document.
2. Double click the document to open it.
3. Right click the document in the tree and select "Delete Document"
4. Click Yes to confirm deletion.

Expected:
The document should be deleted and the tree item removed from DB explorer.

Actual:
The document is deleted but the TreeView not refreshed, need to manually hit refresh then the document will go away from the view.

PrashanthCorp:
Fixes #136

### DIFF
--- a/src/docdb/nodes.ts
+++ b/src/docdb/nodes.ts
@@ -123,6 +123,10 @@ export class DocDBCollectionNode implements INode {
 	addNewDocToCache(document: any): void {
 		this._children.unshift(new DocDBDocumentNode(document.id, this, document))
 	}
+
+	removeNodeFromCache(documentNode: DocDBDocumentNode): void {
+		this._children = this._children.filter(doc => doc.id !== documentNode.id);
+	}
 }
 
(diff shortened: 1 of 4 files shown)


src/docdb/nodes.ts
```diff
@@ -123,6 +123,10 @@ export class DocDBCollectionNode implements INode {
 	addNewDocToCache(document: any): void {
 		this._children.unshift(new DocDBDocumentNode(document.id, this, document))
 	}
+
+	removeNodeFromCache(documentNode: DocDBDocumentNode): void {
+		this._children = this._children.filter(doc => doc.id !== documentNode.id);
```
EricJizbaMSFT: This should work, but the normal pattern is to use the 'splice' function to remove from an array:
```
const index: number = this._children.findIndex(doc => doc.id !== documentNode.id);    
if (index !== -1) {
    array.splice(index, 1);
}
```

PR #163, opened 2017-11-09
Add right-click menu commands to graph nodes

EricJizbaMSFT:
This brings it up to parity with DocumentDB nodes:
1. Open in Portal
1. Refresh
1. Create/Delete database
1. Create/Delete graph

I think there's more code clean-up/organization that we could do, but I'll save that until after the release

### DIFF
--- a/src/docdb/commands.ts
+++ b/src/docdb/commands.ts
@@ -11,9 +11,10 @@ import { DocumentClient } from 'documentdb';
 import { DocumentBase } from 'documentdb/lib';
 import { CosmosDBExplorer } from './../explorer';
 import { DialogBoxResponses } from '../constants'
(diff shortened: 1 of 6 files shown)


src/docdb/commands.ts
```diff
@@ -65,14 +66,20 @@ export class DocDBCommands {
     }
 
 
-    public static async createDocDBCollection(db: DocDBDatabaseNode, explorer: CosmosDBExplorer) {
+    public static async createCollection(db: DocDBDatabaseNode | GraphDatabaseNode, explorer: CosmosDBExplorer) {
+        let placeHolder: string;
+        if (db instanceof GraphDatabaseNode) {
```
PrashanthCorp: Shouldn't this be the other way around? 
If the db is a GraphDatabaseNode I ask for name of a graph, 
else I ask for name of a collection.

PR #171, opened 2017-11-10
Don't show results from edges query in JSON

StephenWeatherford:


src/graph/client/graphClient.ts
```diff
@@ -296,7 +311,10 @@ export class GraphClient {
       // Limit number of edges (done after determining which edges are still valid based on reduced vertex set)
       links = links.slice(0, maxEdges);
 
-      d3.select(htmlElements.stats).text(`Displaying ${nodes.length} of ${vertices.length} vertices and ${links.length} of ${edges.length} edges`);
+      var statsText: string = (nodes.length === vertices.length && links.length === edges.length) ?
+        `Displaying ${nodes.length} vertices and ${links.length} edges` :
+        `Displaying ${nodes.length} of ${vertices.length} vertices and ${links.length} of ${edges.length} edges`;
```
EricJizbaMSFT: Why this change? I prefer the original text since the user probably doesn't know what our limit is (and thus doesn't know if they got back all results)

I suppose you could change it to 'Displaying **_all_** ${nodes.length} vertices and ${links.length} edges' and I would be happy

StephenWeatherford: done

PR #182, opened 2017-11-10
Fix can't open documents in explorer

EricJizbaMSFT:
Any time you expand a DocumentDB collection, you got an error like this:
screen shot 2017-11-09 at 10 01 31 pm

It's because this.getPartitionKeyValue() relies on this.data, which wasn't set yet


EricJizbaMSFT:
Added another fix for 'Update to Azure' after switching tabs
1. textEditors get disposed when the user switches tabs
1. The onDidCloseTextDocument event also occurs when the user switches tabs
Instead, cache the document, which does _not_ get disposed

PR #185, opened 2017-11-10
Keywords

StephenWeatherford:


package.json
```diff
@@ -16,6 +16,14 @@
 	"categories": [
 		"Azure"
 	],
+	"keywords": [
+		"Azure",
+		"CosmosDB",
```
EricJizbaMSFT: The official product name has a space between Cosmos and DB: https://azure.microsoft.com/en-us/services/cosmos-db/
Not sure if that matters


(MongoDB however, does NOT have a space. It's confusing. https://www.mongodb.com/)

PR #193, opened 2017-11-13
Set up tests

StephenWeatherford:


.travis.yml
```diff
@@ -1,15 +1,30 @@
...
+before_install:
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CXX="g++-4.9" CC="gcc-4.9" DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
+      sleep 3;
+    fi
+
 install:
   - npm install
+  - npm run vscode:prepublish
```
EricJizbaMSFT: This does the same as `tsc -p ./` on line 24. (That being said, I would prefer you overwrite that with this `npm run` version)

.travis.yml
```diff
@@ -1,15 +1,30 @@
...
 
 script:
   - tsc -p ./
+  - npm test --silent
```
EricJizbaMSFT: Why silent?

StephenWeatherford: That's what they had in https://code.visualstudio.com/docs/extensions/testing-extensions.  Will change.

.travis.yml
StephenWeatherford: @EricJizbaMSFT You mean this?

install:
  - npm install

script:
  - npm run vscode:prepublish
  - npm test

PR #195, opened 2017-11-13
Dedup vertices in graph Fixes #190

### LINKED ISSUE #190: Need to merge duplicate vertices in graph
Some queries can result in listing the same vertex twice (lame but valid example: "g.V().repeat(both('knows')).times(2)").  Should not display those vertices multiple times in the graph.

StephenWeatherford:


src/graph/client/graphClient.ts
```diff
@@ -355,29 +355,48 @@ export class GraphClient {
...
-        .slice(0, maxNodes) // Limit # of nodes
+      var allNodes: ForceNode[] = vertices
         .map(v => <ForceNode>{ vertex: v });
-      var links: ForceLink[] = [];
 
+      // Create map of nodes by ID, and remove duplicates
       var nodesById = new Map<string, ForceNode>();
-      nodes.forEach(n => nodesById.set(n.vertex.id, n));
+      var uniqueNodes: ForceNode[] = [];
+      allNodes.forEach(n => {
```
EricJizbaMSFT: Why don't you just use `vertices` directly?

src/graph/client/graphClient.ts
```diff
@@ -355,29 +355,48 @@ export class GraphClient {
...
+        var id = n.vertex.id;
+        if (!nodesById.has(id)) {
+          nodesById.set(id, n);
+          uniqueNodes.push(n);
+        }
+      });
+
+      var nodes = uniqueNodes;
```
EricJizbaMSFT: Why don't you just use uniqueNodes directly?

src/graph/client/graphClient.ts
StephenWeatherford: Nodes are a wrapper around vertices

PR #205, opened 2017-11-15
Vertex map should only contain vertices not truncated by max

StephenWeatherford:
This was causing errors displaying some large graphs


StephenWeatherford:
Moving this logic to the server side...

PR #206, opened 2017-11-15
Don't re-use gremlin query for edges, fixes #184

### LINKED ISSUE #184: g.addV('test') query adds two vertices
If you add a vertex through the query executing, it will add two points to your graph.

1.) g.addV('test').property('firstName', 'Jack')
2.) g.V().hasLabel('test')

The results has two vertices with the label "test"

StephenWeatherford:
1) Moved the deduping and size limiting to the server
2) Now querying specifically for the IDs of the vertices from the gremlin query
  a) Size we're limiting size in server, we only query up to max vertices
  b) Nevertheless we still need to chunk up the query
  c) Added retry logic for RequestRateTooLarge, which can happen because we're querying for a lot of edges in multiple queries in parallel
3) This cleans up and simplified the client-size code
4) Also means the edge query should run much faster for many cases


src/graph/GraphViewServer.ts
```diff
@@ -11,12 +11,45 @@ import * as io from 'socket.io';
...
+interface Edge {
+  id: string;
+  type: "edge";
+  inV: string;  // Edge source ID
+  outV: string; // Edge target ID
+};
+
+interface Vertex {
+  id: string;
+  type: "edge";
```
EricJizbaMSFT: Is a vertex really supposed to be type edge?

src/graph/GraphViewServer.ts
```diff
@@ -114,14 +146,29 @@ export class GraphViewServer extends EventEmitter {
...
 
-      // If it returned any vertices, we need to also query for edges
-      if (queryResults.find(v => v.type === "vertex")) {
+      // Full query results - may contain vertices and/or edges and/or other things
+      var fullResults = await this.executeQuery(queryId, gremlinQuery);
+
+      let vertices = this.getVertices(fullResults);
+      let { limitedVertices, countUniqueVertices } = this.limitVertices(vertices);
+      results = {
+        fullResults,
```
EricJizbaMSFT: Should this be:
```
fullResults: fullResults,
```

src/graph/GraphViewServer.ts
```diff
@@ -150,13 +263,41 @@ export class GraphViewServer extends EventEmitter {
...
+        if (iTry > 1) {
+          this.log(`Retry #${iTry - 1} for query ${queryId}: ${truncateQuery(gremlinQuery)}`);
+        }
+        return await this._executeQueryCore(queryId, gremlinQuery);
+      } catch (err) {
+        if (this.isErrorRetryable(err)) {
+          if (iTry >= maxRetries) {
+            this.log(`Max retries reached for query ${queryId}: ${truncateQuery(gremlinQuery)}`);
+          } else {
+            continue;
```
EricJizbaMSFT: If I'm getting 'RequestRateTooLarge' errors - will trying again immediately work most of the time? Or do we need to wait a few seconds before trying again?

src/graph/client/graphClient.ts
```diff
@@ -65,29 +60,36 @@ function logToUI(s: string) {
...
+  id: string;
+  type: "edge";
   inV: string;  // Edge source ID
   outV: string; // Edge target ID
 };
 
-interface ResultVertex extends ResultNode {
+interface Vertex {
+  id: string;
+  type: "edge";
```
EricJizbaMSFT: type vertex?

test/graphClient.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import * as vscode from 'vscode';
+import * as myExtension from '../src/extension';
+
+import * as graphClient from "../src/graph/client/graphClient";
+
+// Defines a Mocha test suite to group tests of similar kind together
+suite("Graph Client Tests", () => {
+
+    // Defines a Mocha unit test
+    test("Something 1", () => {
```
EricJizbaMSFT: Not sure what this file is about?

src/graph/GraphViewServer.ts
StephenWeatherford: Thanks.

src/graph/GraphViewServer.ts
StephenWeatherford: Theoretically the time to wait is in the returned data, but it's embedded inside the error message. Not sure that I want to pick it out.  So far just the latency of trying again seems to be enough delay, but yeah I'll probably add some specific delay later.

test/graphClient.test.ts
StephenWeatherford: Thought I'd deleted that.

src/graph/GraphViewServer.ts
StephenWeatherford: That's actually valid.

PR #213, opened 2017-11-16
Share types between serve and client, add type safety to messages

StephenWeatherford:
(For after Thursday release)

added graphTypes.ts.

As long as it's kept to just types and interfaces, they will be visible to the entire project and thus shared between client and server.


StephenWeatherford:
Good point, I’ll change it.

Issue #198, opened 2017-11-14
Clicking on collection no longer opens the results in a json file


johnpapa:
This used to work prior to v 0.2.0. It is quite convenient so we can see the json contents. Otherwise, we have to create a scrapbook and type the mongo commands.

This is a regression from the previous version.

PrashanthCorp:
Fixed with [this PR](https://github.com/Microsoft/vscode-cosmosdb/pull/219).

PrashanthCorp:
Our new release now has this functionality added back. 
Have a look [here](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb)

Issue #192, opened 2017-11-12
why is update result removed?


TinusScheppers:
From the changelog:
`Removed`

`View Mongo documents in 'result.json' by clicking on a collection in the explorer`
`Edit and persist the result of a Mongo scrapbook command`

Why have these key features been removed?
I used the update documents from the result set of a query very often, why is it removed all of a sudden?

Kind regards
Tinus

PrashanthCorp:
Thanks for the feedback! We didn't estimate the amount of user feedback this feature receive. We shall work on trying to re-incorporate it in the current user experience. 
Currently, clicking on a collection opens up a treeview of it's documents, hence there's a conflict from a user experience. That was the rationale behind replacing the previous UX.

TinusScheppers:
Hi, that sounds great!
Maybe its a possibility to add a right click action to the collection?

PrashanthCorp:
Fixed with [this PR](https://github.com/Microsoft/vscode-cosmosdb/pull/219).

PrashanthCorp:
That's what we did. :-) 

Our new release now has this functionality added back. 
Have a look [here](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb)

Issue #191, opened 2017-11-11
command 'cosmosDB.updateMongoDocuments' not found


ticruz38:
I used to manually update my collections in test mode and used this command a lot, it seems like it's not available in the new version anymore.
Is there any workaround for me to be able to update my mongo collection again?

Thanks

PrashanthCorp:
As for a workaround: Currently you can update each document in the document view mode.

Thanks for the feedback! We didn't estimate the amount of user feedback this feature receive. We shall work on trying to re-incorporate it in the current user experience (currently, clicking on a collection opens up a treeview of it's documents, hence there's a conflict from a user experience).

ticruz38:
Yeah, I attached a keymap to it so this feature made the extension very very helpful, I will be downgrading as long as this feature is missing in the upgrade.
Thank you, looking forward further improvement

PrashanthCorp:
Fixed with [this PR](https://github.com/Microsoft/vscode-cosmosdb/pull/219).

PrashanthCorp:
Our new release now has this functionality added back. 
Have a look [here](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb)

PR #219, opened 2017-11-17
Feature return - View scrapbook results and mongo collection views

PrashanthCorp:
- Re-allow collection views in `mongo` (not yet `docDB`)
- Incorporate select scrapbook results (`find` and `findOne` cases) into documentEditor.
- Use a dummy node (mongo dummy node)  for the above incorporation to decrease the code change in `DocumentEditor`
- Filenames based on result type: document, collection, or scrapbook result


src/nodes.ts
```diff
@@ -20,10 +20,11 @@ import { GraphDatabaseNode } from './graph/graphNodes';
 
 type Experience = "MongoDB" | "DocumentDB" | "Graph" | "Table";
 
-export interface IDocumentNode extends INode {
+export interface IDocument {
```
PrashanthCorp: Minor stylistic change subject to modification - 
Decouple the `IDocumentNode` from `INode`. This makes it more clear that `DocumentEditor` does not need a `vscode.TreeItem` structure to operate on. Moreover, a previous implementation with the `IDocumentNode` extending `INode` left unused properties in `MongoDummyNode` just to satisfy interface implementation.

src/mongo/nodes.ts
```diff
@@ -196,6 +226,42 @@ export class MongoCollectionNode implements INode {
...
+		if (!Array.isArray(data)) {
+			data = [data];
+		}
+		const operations = MongoCollectionNode.getBulkWriteUpdateOperations(data);
+		const result = await this.collection.bulkWrite(operations);
+		this._data = data;
+		return this._data;
+	}
+
+	static getBulkWriteUpdateOperations(data: any): any {
```
PrashanthCorp: static since it's also used by `MongoDummyNode`. Any suggestions for a better place to put this function?

src/mongo/commands.ts
```diff
@@ -30,7 +31,15 @@ export class MongoCommands {
...
 			const result = await database.executeCommand(command);
-			await util.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
+			if (command.name === 'find' || command.name === 'findOne') {
+				const db = await database.getDb();
+				const dummy = new MongoDummyNode(db.collection(command.collection), command.name);
+				dummy.data = JSON.parse(result);
+				await editor.showDocument(dummy);
+			}
+			else {
+				await util.showNewFile(result, extensionPath, 'result-mongo-ouput', '.json', activeEditor.viewColumn + 1);
```
PrashanthCorp: I felt a different name would indicate that it is not persistable. Also helps package.json correctly decide to (not) allow the update command

src/extension.ts
```diff
@@ -57,8 +57,12 @@ export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(azureAccount.onSessionsChanged(() => explorer.refresh()));
 	context.subscriptions.push(vscode.window.registerTreeDataProvider('cosmosDBExplorer', explorer));
 
-	const documentEditor: DocumentEditor = new DocumentEditor(context);
-	context.subscriptions.push(documentEditor);
+	const documentEditor: DocumentEditor = new DocumentEditor(context, 'cosmos-document.json');
+	const collectionEditor: DocumentEditor = new DocumentEditor(context, 'cosmos-collection.json');
+	const resultEditor: DocumentEditor = new DocumentEditor(context, 'cosmos-result.json');
```
PrashanthCorp: Have precisely one file for the resultEditor. 
Regression from previous implementation - can no longer have multiple result.json files which can be persisted to azure.

package.json
```diff
@@ -253,6 +256,24 @@
 					"dark": "/resources/icons/theme-agnostic/UpdateOutline.svg"
 				}
 			},
+			{
+				"category": "CosmosDB",
+				"command": "cosmosDB.updateMongoCollection",
```
EricJizbaMSFT: Can this be 'Update Collection to Cloud'? The users can attach a mongo database from any cloud

package.json
```diff
@@ -262,6 +283,11 @@
 				"category": "CosmosDB",
 				"command": "cosmosDB.openDocument",
 				"title": "Open Document"
+			},
+			{
+				"category": "CosmosDB",
+				"command": "cosmosDB.openCollection",
+				"title": "Open Loaded Documents"
```
EricJizbaMSFT: I would stick with 'Open Collection'

package.json
```diff
@@ -273,13 +299,31 @@
 				{
 					"command": "cosmosDB.update",
 					"when": "resourceFilename==cosmos-document.json"
+				},
+				{
+					"command": "cosmosDB.updateMongoCollection",
+					"when": "resourceFilename==cosmos-document.json"
```
EricJizbaMSFT: This should be cosmos-collection.json

src/mongo/commands.ts
```diff
@@ -30,7 +31,15 @@ export class MongoCommands {
 				throw new Error('Please connect to the database first');
 			}
 			const result = await database.executeCommand(command);
-			await util.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
+			if (command.name === 'find' || command.name === 'findOne') {
+				const db = await database.getDb();
+				const dummy = new MongoDummyNode(db.collection(command.collection), command.name);
```
EricJizbaMSFT: Remove the MongoDummyNode class and just do use the existing class:
```
MongoCollectionNode(db.collection(command.collection), db)
```

src/mongo/nodes.ts
EricJizbaMSFT: Leave it here and remove MongoDummyNode

src/nodes.ts
EricJizbaMSFT: This is unnecessary if you remove MongoDummyNode

src/mongo/commands.ts
EricJizbaMSFT: I would leave the name as it was. This feels too long

src/extension.ts
EricJizbaMSFT: I still think there should be only one DocumentEditor. However, inside of DocumentEditor, there should be a map of:
filepath -> CollectionNode | DocumentNode

If the filePath is a new temporary folder each time, then we don't have to limit the user to one result.json file

package.json
```diff
@@ -273,13 +279,22 @@
 				{
 					"command": "cosmosDB.update",
 					"when": "resourceFilename==cosmos-document.json"
+				},
```
EricJizbaMSFT: Right now there's no behavior difference between `cosmos-document.json` and `cosmos-collection.json`. Let's just combine them into one. Maybe `cosmos-result.json` or `cosmos-editor.json`?

src/DocumentEditor.ts
```diff
@@ -4,154 +4,79 @@
...
-                    }
-                } else if (result === deleteFile) {
-                    await fse.unlink(recoveredFilePath);
-                }
-            }
-        }
-    }
-
-    private async udpateDocumentToNode(node: IDocumentNode, doc: vscode.TextDocument): Promise<void> {
+    private async udpateDocumentToNode(node: IDocument, doc: vscode.TextDocument): Promise<void> {
```
EricJizbaMSFT: Rename to `updateToCloud`

src/DocumentEditor.ts
```diff
@@ -4,154 +4,79 @@
...
     }
 
     public async dispose(): Promise<void> {
-        if (this.recoveredFileName) {
-            await fse.ensureDir(this.recoveredDocsFolder);
-            await fse.move(this.localDocPath, path.join(this.recoveredDocsFolder, this.recoveredFileName));
-        } else {
-            await fse.unlink(this.localDocPath);
-        }
+        Object.keys(this.fileMap).forEach(async (key) => await fse.unlink(key));
```
EricJizbaMSFT: Did you figure out how to fix the error message when you close and re-open VS Code? If not, file an issue so we don't forget

src/DocumentEditor.ts
```diff
@@ -4,154 +4,79 @@
...
+        await this.updateEditor(docNode.data, textEditor);
     }
 
-    public async updateLastDocument(): Promise<void> {
-        if (this.lastOpenedDocNode && this.localDoc) {
-            await this.udpateDocumentToNode(this.lastOpenedDocNode, this.localDoc);
-        } else {
-            throw new Error('You must select a Document in the CosmosDB explorer before updating to Azure.');
-        }
+    public async updateLastDocument(doc): Promise<void> {
```
EricJizbaMSFT: Rename to `updateMatchingNode` or something. It's not the 'last document' anymore

package.json
PrashanthCorp: I tried that. Looking at the use of `&&` in the view/title clauses suggested it would work, but it didn't. :(

EricJizbaMSFT: I'm not talking about combining the _where_ clause, I'm talking about combining the file names. You would only have one file with one update command:
```json
"editor/title": [
  {
    "command": "cosmosDB.update",
    "when": "resourceFilename==cosmos-editor.json",
    "group": "navigation"
  },
]
```

src/mongo/nodes.ts
```diff
@@ -185,9 +185,12 @@ export class MongoDatabaseNode implements INode {
 	}
 }
 
-export class MongoCollectionNode implements INode {
+export class MongoCollectionNode implements IEditableNode {
 
-	constructor(readonly collection: Collection, readonly db: MongoDatabaseNode) {
+	constructor(readonly collection: Collection, readonly db: MongoDatabaseNode, data: Array<any>, readonly query: string) {
+		data.forEach(element => {
```
EricJizbaMSFT: If you pass in data, then 'Load more...' won't work because the iterator isn't set. Instead, don't pass in data (don't even execute the command for `find`) and just call node.getChildren to fetch the first batch

(Down the line we might add a 'Load More...' right click menu to each file - in which case the iterator needs to be set properly)

src/DocumentEditor.ts
PrashanthCorp: Filed one.

src/mongo/commands.ts
```diff
@@ -30,7 +31,21 @@ export class MongoCommands {
 				throw new Error('Please connect to the database first');
 			}
 			const result = await database.executeCommand(command);
```
EricJizbaMSFT: Don't run executeCommand for a 'find'. Just run it in the node itself as a part of getChildren

src/mongo/nodes.ts
```diff
@@ -197,6 +200,47 @@ export class MongoCollectionNode implements INode {
...
+						"upsert": false
+					}
+			};
+			operationsArray.push(operation);
+		}
+		return operationsArray;
+	}
+
+	findDocById(id: string): MongoDocumentNode {
+		let currentDoc;
```
EricJizbaMSFT: return this._children((child) => child.id.toString() === id);

src/DocumentEditor.ts
```diff
@@ -4,154 +4,79 @@
...
     }
 
     public async dispose(): Promise<void> {
-        if (this.recoveredFileName) {
-            await fse.ensureDir(this.recoveredDocsFolder);
-            await fse.move(this.localDocPath, path.join(this.recoveredDocsFolder, this.recoveredFileName));
-        } else {
-            await fse.unlink(this.localDocPath);
-        }
+        Object.keys(this.fileMap).forEach(async (key) => await fse.unlink(key));
```
EricJizbaMSFT: We want to remove the directory, too. Just do:
```
Object.keys(this.fileMap).forEach(async (key) => await fse.remove(path.dirname(key)));
```

src/DocumentEditor.ts
```diff
@@ -4,154 +4,79 @@
...
-                        throw new UserCancelledError();
-                    } else if (result === DialogBoxResponses.DontShowAgain) {
-                        await globalState.update(this.dontShowKey, true);
-                    }
+        const filePath = Object.keys(this.fileMap).find((filePath) => path.relative(doc.uri.fsPath, filePath) === '');
+        if (!this.ignoreSave && filePath) {
+            const node: IEditableNode = this.fileMap[filePath][1];
+            const dontShow: boolean | undefined = globalState.get(this.dontShowKey);
+            if (dontShow !== true) {
+                const message: string = `Saving this file will update the entity "${node.label}" in Azure.`;
```
EricJizbaMSFT: 'cosmos-editor.json' instead of 'this file'
'to the Cloud' instead of 'in Azure'

src/mongo/nodes.ts
PrashanthCorp: getChildren is async. I can't really call it from a property (get data()) or the constructor. 
For a find result, I don't see how we can call node.getChildren at the point of showDocument because we need to await it.

src/mongo/commands.ts
```diff
@@ -30,7 +31,21 @@ export class MongoCommands {
...
-			await util.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
+			const parsed = JSON.parse(result);
+			if (command.name === 'find' || command.name === 'findOne') {
+				const db = await database.getDb();
+				let node: MongoCollectionNode | MongoDocumentNode;
+				if (command.name === 'find') {
+					node = new MongoCollectionNode(db.collection(command.collection), database, parsed, command.arguments);
```
EricJizbaMSFT: Call node.getChildren() here right after you create the node.
This method is async

src/mongo/commands.ts
PrashanthCorp: I'll have to run `executeCommand` for a `findOne` too. Excluding `executeCommand` for a find alone seems clunky, given that the statements around the node creation+getChildren are the same for the `find` and `findOne` case. 

That said, I'm performing this change, have a look in the next iteration.

EricJizbaMSFT:
@PrashanthCorp please make sure you do a squash merge for this. When we go to release we will cherry-pick just this commit onto a 'release' branch and it's much harder to do if it's a bunch of commits

PrashanthCorp:
Linking issues #191 #192 #198

PR #221, opened 2017-11-18
Height resizes responsively

nturinski:
The hardcoded pixel numbers are to account for the static headers

### DIFF
--- a/src/graph/GraphViewsManager.ts
+++ b/src/graph/GraphViewsManager.ts
@@ -92,7 +92,7 @@ class GraphViewDocumentContentProvider implements vscode.TextDocumentContentProv
     <!DOCTYPE html>
     <html>
       <body>
-        <iframe src="file://${clientHtmlPath}?port=${server.port}" width="100%" height="1000px" style="border:0"></iframe>
+        <iframe src="file://${clientHtmlPath}?port=${server.port}" style="width: 100%; height: 100%; position: absolute; padding: 0; margin: 0; border: none"></iframe>
       </body>
     </html>
     `;
--- a/src/graph/client/graphClient.ts
+++ b/src/graph/client/graphClient.ts
@@ -13,10 +13,8 @@
 declare let d3: any;
 
 const animationStepMs = 50;
-const graphWidth = 1200, graphHeight = 700;
 const defaultQuery = "g.V()";
-
-const linkDistance = graphWidth / 3;
+const linkDistance = 400;
 const linkStrength = 0.01; // Reduce rigidity of the links (if < 1, the full linkDistance is relaxed)
 const charge = -3000;
 const markerDistanceFromVertex = 10;
@@ -33,7 +31,8 @@ let htmlElements: {
   queryError: HTMLTextAreaElement,
   queryInput: HTMLInputElement,
   stats: HTMLSpanElement,
-  title: HTMLElement
+  title: HTMLElement,
+  resultsBackground: HTMLDivElement
 };
 
 type State = "empty" | "querying" | "error" | "json-results" | "graph-results";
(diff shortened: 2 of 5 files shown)


src/graph/client/graphClient.ts
```diff
@@ -51,6 +51,18 @@ window.onerror = (message) => {
   logToUI("ERROR: " + message);
 };
 
+window.addEventListener('resize', resizeCanvas, false);
+
+function resizeCanvas() {
+  const resultsBackground = document.getElementById("resultsBackground");
```
StephenWeatherford: extract duplicate code

src/graph/client/graphClient.ts
```diff
@@ -51,6 +51,18 @@ window.onerror = (message) => {
   logToUI("ERROR: " + message);
 };
 
+window.addEventListener('resize', resizeCanvas, false);
+
+function resizeCanvas() {
```
StephenWeatherford: nit: canvas?

src/graph/client/graphClient.ts
```diff
@@ -367,7 +379,13 @@ export class GraphClient {
       if (this._force) {
         this._force.stop();
       }
+      const resultsBackground = document.getElementById("resultsBackground");
```
StephenWeatherford: duplicate code

src/graph/GraphViewsManager.ts
```diff
@@ -92,7 +92,7 @@ class GraphViewDocumentContentProvider implements vscode.TextDocumentContentProv
     <!DOCTYPE html>
     <html>
       <body>
-        <iframe src="file://${clientHtmlPath}?port=${server.port}" width="100%" height="1000px" style="border:0"></iframe>
+        <iframe src="file://${clientHtmlPath}?port=${server.port}" style="width: calc(100% - 40px); height: 100%; position: absolute; padding: 0; margin: 0; border: none"></iframe>
```
StephenWeatherford: I'd rather not having sizing here (the 100%-40px) - that should be in the .css file.

src/graph/client/graphClient.ts
```diff
@@ -13,8 +13,8 @@
 declare let d3: any;
 
 const animationStepMs = 50;
-const graphWidth = 1200, graphHeight = 700;
 const defaultQuery = "g.V()";
+let graphWidth = 1200, graphHeight = 700;
```
StephenWeatherford: Should this be calculated immediately rather than using arbitrary values?  I don't want it to flicker.  Do we even need these variables?

src/graph/GraphViewsManager.ts
nturinski: I tried to implement this in the .css, but I don't think the css properties were being applied to the class before the iframe would render.  It may have to do with the fact that the .css is loaded inthe graphClient.html, which is loaded within the iframe being created here.

src/graph/client/graphClient.ts
nturinski: I don't know if we need this or not.  I kept them as a default in case resultsBackground didn't render by the time I set the width/height of the force window.  I didn't notice any flickering.

I'm fine with removing them though.

src/graph/client/graphClient.ts
nturinski: I called it canvas as it was where the graph results are being drawn.  Did you have a name in mind?

src/graph/GraphViewsManager.ts
StephenWeatherford: This is just for margins, right?  Let the iframe fill out the entire available space, and put the margins into the graphClient.html body or something, which can then be in the graphClient.css.

src/graph/client/graphClient.ts
StephenWeatherford: "Canvas" is a specific HTML element kind, I'm using an SVG instead, which is what is being resized

src/graph/client/graphClient.ts
```diff
@@ -367,7 +379,13 @@ export class GraphClient {
       if (this._force) {
         this._force.stop();
       }
+      const resultsBackground = document.getElementById("resultsBackground");
+      if (resultsBackground) {
```
StephenWeatherford: Here you're not resizing the SVG, you're setting up the initial size of the simulation, which you can go ahead and do based on the current resultsBackground.clientWidth/Height.  No need for file variables.

src/graph/client/graphClient.ts
StephenWeatherford: 1) hard-code linkDistance, so you can get rid of the file-scoped variables
2) Move resutsBackground to htmlElements
3) No need for try - if there is no resultsBackground element, it should simply cause an error - being defensive without a reason can hide issues with the code, so you shouldn't just use it everywhere.

src/graph/client/graphClient.ts
```diff
@@ -51,13 +50,24 @@ window.onerror = (message) => {
...
+    .attr("width", htmlElements.resultsBackground.clientWidth).attr("height", htmlElements.resultsBackground.clientHeight);
+}
 function logToUI(s: string) {
   console.log(s);
   // let v = htmlElements.debugLog.value;
   // v += "\r\n" + s;
   // htmlElements.debugLog.value = v;
 }
 
+function getHtmlElementDimensions(id: string): number[] {
```
StephenWeatherford: I don't see this getting used?

PR #224, opened 2017-11-21
Minor refactoring: GraphView

StephenWeatherford:
Separate graphics a little bit better, still in same file for now.

### DIFF
--- a/src/graph/client/graphClient.ts
+++ b/src/graph/client/graphClient.ts
@@ -51,6 +51,10 @@ window.onerror = (message) => {
   logToUI("ERROR: " + message);
 };
 
+function getErrorMessage(error: any) {
+  return error.message || error.toString();
+}
+
 function logToUI(s: string) {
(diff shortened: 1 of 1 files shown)


src/graph/client/graphClient.ts
```diff
@@ -273,16 +286,176 @@ export class GraphClient {
...
+
+  public display(countUniqueVertices: number, vertices: GraphVertex[], countUniqueEdges: number, edges: GraphEdge[]) {
+    this._countUniqueVertices = countUniqueVertices;
+    this._vertices = vertices;
+    this._countUniqueEdges = countUniqueEdges;
+    this._edges = edges;
+
+    this.clear();
+
+    // Set up nodes and links for the force simulation
```
StephenWeatherford: All I did here was move the code and also remove a try/catch (thus all the detected whitespace changes), which was moved to the generic server message handler.

PR #225, opened 2017-11-21
Add vscodeignore

EricJizbaMSFT:
This is used when creating a vsix. It actually doesn't affect the vsix size that much, but it's still the clean thing to do

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,0 +1,11 @@
+.vscode/**
+.vscode-test/**
+out/test/**
+test/**
+src/**
+**/*.map
+.gitignore
+.travis.yml
+package-lock.json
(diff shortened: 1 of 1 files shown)


.vscodeignore
```diff
@@ -0,0 +1,11 @@
+.vscode/**
+.vscode-test/**
+out/test/**
+test/**
+src/**
+**/*.map
```
PrashanthCorp: What does this do? I don't quite understand this

EricJizbaMSFT: Specifically the line about '.map'? They're generated for every build, but their not necessary in production

https://stackoverflow.com/questions/17493738/what-is-a-typescript-map-file

PR #230, opened 2017-11-27
Leverage shared azure tree provider

EricJizbaMSFT:
This commit should have no functional change (I didn't even add the showNodePicker support yet). The refactoring for edit & persist will also be done separately.

This will require https://github.com/Microsoft/vscode-azuretools/pull/30 in order to work.

I left a _few_ file names as-is for the sake of review. I will rename before merging.


EricJizbaMSFT:
@StephenWeatherford nothing in particular, no. However, I would like you to review this PR as well: https://github.com/Microsoft/vscode-cosmosdb/pull/231

I split it up into 2 for the sake of review, but they should be merged at the same time

PR #242, opened 2017-11-29
Rename openGraph to match convention

StephenWeatherford:


src/extension.ts
```diff
@@ -112,7 +112,7 @@ export function activate(context: vscode.ExtensionContext) {
 		await node.parentNode.addMoreChildren();
 		explorer.refresh(node.parentNode);
 	});
-	initAsyncCommand(context, 'graph.openExplorer', async (graph: GraphNode) => {
+	initAsyncCommand(context, 'cosmosDB.openGraphExplorer', async (graph: GraphNode) => {
```
EricJizbaMSFT: You have to update this reference as well:
https://github.com/Microsoft/vscode-cosmosdb/blob/9f97c309bed04c8f1bf2bf6a7014939c916fae37/src/graph/tree/GraphCollectionTreeItem.ts#L18

StephenWeatherford: Added command, although it doesn't do anything yet if not called with a context.  Is that an issue with all of Cosmos?

EricJizbaMSFT: Yes - that bug is assigned to me (https://github.com/Microsoft/vscode-cosmosdb/issues/35). In order to fix it, the first step was to move to the shared UI package that supports `showNodePicker`. I'll be getting to the second part (actually using `showNodePicker`) sometime soon

PR #248, opened 2017-12-04
Changes to DocumentEditor dispose to improve VSCode restarting experience

PrashanthCorp:
Create a backup of the file before clearing it.

### DIFF
--- a/src/DocumentEditor.ts
+++ b/src/DocumentEditor.ts
@@ -51,7 +51,10 @@ export class DocumentEditor implements vscode.Disposable {
     }
 
     public async dispose(): Promise<void> {
-        Object.keys(this.fileMap).forEach(async (key) => await fse.remove(path.dirname(key)));
+        Object.keys(this.fileMap).forEach((key) => {
+            fse.copySync(key, "backup-" + key);
+            fse.writeFileSync(key, "");
+        });
     }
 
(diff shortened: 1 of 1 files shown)


src/DocumentEditor.ts
```diff
@@ -51,7 +51,12 @@ export class DocumentEditor implements vscode.Disposable {
     }
 
     public async dispose(): Promise<void> {
-        Object.keys(this.fileMap).forEach(async (key) => await fse.remove(path.dirname(key)));
+        Object.keys(this.fileMap).forEach((key) => {
+            const backupFileName = key.substring(0, key.lastIndexOf('.')) + "-backup.json";
+            fse.ensureFileSync(backupFileName)
+            fse.copySync(key, backupFileName);
+            fse.writeFileSync(key, "We do not support editing entities across sessions. Please open an entity from the tree on the left to continue. The previous changes you made here are in the corresponding back-up file.");
```
EricJizbaMSFT: 'across sessions' -> 'across VS Code sessions'

EricJizbaMSFT: Instead of this:
"Please open an entity from the tree on the left to continue. The previous changes you made here are in the corresponding back-up file."

How about this:
```
`Reopen the entity or view your previous changes here: ${backupFileName}`
```

It's shorter, more general (since they might've executed a mongo command instead of using the tree), and has the path to the file

PR #254, opened 2017-12-05
Pre-release bugs pt1. Fix #252 #253

### LINKED ISSUE #252: Mongo scrapbook command with query fails
1. Connect to a mongo db
1. Execute a 'find' from a mongo scrapbook with a query (for example: `db.Test.find({"value": "abc"})`)

Expected: It displays the results of the query
Actual: 
screen shot 2017-12-05 at 12 05 49 pm

We're probably just missing a `JSON.parse` somewhere in the code

### LINKED ISSUE #253: Cosmos editor files should not be stored in extension path
We're worried users will navigate to this folder and accidentally mess up the extension. Instead we could do something like `<OS Temp folder>/vscode-cosmosdb-editors/`

We can also improve the look of the error message. Here's the current view:
screen shot 2017-12-05 at 12 16 23 pm

Per discussion with Jing:
1. Add comments to get rid of the red squiggle
1. Make it multi-line

screen shot 2017-12-05 at 12 16 10 pm

PrashanthCorp:
Fix #253 #252


src/DocumentEditor.ts
```diff
@@ -20,8 +20,8 @@ export class DocumentEditor implements vscode.Disposable {
 
     private readonly dontShowKey: string = 'cosmosDB.dontShow.SaveEqualsUpdateToAzure';
 
-    public async showDocument(docNode: IEditableNode, extensionPath: string, fileName: string): Promise<void> {
-        const localDocPath = path.join(extensionPath, fileName);
+    public async showDocument(docNode: IEditableNode, fileName: string): Promise<void> {
+        const localDocPath = path.join(os.tmpdir(), 'vscode-cosmosdb-editor-json', fileName);
```
EricJizbaMSFT: '-json' seems unnecessary

cosmos-result.json
```diff
@@ -0,0 +1 @@
+null
```
EricJizbaMSFT: Remove this file

src/mongo/nodes.ts
```diff
@@ -257,7 +257,7 @@ export class MongoCollectionNode implements IEditableNode {
 	async getChildren(): Promise<INode[]> {
 		if (!this._hasFetched) {
 			this._children = [];
-			this._iterator = this.collection.find(this.query);
+			this._iterator = this.collection.find(JSON.parse(this.query));
```
EricJizbaMSFT: You did a lot of work to ensure `this.query` is `null` instead of `undefined`, but it would seem less bug-prone if you just did this:
```
this._iterator = this.query ? this.collection.find(JSON.parse(this.query)) : this.collection.find();
```

That way we don't have to worry if you 'missed' one case where it was supposed to be null instead of undefined

src/mongo/commands.ts
```diff
@@ -34,15 +34,19 @@ export class MongoCommands {
...
 				return command;
 			}
 			const result = await database.executeCommand(command);
 			const parsed = JSON.parse(result);
 			if (command.name === 'findOne') {
 				const db = await database.getDb();
-				let node = new MongoDocumentNode(parsed._id, null, parsed);
-				await editor.showDocument(node, extensionPath, 'cosmos-result.json');
+				let id = parsed && parsed._id;
+				if (typeof id === "undefined") {
```
EricJizbaMSFT: Similar to the other case, it seems bug-prone to fix this here rather than where you're using id. (I assume you did this because of `findDocById`? You should just do `child && child.id.toString() === id`)

PrashanthCorp: I did this because `JSON.parse(null)` works, while `JSON.parse(undefined)` doesn't

PR #267, opened 2017-12-08
Pick up VS Code theme, fixes #228

### LINKED ISSUE #228: Graph does not follow VSCode theme

StephenWeatherford:
There will still be some adjustments after meeting with UX, but this handles the major changes.

image

### DIFF
--- a/resources/graphClient/graphClient.css
+++ b/resources/graphClient/graphClient.css
@@ -12,12 +12,26 @@
 html {
     width: 100%;
     height: 100%;
+
+    /*
+    NOTE:
+
+        var(--background-color) and var(--color)
+            - Picked up at runtime and represent the theme background and foreground colors
+        var(--graph-background-color) and var(--graph-color):
+            Specifically just for the graph display, regardless of theme
+    */
+
+    --graph-background-color: black;
+    --graph-color: #ccc;
+    --graph-border:  #454545 1px solid;
+    --button-background-color: #0e639c;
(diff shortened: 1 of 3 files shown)


resources/graphClient/graphClient.css
```diff
@@ -12,12 +12,26 @@
 html {
```
EricJizbaMSFT: The colors in the text box seem backwards. For example this is what it looks like in dark theme:
screen shot 2017-12-08 at 3 50 37 pm

But This is what the colors look like when I'm editing a normal file (Notice the _text_ is the light color and the background is dark):
screen shot 2017-12-08 at 3 55 24 pm

resources/graphClient/graphClient.css
```diff
@@ -12,12 +12,26 @@
...
     width: 100%;
```
EricJizbaMSFT: It's hard to read 'No data is loaded' text in certain themes:
screen shot 2017-12-08 at 3 51 20 pm

resources/graphClient/graphClient.css
StephenWeatherford: It's intentional based on the original design from UX. Agreed it doesn't look as good with this theme.  I'll be discussing stuff like that next week and can adjust.

StephenWeatherford:
I’m aware of it, yes.  Not sure of an easy fix, but the work-around is pretty easy so I’m not worrying about it now.

PR #268, opened 2017-12-09
Add telemetry for gremlin queries

StephenWeatherford:
1) Refactors functions in utils.ts -> telemetry.ts, vscodeUtils.ts
2) Move a couple files to utils 
3) Pull telemetry code out of wrapAsyncCallback into telemetry.ts
4) Add telemetry for graph queries


src/utils/telemetry.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+export async function callWithTelemetry<T>(eventName: string, callback: (telemetryProperties: { [key: string]: string; }, measurements: { [key: string]: number }) => Promise<void>): Promise<void> {
+    const start = Date.now();
+    let properties: { [key: string]: string; } = {};
+    properties.result = 'Succeeded';
+    let measurements: { [key: string]: number; } = {};
+    let errorData: ErrorData | undefined = null;
+    let result: T = undefined;
```
EricJizbaMSFT: I don't think result is ever used? And I think this is the only use of T? Could remove both

src/mongo/commands.ts
```diff
@@ -38,7 +38,7 @@ export class MongoCommands {
 				if (command.name === 'findOne') {
 					await editorManager.showDocument(new MongoFindOneResultEditor(database, command.collection, result));
 				} else {
-					await util.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
+					await vscodeUtil.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
```
PrashanthCorp: I expect a small merge conflict on this line. Just a heads-up.

PrashanthCorp: This block of code was refactored a bit. Heads-up about any merge conflicts.

src/utils/vscodeUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function getOutputChannel(): vscode.OutputChannel {
+    return outputChannel;
+}
+
+export interface IDisposable {
+    dispose(): void;
+}
+
+export function dispose<T extends IDisposable>(disposables: T[]): T[] {
```
PrashanthCorp: I realize this is a file rename, so this line of code may not be down to you. 

This [documentation page](https://vshaxe.github.io/vscode-extern/vscode/Disposable.html) mentions a void return.  Any reason why we're returning an array here?

PR #269, opened 2017-12-09
Merge commit from release-0.2.2 to master

PrashanthCorp:


src/CosmosEditorManager.ts
```diff
@@ -24,11 +24,22 @@ export class CosmosEditorManager implements vscode.Disposable {
 
     private readonly dontShowKey: string = 'cosmosDB.dontShow.SaveEqualsUpdateToAzure';
 
-    public async showDocument(editor: ICosmosEditor): Promise<void> {
-        const localDocPath = path.join(os.tmpdir(), randomUtils.getRandomHexString(12), 'cosmos-editor.json');
+    public async showDocument(editor: ICosmosEditor, fileName: string): Promise<void> {
+        const localDocPath = path.join(os.tmpdir(), fileName);
```
EricJizbaMSFT: This should also have 'vscode-cosmosdb-editor' in the path

src/mongo/commands.ts
```diff
@@ -32,11 +32,11 @@ export class MongoCommands {
 			}
 
 			if (command.name === 'find') {
-				await editorManager.showDocument(new MongoFindResultEditor(database, command));
+				await editorManager.showDocument(new MongoFindResultEditor(database, command), 'cosmos-collection.json');
```
EricJizbaMSFT: Both instances in this file should be 'cosmos-result.json', correct?

PR #271, opened 2017-12-11
Add mongo.collection.count functionality to scrapbook

### LINKED ISSUE #200: Prompting for input - not sure what to do
- I ran db.mycollectionname.count() and was prompted by VS Code to enter my input. Not sure what that is. I got an error from VS Code and no results returned for the scrapbook. I expected to see the count from https://docs.mongodb.com/manual/reference/method/db.collection.count/#examples

PrashanthCorp:
Fixes #200

### DIFF
--- a/src/mongo/tree/MongoCollectionTreeItem.ts
+++ b/src/mongo/tree/MongoCollectionTreeItem.ts
@@ -125,6 +125,9 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 			if (name === 'remove') {
 				return reportProgress(this.remove(args ? parseJSContent(args) : undefined), 'Removing');
 			}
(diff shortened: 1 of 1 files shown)


src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -189,6 +192,11 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 			});
 	}
 
+	private async count(args?: any): Promise<string> {
+		const count = await this.collection.count(args);
+		return String(count);
```
EricJizbaMSFT: Let's do `this.stringify({ count })` instead of `String(count)` so that we're consistent with other commands

PR #276, opened 2017-12-12
Use 'cancel' instead of 'close'

### LINKED ISSUE #222: Use 'cancel' instead of 'close' on messages

EricJizbaMSFT:
Fixes #222

### DIFF
--- a/src/CosmosEditorManager.ts
+++ b/src/CosmosEditorManager.ts
@@ -11,6 +11,7 @@ import { DialogBoxResponses } from './constants';
 import { UserCancelledError } from 'vscode-azureextensionui';
 import * as util from './utils/vscodeUtils';
 import { randomUtils } from './utils/randomUtils';
+import { MessageItem } from 'vscode';
 
 export interface ICosmosEditor<T = {}> {
     label: string;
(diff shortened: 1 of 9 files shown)


src/CosmosEditorManager.ts
```diff
@@ -90,12 +88,12 @@ export class CosmosEditorManager implements vscode.Disposable {
...
             if (dontShow !== true) {
                 const message: string = `Saving 'cosmos-editor.json' will update the entity "${editor.label}" to the Cloud.`;
-                const result: string | undefined = await vscode.window.showWarningMessage(message, DialogBoxResponses.OK, DialogBoxResponses.DontShowAgain);
+                const result: MessageItem | undefined = await vscode.window.showWarningMessage(message, DialogBoxResponses.OK, DialogBoxResponses.DontShowAgain, DialogBoxResponses.Cancel);
 
-                if (!result) {
-                    throw new UserCancelledError();
-                } else if (result === DialogBoxResponses.DontShowAgain) {
+                if (result === DialogBoxResponses.DontShowAgain) {
                     await globalState.update(this.dontShowKey, true);
```
StephenWeatherford: Shouldn't this throw canceled?

EricJizbaMSFT: In this case "Don't show again" means "Upload and don't warn again". We've actually been discussing improving that message and I just filed a bug here: https://github.com/Microsoft/vscode-cosmosdb/issues/280

For now, this PR keeps the behavior the same as it was and I would prefer to fix the message later (since people have strong opinions on that and not this change)

StephenWeatherford: Yep, I had assumed it meant "no and don't ask again".

PR #281, opened 2017-12-12
Handles errors that have message instead of body, Fixes #186

### LINKED ISSUE #186: Creating a graph fails silently if graph name is invalid
Graph name: @##$#$#
Partition Key: asdfasdf

If the partition key is valid, but the graph name is not, the "Create Graph" command will silently fail.

StephenWeatherford:


src/utils/makeError.ts
```diff
@@ -0,0 +1,72 @@
+import { error } from "util";
+import { ErrorCodes } from "vscode-languageserver/lib/main";
```
EricJizbaMSFT: Move these after the copyright notice

src/docdb/tree/DocDBCollectionTreeItemBase.ts
```diff
@@ -58,7 +59,7 @@ export abstract class DocDBCollectionTreeItemBase extends DocDBTreeItemBase<Retr
             const client = this.getDocumentClient();
             await new Promise((resolve, reject) => {
                 client.deleteCollection(this.link, function (err) {
-                    err ? reject(new Error(err.body)) : resolve();
+                    err ? reject(makeError(err)) : resolve();
```
EricJizbaMSFT: You shouldn't need to call `makeError` everywhere. All errors should be caught and handled in one place. 

Can we remove the `.body` part and just just do `reject(err)` and still handle the error as a part of ErrorData here?: https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L168

whiskeyjay:
Why is the package-lock.json in the repo?

EricJizbaMSFT:
@whiskeyjay not sure why it was there, but I removed it in a PR this morning: https://github.com/Microsoft/vscode-cosmosdb/pull/297

EricJizbaMSFT:
Can we make the bulk of these changes here instead?: https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/parseError.ts

I think other extensions would benefit from this work

StephenWeatherford:
@EricJizbaMSFT I'd rather move it to UI after checking it in here.

EricJizbaMSFT:
Any particular reason why? My biggest concern with that is the unit tests. There are several in the UI package that you have not included in your changes here. In other words - I can't confidently say that your changes will not cause a regression.

StephenWeatherford:
Okay.  Before I can do so, we need to figure out what to do with callWithTelemetry in Cosmos, which still has a dependence on ErrorData (in order to get a proper error message).  I had refactored this in Cosmos so that the Command handling code and the gremlin query telemetry go through the same code path, but you've moved the command handling to UI but not callWithTelemetry.  callWithTelemetry is essentially wrapCallback intended for general usage (telemetry sometimes is needed at finer granularity than just at the command/event level).

StephenWeatherford:
@EricJizbaMSFT In other words, I think we should expose callWithTelemetry, unless you have a better suggestion.

EricJizbaMSFT:
I've already exposed ErrorData from the UI package (although it was renamed to parseError because I thought it made more sense as a function). If you want the simplest path, just leverage parseError and leave callWithTelemetry in Cosmos.

However I agree that we should expose callWithTelemetry. I tried it yesterday and it wasn't a simple copy/paste of code (because of my new 'trackTelemetry' parameter) - so I basically just put that work off for later.

PR #283, opened 2017-12-12
Add MongoDB database name validation. Closes #140

### LINKED ISSUE #140: Create Database InputBox doesn't do input validation

peterblazejewicz:
This commit uses MongodB 3.6 docs specs to verify input when new
MongoDB is to be created:

https://docs.mongodb.com/manual/reference/limits/\#naming-restrictions

Thanks!

### DIFF
--- a/src/mongo/tree/MongoAccountTreeItem.ts
+++ b/src/mongo/tree/MongoAccountTreeItem.ts
@@ -59,7 +59,8 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
     public async createChild(_node: IAzureNode, showCreatingNode: (label: string) => void): Promise<IAzureTreeItem> {
         const databaseName = await vscode.window.showInputBox({
             placeHolder: "Database Name",
-            prompt: "Enter the name of the database"
+            prompt: "Enter the name of the database",
+            validateInput: validateDatabaseName,
         });
         if (databaseName) {
             const collectionName = await vscode.window.showInputBox({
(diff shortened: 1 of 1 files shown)


src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -78,6 +79,20 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
 
         throw new UserCancelledError();
     }
+
+}
+
+function validateDatabaseName(database: string): string | undefined | null {
+    // https://docs.mongodb.com/manual/reference/limits/#naming-restrictions
+    const min = 1;
+    const max = 64;
```
PrashanthCorp: >Database names cannot be empty and must have *fewer* than 64 characters.
Sounds like max should be 63?

src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -78,6 +79,20 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
...
+    if(!database || database.length < min || database.length > max) {
+        return `The name must be between ${min} and ${max} characters.`;
+    }
+    if (/[/\\. "$*<>:|?]/.test(database)) {
```
PrashanthCorp: Note : We allow the user to attach to their own mongo servers too. This can reside on another cloud, on-prem, independent of Azure. 

I understand you are implementing the more stringent Windows restrictions. Given that these are placeholder validations, we feel the less restrictive ones (Linux) would be better.  We feel denying a valid resource name could hinder the experience more than allowing a bad use case (an error is thrown).

src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -78,6 +79,20 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
 
         throw new UserCancelledError();
     }
+
+}
+
+function validateDatabaseName(database: string): string | undefined | null {
```
PrashanthCorp: I don't see it used elsewhere. Specifying it as `private validateDatabaseName` would help.

peterblazejewicz: Prashanth, that's the module level function. Should I move it into class and then make private?

src/mongo/tree/MongoAccountTreeItem.ts
peterblazejewicz: Good catch indeed, I misread 👍

src/mongo/tree/MongoAccountTreeItem.ts
PrashanthCorp: Hmm. It seems private to the file since there's no export. This should be fine. :-)

PrashanthCorp:
Merging this. Thanks again for your contribution!

PR #284, opened 2017-12-13
Document Update : throw error if doc ID is not found.

### LINKED ISSUE #126: Update mongo document doesn't throw error if id is not found
Repro steps:
1. Open a mongo document
1. Edit _id to an id that doesn't exist
1. Update to Azure

Expected: Error saying we could not find that id
Actual: No error

PrashanthCorp:
Fixes #126

### DIFF
--- a/src/mongo/tree/MongoDocumentTreeItem.ts
+++ b/src/mongo/tree/MongoDocumentTreeItem.ts
@@ -66,7 +66,7 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
         const filter: object = { _id: new ObjectID(newDocument._id) };
         const result = await collection.updateOne(filter, _.omit(newDocument, '_id'));
         if (result.upsertedCount != 1) {
-            throw new Error(`Failed to update document with _id ${newDocument._id}.`);
+            throw new Error(`Failed to update document with _id '${newDocument._id}'.`);
         }
         return newDocument;
     }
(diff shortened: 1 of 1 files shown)


src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -64,7 +64,10 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
 
     public static async update(collection: Collection, newDocument: IMongoDocument): Promise<IMongoDocument> {
         const filter: object = { _id: new ObjectID(newDocument._id) };
-        await collection.updateOne(filter, _.omit(newDocument, '_id'));
+        const result = await collection.updateOne(filter, _.omit(newDocument, '_id'));
+        if (result.upsertedCount != 1) {
+            throw new Error("Document not found. Please check if ID exists.");
```
EricJizbaMSFT: We don't know if "Document not found" is the reason it failed. It could have been any number of reasons.

What if we generalize it a little bit: ``Failed to update document with _id "${newDocument._id}".``

src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -66,7 +66,7 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
         const filter: object = { _id: new ObjectID(newDocument._id) };
         const result = await collection.updateOne(filter, _.omit(newDocument, '_id'));
         if (result.upsertedCount != 1) {
-            throw new Error("Document not found. Please check if ID exists.");
+            throw new Error(`Failed to update document with _id ${newDocument._id}.`);
```
StephenWeatherford: Put quotes around the ID.

PR #285, opened 2017-12-14
Add view settings for graph customization, fixes #234 

### LINKED ISSUE #234: Allow customizing the text to show for a node

StephenWeatherford:


README.md
```diff
@@ -36,6 +36,33 @@
 
 ![Query Graphs](resources/Graph.gif)
 
+- Configure the user setting `mongocosmosDB.graph.viewSettings` to customize which properties to display and which colors to use based on vertex label.
```
EricJizbaMSFT: Remove 'mongo'

src/graph/client/graphTypes.ts
```diff
@@ -27,6 +27,39 @@ interface GraphEdge {
...
+  displayProperty?: string[];
+  color?: string;
+}
+
+interface GraphSettingsGroup {
+  vertices?: VertexSettingsGroup[];
+}
+
+interface ViewSettings {
+  // Currently on default group is supported
```
EricJizbaMSFT: "only" instead of "on"?

package.json
```diff
@@ -555,6 +555,77 @@
 					"type": "integer",
 					"default": 500,
 					"description": "Set a limit for number of edges displayed in a graph visualization"
+				},
+				"cosmosDB.graph.viewSettings": {
+					"type": "object",
+					"description": "Settings for graph visualization. Currently only the 'default' group is supported.",
```
EricJizbaMSFT: I don't understand the purpose of the 'default' group. Can you explain what other 'groups' you plan to support in the future? Just an example or two?

README.md
```diff
@@ -36,6 +36,33 @@
```
EricJizbaMSFT: Something's not quite right. This is what I see with the default settings:
screen shot 2017-12-14 at 11 49 17 am
Zoomed in:
screen shot 2017-12-14 at 11 47 42 am
(It's not centered by default and the labels are overlapping)

This is what I see from master:
screen shot 2017-12-14 at 12 04 39 pm

README.md
```diff
@@ -36,6 +36,33 @@
 
 ![Query Graphs](resources/Graph.gif)
 
+- Configure the user setting `mongocosmosDB.graph.viewSettings` to customize which properties to display and which colors to use based on vertex label.
+```javascript
+    "cosmosDB.graph.viewSettings": {
+        "default": {
+            "vertices": [
+                {
```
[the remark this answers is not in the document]
StephenWeatherford: Label (poor name) is the "type" of node e.g. "person" or "company".  So it won't generally help distinguish between nodes.  Portal shows the ID by default, I think it makes sense for us as well (since label and ID are always there, only one can be in the default).

package.json
StephenWeatherford: What I want in the future is a way to have different settings for different graphs or sets of graphs.  My current thinking is that I would probably end up having named settings groups.  These could either be controlled by the user or auto-generated and not seen by the user in the UI. Then each group could specify which graphs it refers to.  This seems like the most flexible feature for the future.  I'd rather not be designing for the future with it now, but can't think of a way to do that that wouldn't require changes in the settings model later.  I'm open to suggestions, though.

The other main option I considered was like I'm doing with the vertex settings groups - having an array of objects, and the object contains some properties to filter them.  But that seems even more confusing.  Thoughts?

README.md
StephenWeatherford: fixed

README.md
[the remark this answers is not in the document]
StephenWeatherford: Although an interesting option would be to allow showing the label in parentheses.  I wanted to allow expressions like "${name} (${label})", but there are semantics issues to sort out, and I figured I could add that later with a different property like "displayText" so I postponed it.

README.md
StephenWeatherford: Good catch, didn't repro with the test data I was using.

EricJizbaMSFT: Looks better now - thanks

package.json
EricJizbaMSFT: Yes - I think it makes more sense to have an array and mirror the structure of the vertex settings groups. What about this structure:
```json
"cosmosDB.graphViewSettings": [
    {
        "appliesTo": [
            {
                "accountName": "",
                "databaseName": "",
                "graphName": ""
            }
        ],
        "vertexViewSettings": [
            {
                "appliesTo": [
                    {
                        "vertexLabel": ""
                    }
                ],
                "displayProperty": [],
                "color": ""
            }
        ]
    }
]
```

You'll notice the name "graphViewSettings" mirrors "vertexViewSettings" and they both have the same "appliesTo" array (I also considered naming this "filter" instead of "appliesTo")

EricJizbaMSFT: And for now, you could just not support "appliesTo" at the graph level (It would be the same as only having "default" as an option)

src/graph/client/graphTypes.ts
StephenWeatherford: In next PR

package.json
StephenWeatherford: Okay, I like that.

PR #292, opened 2017-12-15
Improve "Don't show again"

### LINKED ISSUE #245: Make "Don't show again" a user setting
We prompt users that saving their document will update it to Azure. Today, we give users the "Don't show again" option, but there's no way to 'reset' that if the user did it on accident.

Instead of storing the value in global state, we could save it as a user setting (which is easy for the user to change).

### LINKED ISSUE #280: Clarify "Don't show again" message
It's not clear if "Don't show again" is 'Yes and don't show again' or 'No and don't show again'

Nathan updated the messages [here](https://github.com/Microsoft/vscode-azuretools/pull/43) - we could follow the same pattern or just wait and leverage the shared editor directly

EricJizbaMSFT:
1. Change it to a user setting
1. Clarify that it still uploads

Fixes #245 
Fixes #280 

This is consistent with the appservice experience implemented by Nathan here: https://github.com/Microsoft/vscode-azuretools/pull/43

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -11,6 +11,8 @@ export namespace DialogBoxResponses {
     export const Yes: MessageItem = { title: "Yes" };
     export const OK: MessageItem = { title: "OK" };
     export const DontShowAgain: MessageItem = { title: "Don't Show Again" };
+    export const upload: MessageItem = { title: "Upload" };
+    export const uploadDontWarn: MessageItem = { title: "Upload, don't warn again" };
     export const No: MessageItem = { title: "No" };
     export const Cancel: MessageItem = { title: "Cancel", isCloseAffordance: true };
 }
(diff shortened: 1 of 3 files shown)


src/constants.ts
```diff
@@ -11,6 +11,8 @@ export namespace DialogBoxResponses {
     export const Yes: MessageItem = { title: "Yes" };
     export const OK: MessageItem = { title: "OK" };
     export const DontShowAgain: MessageItem = { title: "Don't Show Again" };
+    export const upload: MessageItem = { title: "Upload" };
+    export const uploadDontWarn: MessageItem = { title: "Upload, don't warn again" };
```
StephenWeatherford: How about "Always upload"?

EricJizbaMSFT: We discussed that and I think Matt preferred this message for some reason. In any case, this is consistent with our other extensions I'm not going to change the other extensions before the release

PR #293, opened 2017-12-15
Name validation - DocDB DB+Collections, Mongo collections

### LINKED ISSUE #139: "Create New Collection" doesn't validate input?

PrashanthCorp:
Fixes #139

### DIFF
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -122,7 +122,7 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         if (/^[#?\\]*$/.test(key)) {
             return "Cannot contain these characters - ?,#,\\, etc."
         }
-        return null;
+        return undefined;
     }
 
     private static validateThroughput(input: string): string | undefined | null {
@@ -134,6 +134,16 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         } catch (err) {
             return "Input must be a number"
         }
-        return null;
+        return undefined;
+    }
+
+    private static validateCollectionName(name: string): string | undefined | null {
+        if (name.endsWith(" ")) {
+            return "Name may not end with space";
+        }
+        if (/[/\\?#]/.test(name)) {
+            return `Name cannot contain the characters '\\', '/', '#', '?'`;
+        }
+        return undefined;
     }
 }
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -138,6 +138,9 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
     }
 
     private static validateCollectionName(name: string): string | undefined | null {
+        if (!name) {
+            return "Name may not be empty";
+        }
         if (name.endsWith(" ")) {
             return "Name may not end with space";
         }
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -10,6 +10,7 @@ import { DocDBTreeItemBase } from './DocDBTreeItemBase';
 import * as vscode from 'vscode';
 import { DocumentBase } from 'documentdb/lib';
 import { DialogBoxResponses } from '../../constants';
+import { DocDBDatabaseTreeItem } from './DocDBDatabaseTreeItem';
 
 /**
  * This class provides common logic for DocumentDB, Graph, and Table databases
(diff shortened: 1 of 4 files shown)


src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -146,6 +147,21 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
 	}
 }
 
+export function validateCollectionName(collectionName: string): string | undefined | null {
+	// https://docs.mongodb.com/manual/reference/limits/#Restriction-on-Collection-Names
+	if (collectionName.length === 0) {
```
StephenWeatherford: if (!collectionName) ...

I've seen cases where we can get called with an undefined value and that would throw.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -134,6 +134,16 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         } catch (err) {
             return "Input must be a number"
         }
-        return null;
+        return undefined;
+    }
+
+    private static validateCollectionName(name: string): string | undefined | null {
+        if (name.endsWith(" ")) {
```
StephenWeatherford: if (!name) ...

I've seen cases where we can get called with an undefined value and that would throw.

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -146,6 +147,21 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
+		return "Collection name cannot be empty";
```
StephenWeatherford: Does this need to be case insensitive?

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -146,6 +147,21 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
+	}
+	const systemPrefix = "system."
+	if (collectionName.startsWith(systemPrefix)) {
+		return `The "${systemPrefix}" prefix is reserved for internal use. Please refrain from using it`;
```
StephenWeatherford: Does this need to be case insensitive?

src/mongo/tree/MongoDatabaseTreeItem.ts
PrashanthCorp: I don't quite get you here.

PrashanthCorp: If you mean the system. prefix, the link mentioned only talks about the lower case s.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -134,6 +134,19 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         } catch (err) {
             return "Input must be a number"
         }
-        return null;
+        return undefined;
+    }
+
+    private static validateCollectionName(name: string): string | undefined | null {
```
EricJizbaMSFT: I don't see this being used

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -146,6 +147,21 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
 	}
 }
 
+export function validateCollectionName(collectionName: string): string | undefined | null {
```
EricJizbaMSFT: Rename this to `validateMongoCollectionName` since it's exported. We don't want to accidently use this for DocDb or something weird like that

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -146,6 +147,21 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
+	// https://docs.mongodb.com/manual/reference/limits/#Restriction-on-Collection-Names
+	if (!collectionName) {
+		return "Collection name cannot be empty";
+	}
+	const systemPrefix = "system."
+	if (collectionName.startsWith(systemPrefix)) {
+		return `The "${systemPrefix}" prefix is reserved for internal use. Please refrain from using it`;
```
EricJizbaMSFT: nit: I would shorten this to `"${systemPrefix}" is reserved for internal use.`

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -134,6 +134,19 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
...
+        if (!name) {
+            return "Name may not be empty";
```
EricJizbaMSFT: nit: You're not consistent with your error messages accross docdb and mongo. I would prefer "Collection name cannot" over "Name may not" or "Name cannot"

src/mongo/tree/MongoDatabaseTreeItem.ts
StephenWeatherford: Yes that's what I meant

PR #296, opened 2017-12-15
Default throughput to cheapest option

EricJizbaMSFT:
The portal lists 1,000 as the lower limit (not 2,500)
screen shot 2017-12-15 at 11 16 00 am

This fixes it for graph and docdb (the graph tree item inherits from the docdbbase class). It does not fix it for mongo since that goes directly though the mongo api (aka not the cosmos api)

### DIFF
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -132,7 +132,7 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         try {
             const value = Number(input);
             if (value < minThroughput || value > maxThroughput) {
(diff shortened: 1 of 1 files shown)


src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -128,8 +131,8 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
     private static validateThroughput(input: string): string | undefined | null {
         try {
             const value = Number(input);
-            if (value < 2500 || value > 100000) {
-                return "Value needs to lie between 2500 and 100,000"
+            if (value < minThroughput || value > maxThroughput) {
+                return `Value needs to lie between ${minThroughput} and ${maxThroughput}`
```
StephenWeatherford: Since you're changing this, how about simply "Value must be between...", which seems much more natural.

PR #313, opened 2018-01-03
Updating empty documents

### LINKED ISSUE #256: Updating empty documents fails
Tries to get an _id, and fails. 
empty document update

PrashanthCorp:
Fixes #256

### DIFF
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -70,18 +70,22 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     public async update(newData: RetrievedDocument): Promise<RetrievedDocument> {
         const client: DocumentClient = this._collection.getDocumentClient();
         const _self: string = this.document._self;
-        this._document = await new Promise<RetrievedDocument>((resolve, reject) => {
-            client.replaceDocument(_self, newData,
-                { accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this.partitionKeyValue },
-                (err, updated: RetrievedDocument) => {
-                    if (err) {
-                        reject(new Error(err.body));
-                    } else {
-                        resolve(updated);
-                    }
-                });
-        });
-
+        if (Object.getOwnPropertyNames(newData).length === 0) {
+            throw new Error("Updating a document with no fields in the document does nothing");
+        }
+        else {
+            this._document = await new Promise<RetrievedDocument>((resolve, reject) => {
+                client.replaceDocument(_self, newData,
+                    { accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this.partitionKeyValue },
+                    (err, updated: RetrievedDocument) => {
+                        if (err) {
+                            reject(new Error(err.body));
+                        } else {
+                            resolve(updated);
+                        }
+                    });
+            });
+        }
         return this.document;
     }
 
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -85,8 +85,8 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
                         }
                     });
             });
+            return this.document;
         }
-        return this.document;
     }
 
     private getPartitionKeyValue(): string | undefined {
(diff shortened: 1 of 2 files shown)


src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -70,19 +70,23 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
...
-                        reject(new Error(err.body));
-                    } else {
-                        resolve(updated);
-                    }
-                });
-        });
-
-        return this.document;
+        if (Object.getOwnPropertyNames(newData).length === 0) {
+            throw new Error("Updating a document with no fields in the document does nothing");
```
EricJizbaMSFT: I'm not sure this error is very helpful and the user would run into the same problem if they just deleted the _self/_etag fields and left all other fields. Maybe just check if _self and _etag are defined and change the message to something like "The '_self' and '_etag' fields are required to update a document."

src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -58,8 +58,13 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
     }
 
     public async update(newDocument: IMongoDocument): Promise<IMongoDocument> {
-        this.document = await MongoDocumentTreeItem.update(this._collection, newDocument);
-        return this.document;
+        if (Object.getOwnPropertyNames(newDocument).length === 0) {
+            throw new Error("Updating a document with no fields in the document does nothing");
```
EricJizbaMSFT: Same comment as above (Except this time check for _id instead of _self/_etag).

Also move the check down into the static update method instead of here (otherwise anything that calls the static method directly won't get this message).

src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -58,16 +58,14 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
...
-        else {
-            this.document = await MongoDocumentTreeItem.update(this._collection, newDocument);
-            return this.document;
-        }
+        this.document = await MongoDocumentTreeItem.update(this._collection, newDocument);
+        return this.document;
     }
 
     public static async update(collection: Collection, newDocument: IMongoDocument): Promise<IMongoDocument> {
+        if (Object.getOwnPropertyNames(newDocument).indexOf("_id") < 0) {
```
StephenWeatherford: suggest: 
  if ("_id" in newDocument)

or

const id = newDocument._id;
if (!id) {
   throw ...
}
const filter: object = { _id: id };

StephenWeatherford: Note that the last one will throw if "_id" exists but is empty.  The other checks you've got will pass if _id is an empty string (or if it's a truthy non-string).

src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -63,6 +63,9 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
     }
 
     public static async update(collection: Collection, newDocument: IMongoDocument): Promise<IMongoDocument> {
+        if (!newDocument["_id"]) {
```
EricJizbaMSFT: Does this work if newDocument is null? Or do you need to do:
```
if (!newDocument || !newDocument["_id"])
```

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -70,19 +70,24 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
...
-                        reject(new Error(err.body));
-                    } else {
-                        resolve(updated);
-                    }
-                });
-        });
-
-        return this.document;
+        const newDocProperties = Object.getOwnPropertyNames(newData);
+        if (["_self", "_etag", "_rid"].some((element) => !newDocProperties[element])) {
```
EricJizbaMSFT: This should be `!newData[element]` instead of `!newDocProperties[element]`

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -70,19 +70,24 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
...
+            throw new Error(`The "_self", "_etag" and "_rid" fields are required to update a document`);
```
EricJizbaMSFT: I don't think _rid is required? You only use _etag and _self below...

src/mongo/tree/MongoDocumentTreeItem.ts
PrashanthCorp: I tried the following : 
1. Empty document ({}) - In this case, `newDocument` exists, but `newDocument["_id"]` does not. 
2. Empty text  - it doesn't hit this line, since the JSON parser throwsan `Unexpected end of input` error

PR #315, opened 2018-01-04
Leverage shared AzureActionHandler

EricJizbaMSFT:
See this PR for more details: https://github.com/Microsoft/vscode-azuretools/pull/53
Builds will fail until that is merged.


src/extension.ts
```diff
@@ -53,51 +54,52 @@ export function activate(context: vscode.ExtensionContext) {
 
 	context.subscriptions.push(vscodeUtil.getOutputChannel());
 
+	const actionHandler: AzureActionHandler = new AzureActionHandler(context, getOutputChannel(), reporter);
 	// Commands
-	initAsyncCommand(context, 'cosmosDB.createAccount', async (node?: IAzureParentNode) => {
+	actionHandler.registerCommand('cosmosDB.createAccount', async (node?: IAzureParentNode) => {
```
PrashanthCorp: Is there a specific reason why we're moving the initAsyncCommand to the shared lib under the actionHandler?

EricJizbaMSFT: As a general practice, it's better to share code than copy code. In this case, I was spurred to do this work by the bugs mentioned in [this PR](https://github.com/Microsoft/vscode-azuretools/pull/53). I wanted to fix them in one place instead of duplicating the work in each extension.

PR #318, opened 2018-01-08
Leverage callWithTelemetry from shared package

EricJizbaMSFT:
Builds will fail until this is merged: https://github.com/Microsoft/vscode-azuretools/pull/55


package.json
```diff
@@ -668,7 +668,7 @@
 		"socket.io": "^1.7.3",
 		"socket.io-client": "^1.7.3",
 		"underscore": "^1.8.3",
-		"vscode-azureextensionui": "~0.5.0",
+		"vscode-azureextensionui": "~0.5.2",
```
PrashanthCorp: This might need to be 0.5.3 as per the latest PR

src/utils/telemetry.ts
```diff
@@ -47,35 +46,3 @@ function sendTelemetry(eventName: string, properties?: { [key: string]: string;
         reporter.sendTelemetryEvent(eventName, properties, measures);
```
PrashanthCorp: I don't see the function `sendTelemetry` being used, now that the `callWithTelemetry` function is removed.

PR #323, opened 2018-01-12
Fix running commands from command palette

### LINKED ISSUE #35: Several commands fail when run from the Command Pallete

EricJizbaMSFT:
I also cleaned up extension.ts a little bit since the diff was going to look terrible anyways:
1. Split mongo, docdb, and graph commands into separate files
1. Removed azure-account.api.d.ts since it's no longer used

This relies on these changes in the UI package: https://github.com/Microsoft/vscode-azuretools/pull/57 and builds will fail until that PR is merged/published

Fixes #35


EricJizbaMSFT:
ping @StephenWeatherford @PrashanthCorp

Issue #215, opened 2017-11-16
[From marketplace feedback] Support the Cosmos DB Emulator


fiveisprime:
>Excellent extension. Support for the Azure CosmosDB Emulator would be a useful addition though.

tjrobinson:
Thanks for taking this feedback on board. I would like to be able to use the same tool for both development and production and to be able to connect to the Emulator when developing code locally, all from the same tool.

EricJizbaMSFT:
This might be as easy as attaching with a 'localhost' address. Let's look into for 0.4.0

jasonchester:
Connecting to a graph on the emulator with the default connection string:
>AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw s/Jw==;

 I get the following error in the electron developer console:

`messageService.ts:126 self signed certificate: Error: self signed certificate
	at TLSSocket.<anonymous> (_tls_wrap.js:1088:38)
	at emitNone (events.js:86:13)
	at TLSSocket.emit (events.js:188:7)
	at TLSSocket._finishInit (_tls_wrap.js:610:8)
	at TLSWrap.TLSSocket._init.ssl.onhandshakedone (_tls_wrap.js:440:38)
e.doShow @ messageService.ts:126
e.show @ messageService.ts:105
(anonymous) @ mainThreadTreeViews.ts:83
v @ winjs.base.raw.js:1209
enter @ winjs.base.raw.js:901
_run @ winjs.base.raw.js:1068
_error @ winjs.base.raw.js:1041
e.resolveErr @ lazyPromise.ts:78
e._receiveOneMessage @ rpcProtocol.ts:68
(anonymous) @ rpcProtocol.ts:32
(anonymous) @ rpcProtocol.ts:157
e.fire @ event.ts:142
(anonymous) @ ipc.net.ts:81
emitOne @ events.js:96
emit @ events.js:191
readableAddChunk @ _stream_readable.js:178
Readable.push @ _stream_readable.js:136
onread @ net.js:560
log.ts:165   ERR Please return an array of children.: Error: Please return an array of children.
    at file:///C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:299:343
[...]

PR #324, opened 2018-01-12
Add Emulator support - Mongo and DocumentDB entry points

PrashanthCorp:
Addresses #215. 

- Add an 'attach emulator' command in the context menu. 
- Add SSL verification flags to ensure VSCode allows connections despite self-signed certificates in emulators. 
- Modify treeItem creation and persistence to incorporate labels. This helps the `(emulator - Mongo)` tag to persist across sessions. 

ToDo: 
Graphs - remove endpoint checks to allow connections to localhost


src/docdb/tree/DocDBTreeItemBase.ts
```diff
@@ -19,6 +21,8 @@ export abstract class DocDBTreeItemBase<T> implements IAzureParentTreeItem {
     public readonly documentEndpoint: string;
     public readonly masterKey: string;
 
+    public SSLVerify: boolean;
+
```
PrashanthCorp: This is the only public member that's not read only. Not a stray line.

src/docdb/tree/DocDBTreeItemBase.ts
```diff
@@ -33,7 +37,12 @@ export abstract class DocDBTreeItemBase<T> implements IAzureParentTreeItem {
     }
 
     public getDocumentClient(): DocumentClient {
-        return new DocumentClient(this.documentEndpoint, { masterKey: this.masterKey });
+        const documentBase = lib.DocumentBase;
+        var connectionPolicy = new documentBase.ConnectionPolicy();
+        const userPreferenceSSLVerify: boolean = vscode.workspace.getConfiguration().get("http.proxyStrictSSL");
+        connectionPolicy.DisableSSLVerification = !userPreferenceSSLVerify && !this.SSLVerify;
```
PrashanthCorp: This line requires the user setting `http.proxyStrictSSL` to be set to false to pass.

src/extension.ts
```diff
@@ -65,10 +65,18 @@ export function activate(context: vscode.ExtensionContext) {
 	});
 	actionHandler.registerCommand('cosmosDB.attachDatabaseAccount', async () => {
 		const rootNodes = await explorer.getChildren();
-		const attachedNode = <IAzureParentNode<AttachedAccountsTreeItem>>rootNodes.find((node) => node.treeItem instanceof AttachedAccountsTreeItem);
-		if (attachedNode) {
-			await attachedNode.treeItem.attachNewAccount();
-			explorer.refresh(attachedNode);
+		const attachedDatabasesNode = <IAzureParentNode<AttachedAccountsTreeItem>>rootNodes.find((node) => node.treeItem instanceof AttachedAccountsTreeItem);
```
PrashanthCorp: Variable name change because I found it a bit unclear whether attachedNode meant the _attached Database_ or the _Attached Database Accounts_ root

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -81,21 +83,73 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
+                    return;
+                }
+                else {
+                    return "Port must be finite integer";
+                }
+            }
+            catch {
+                return "Port must be finite integer";
+            }
+        };
```
PrashanthCorp: Defining this validation check makes this function look larger than usual. Didn't find a good enough reason to move it outside the `attachEmulator` function though.

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -81,21 +83,73 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
+            }
+            else {
+                defaultPort = "8081";
+            }
+            const port: string = await vscode.window.showInputBox({
+                value: defaultPort,
+                prompt: 'Enter the localhost port for the emulator',
+                validateInput: validateInput,
+                ignoreFocusOut: true
+            });
```
PrashanthCorp: Allow entering a port for this: 
https://stackoverflow.com/questions/44851311/cosmosdb-emulator-cant-start-since-port-is-already-in-use

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -136,45 +190,50 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
                 const accounts: any[] = JSON.parse(value);
                 await Promise.all(accounts.map(async account => {
                     let id: string;
+                    let label: string;
```
PrashanthCorp: ID would be `localhost:8081`, label would be `localhost:8081 (Emulator - MongoDB)`.

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -81,21 +83,73 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
+                ignoreFocusOut: true
+            });
+            if (port) {
+                if (defaultExperience === Experience.MongoDB) {
+                    connectionString = `mongodb://localhost:C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==@localhost:${port}/admin?ssl=true`;
+                }
+                else {
+                    connectionString = `AccountEndpoint=https://localhost:${port}/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==;`;
+                }
+                const label = `localhost:${port} (Emulator - ${defaultExperience})`
```
EricJizbaMSFT: I'm not sure we need 'localhost:8081' if we're going to display 'Emulator' in the label. Why not just make the label "MongoDB Emulator" or "DocDB Emulator"?

src/docdb/tree/DocDBTreeItemBase.ts
EricJizbaMSFT: I'd like to get some second opinions on this since it has potential security risks to the user. @whiskeyjay @StephenWeatherford the Cosmos DB emulator always uses self-signed certificates so we have to set connectionPolicy.DisableSSLVerification for it to work. How do you think we should determine when to set DisableSSLVerification?

The 'http.proxyStrictSSL' Prashanth is using here is defined by VS Code

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -81,21 +83,73 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
+                let treeItem: IAzureTreeItem = await this.createTreeItem(connectionString, defaultExperience, label);
+                if (treeItem instanceof DocDBAccountTreeItem || treeItem instanceof GraphAccountTreeItem || treeItem instanceof TableAccountTreeItem) {
+                    treeItem.SSLVerify = false;
```
EricJizbaMSFT: It feels like this boolean should be 'treeItem.isEmulator = true' since the emulator accounts might have other behavior specific to them.

src/tree/AttachedAccountsTreeItem.ts
PrashanthCorp: I agree more with the vscode setting. The default attach will have the same issue of having to identify when it's an emulator to disable SSL-checks.

src/docdb/tree/DocDBTreeItemBase.ts
StephenWeatherford: Swamped, I'll get back to you.

whiskeyjay: If connectionPolicy.DisableSSLVerification is for the emulator connection only, and setting this flag alone can make things work, I'd show a prompt the first time when a user tries to add the connection explaining the connection will ignore the SSL cert, and ask for their permission to proceed. I believe name "localhost" can be hijacked in some circumstances, causing the extension trying to connect to an unexpected server, that plus ignoring SSL cert can be a problem.

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -81,21 +83,73 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
             throw new UserCancelledError();
         }
     }
 
+    public async attachEmulator(): Promise<void> {
+        let connectionString: string;
+        const defaultExperience = <Experience>await vscode.window.showQuickPick(['MongoDB', 'DocumentDB'], { placeHolder: "Select a Database Account API...", ignoreFocusOut: true });
+        let validateInput: (input: string) => string | undefined | null = (input: string) => {
+            try {
+                if (parseInt(input).toString() === input) {
```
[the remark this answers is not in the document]
PrashanthCorp: Systems have a range reserved for [ephemeral ports](https://en.wikipedia.org/wiki/Ephemeral_port#Range), which vary based on the kernel. Hence, I didn't enforce the number restriction, only that the port is a whole number. Based on [Eric's comment](https://github.com/Microsoft/vscode-cosmosdb/pull/324#discussion_r161867499), I shall be re-writing this portion of the code.

src/extension.ts
```diff
@@ -56,9 +56,20 @@ export function activate(context: vscode.ExtensionContext) {
 		await node.createChild();
 	});
 	actionHandler.registerCommand('cosmosDB.attachDatabaseAccount', async () => {
-		const attachedNode: IAzureParentNode<AttachedAccountsTreeItem> = await getAttachedNode(tree);
-		await attachedNode.treeItem.attachNewAccount();
-		await tree.refresh(attachedNode);
+		const rootNodes = await tree.getChildren();
+		const attachedAccountsNode = <IAzureParentNode<AttachedAccountsTreeItem>>rootNodes.find((node) => node.treeItem instanceof AttachedAccountsTreeItem);
```
EricJizbaMSFT: Use `getAttachedNode` (Although you can change the name to `getAttachedAccountsNode` if you want)

In addition, the node should always exist, so you don't need to check `if (attachedAccountsNode)`

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -28,7 +28,12 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
 		this._accountConnectionString = accountConnectionString;
 		this._parentId = parentId;
 		const uri: vscode.Uri = vscode.Uri.parse(accountConnectionString);
-		this.connectionString = `${uri.scheme}://${uri.authority}/${this._databaseName}?${uri.query}`;
+		const path = uri.path.substring(0, uri.path.lastIndexOf('/'));
+		if (accountConnectionString.includes('localhost')) {
+			this.connectionString = `${uri.scheme}://${uri.authority}${path}/${this._databaseName}?${uri.query}`;
+		} else {
+			this.connectionString = `${uri.scheme}://${uri.authority}/${this._databaseName}?${uri.query}`;
```
EricJizbaMSFT: Path only matters for localhost? It seems like we should include it in both

Something like this should work in all cases:
```
this.connectionString = `${uri.scheme}://${uri.authority}${uri.path}${this._databaseName}?${uri.query}`;
```

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -100,21 +102,53 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
         }
     }
 
+    public async attachEmulator(): Promise<void> {
+        let connectionString: string;
+        const defaultExperience = <Experience>await vscode.window.showQuickPick(['MongoDB', 'DocumentDB', 'Graph'], { placeHolder: "Select a Database Account API...", ignoreFocusOut: true });
+        if (defaultExperience) {
+            let port: number;
+            if (defaultExperience === Experience.MongoDB) {
+                port = vscode.workspace.getConfiguration().get<number>("cosmoDB.emulator.mongoPort") || 10255;
```
EricJizbaMSFT: I thought `WorkspaceConfiguration.get` already returned the default it if the value didn't exist. No need to have the port hard-coded in two places if we can avoid it

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -100,21 +102,53 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
+            }
+            else {
+                port = vscode.workspace.getConfiguration().get<number>("cosmosDB.emulator.port") || 8081;
+            }
+            if (port) {
+                if (defaultExperience === Experience.MongoDB) {
+                    connectionString = `mongodb://localhost:C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==@localhost:${port}/admin?ssl=true`;
```
EricJizbaMSFT: This is the connection string for the 'admin' database. I would suggest removing 'admin' so that it's just the connection string for the mongo account. I believe that would help simplify your connection string logic in MongoDatabaseTreeItem

src/tree/AttachedAccountsTreeItem.ts
PrashanthCorp: That was my understanding too, but my testing seemed to throw an error here. Specifically, it looked at the default for DocDB but not Mongo. 
A restart of VSCode seemed to use the defaults correctly. :)

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -28,7 +28,7 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
 		this._accountConnectionString = accountConnectionString;
 		this._parentId = parentId;
 		const uri: vscode.Uri = vscode.Uri.parse(accountConnectionString);
-		this.connectionString = `${uri.scheme}://${uri.authority}/${this._databaseName}?${uri.query}`;
+		this.connectionString = `${uri.scheme}://${uri.authority}${uri.path}/${this._databaseName}?${uri.query}`;
```
EricJizbaMSFT: Make sure to test non-emulator mongo accounts. This doesn't work in that case:
screen shot 2018-01-22 at 3 58 53 pm

There's an extra slash. You can either remove it after `${uri.path}` and assume the path has it or (even better) do some extra error-checking like this:
```
const path: string = uri.path.endsWith('/') ? uri.path.slice(0, -1) : uri.path;
```

PrashanthCorp: I could not repro this error. 
The `uri.path` takes a `"/"` in that case.  I'll try to repro this and then put in the error-checks.

src/extension.ts
```diff
@@ -56,9 +56,16 @@ export function activate(context: vscode.ExtensionContext) {
 		await node.createChild();
 	});
 	actionHandler.registerCommand('cosmosDB.attachDatabaseAccount', async () => {
-		const attachedNode: IAzureParentNode<AttachedAccountsTreeItem> = await getAttachedNode(tree);
-		await attachedNode.treeItem.attachNewAccount();
-		await tree.refresh(attachedNode);
+		const rootNodes = await tree.getChildren();
```
EricJizbaMSFT: rootNodes is unused. Remove it from here and below

PR #343, opened 2018-01-25
Make mongo connections simpler and more robust

EricJizbaMSFT:
Rather than trying to parse the connection string, just use `.db(databaseName)` after calling `MongoClient.connect`

This also ensure that connection strings that already have a database name will work:
>mongodb://examplemongo:xxx==@examplemongo.documents.azure.com:10255/**_admin_**?ssl=true

Let me know if you think it's worth merging before the release. I think it is, but I'm somewhat on the fence.


EricJizbaMSFT:
We already had one other fix go in this morning and we've got a few more, so it shouldn't hurt to merge this. We have to do more testing anyways

PR #354, opened 2018-02-01
Delete CosmosDB accounts from the extension

### LINKED ISSUE #348: Delete accounts in extension
I'd like to be able to delete my account through the extension instead of having to go through the portal.

PrashanthCorp:
Fixes #348

### DIFF
--- a/src/commands/deleteCosmosDBAccount.ts
+++ b/src/commands/deleteCosmosDBAccount.ts
@@ -1,3 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
 import * as vscode from 'vscode';
 import CosmosDBManagementClient = require("azure-arm-cosmosdb");
 import { IAzureNode } from 'vscode-azureextensionui';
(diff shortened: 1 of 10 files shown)


PrashanthCorp:
From my initial testing, it looks like the account deletion is non-blocking. Also note that the order in which the operations are finished =/= order of operations started. 

image

src/commands/deleteCosmosDBAccount.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
```
EricJizbaMSFT: Add the copyright header

src/commands/deleteCosmosDBAccount.ts
```diff
@@ -0,0 +1,26 @@
...
+import CosmosDBManagementClient = require("azure-arm-cosmosdb");
+import { IAzureNode } from 'vscode-azureextensionui';
+import { DialogBoxResponses } from '../constants';
+import { azureUtils } from '../utils/azureUtils';
+import * as util from '../utils/vscodeUtils';
+import { UserCancelledError } from 'vscode-azureextensionui';
+
+export async function deleteCosmosDBAccount(node: IAzureNode): Promise<void> {
+    const message: string = `Are you sure you want to delete account '${node.treeItem.label}'?`;
```
EricJizbaMSFT: Add 'and its contents' to the end of the message

src/commands/deleteCosmosDBAccount.ts
```diff
@@ -0,0 +1,26 @@
...
+    const result = await vscode.window.showWarningMessage(message, DialogBoxResponses.Yes, DialogBoxResponses.Cancel);
+    if (result === DialogBoxResponses.Yes) {
+        const docDBClient = new CosmosDBManagementClient(node.credentials, node.subscription.subscriptionId);
+        const resourceGroup: string = azureUtils.getResourceGroupFromId(node.treeItem.id);
+        const label: string = node.treeItem.label;
+        const accountName: string = label.substring(0, label.indexOf(" "));
```
EricJizbaMSFT: I would rather get the accountName from node.treeItem.id than node.treeItem.label (The id should never change, but we might change how the label is displayed - also it could be different formatting for different languages)

Take a look at `azureUtils.getResourceGroupFromId`. You can do something similar to get the accountName

src/commands/deleteCosmosDBAccount.ts
```diff
@@ -0,0 +1,26 @@
...
+        const output = util.getOutputChannel();
+        output.appendLine(`Starting removal of account ${accountName}`);
```
EricJizbaMSFT: Please use the same format as the [other extensions](https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/SiteWrapper.ts#L131)

Something like:
`Deleting account "${accountName}"...`
`Successfully deleted account "${accountName}".`

src/extension.ts
```diff
@@ -55,6 +56,12 @@ export function activate(context: vscode.ExtensionContext) {
 
 		await node.createChild();
 	});
+	actionHandler.registerCommand('cosmosDB.deleteAccount', async (node?: IAzureNode) => {
+		if (node) {
```
EricJizbaMSFT: Use the `showNodePicker` method in case the user runs this command from the command palette. Take a look at 'cosmosDB.openInPortal' below. It should be the exact same

src/extension.ts
```diff
@@ -27,6 +27,7 @@ import { DocDBAccountTreeItemBase } from './docdb/tree/DocDBAccountTreeItemBase'
 import { GraphAccountTreeItem } from './graph/tree/GraphAccountTreeItem';
 import { DocDBAccountTreeItem } from './docdb/tree/DocDBAccountTreeItem';
 import { TableAccountTreeItem } from './table/tree/TableAccountTreeItem';
+import { deleteCosmosDBAccount } from './commands/deleteCosmosDBAccount';
```
EricJizbaMSFT: nit: Remove this import - it's not used

PR #367, opened 2018-02-22
Fix query filters in find() with a JSON parse

### LINKED ISSUE #365: Scapbook find caused error "query selector must be an object"
Hello,
ok I'am new in mongodb but I whant to teach my students (I'am a teacher) how to use MongoDB (as an example for a nosql database).

So why the command 'db.schueler.find({"name": "tuttas"})' in the Srapbook cause the error "query selector must be an object" !?! in the shell mongodb.exe everythink workes like expected, but I what to use the Scapbook in VS Code as the shell!

Looking forward to hear from you.

best regrads

PrashanthCorp:
Fixes #365


PrashanthCorp:
Ah, `JSON.parse` doesn't like `undefined` values.

PR #373, opened 2018-02-27
Enforce noUnusedLocals in src

PrashanthCorp:


test/arrayUtil.test.ts
```diff
@@ -21,11 +15,11 @@ suite("arrayUtil Tests", () => {
             { id: "id1", data: "data1" },
             { id: "id2", data: "data2" },
             { id: "id2", data: "data2" },
-            { id: "id1", data: "data1" },
-            { id: "id2", data: "data2" }
+                    { id: "id1", data: "data1" },
```
PrashanthCorp: Oops. Will fix in next commit.

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -40,7 +39,7 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 			};
 		});
 
-		const result = await this.collection.bulkWrite(operations);
+		await this.collection.bulkWrite(operations);
```
StephenWeatherford: Have you verified this result is something okay to ignore?

src/graph/client/graphClient.ts
```diff
@@ -11,7 +11,7 @@
  */
 
 declare let d3: any;
-
+/* tslint:disable */
```
StephenWeatherford: This disables all lint, right? I'd rather you disabled just certain rules.  And it should be at the top of the file.

src/mongo/tree/MongoCollectionTreeItem.ts
PrashanthCorp: I placed this line back in [this commit](https://github.com/Microsoft/vscode-cosmosdb/blame/21c85e9cc850a7ed188abca3a0d36cb2fcabfb0b/src/mongo/nodes.ts#L205). I placed it solely for debugging the code. 
Tested it again, see no issues atm.

src/graph/client/graphClient.ts
PrashanthCorp: Had a second look at it. I'll remove it since I won't need it at this moment.

src/mongo/shell.ts
```diff
@@ -99,7 +99,7 @@ export class Shell {
 				disposable.dispose();
 				let lines = (<string>result.result).split(os.EOL).filter(line => !!line && line !== 'Type "it" for more');
 				lines = lines[lines.length - 1] === 'Type "it" for more' ? lines.splice(lines.length - 1, 1) : lines;
-				let value = lines.join(os.EOL);
+				lines.join(os.EOL);
```
EricJizbaMSFT: Just remove this line. We already do `lines.join(os.EOL)` on line 104

src/graph/GraphViewsManager.ts
```diff
@@ -37,7 +34,7 @@ export class GraphViewsManager implements IServerProvider {
     config: GraphConfiguration
   ): Promise<void> {
     try {
-      var [id, server] = await this.getOrCreateServer(config);
+      var [id,] = await this.getOrCreateServer(config);
```
EricJizbaMSFT: nit: You could change the signature of getOrCreateServer to return only id

src/mongo/tree/MongoCollectionTreeItem.ts
EricJizbaMSFT: I don't think we should ignore this. You could at least print something like this to the output window: `Successfully updated ${result.modifiedCount} document(s).`

src/graph/registerGraphCommands.ts
```diff
@@ -12,7 +12,6 @@ import { GraphViewsManager } from "./GraphViewsManager";
 
 export function registerGraphCommands(context: vscode.ExtensionContext, actionHandler: AzureActionHandler, tree: AzureTreeDataProvider): void {
     let graphViewsManager = new GraphViewsManager(context, actionHandler);
-    context.subscriptions.push(this.graphView);
```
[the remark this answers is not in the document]
StephenWeatherford: this can't be null, or nothing in the graph would be working. Definitely a bug, though. I think you can delete that line. Thx.

src/graph/GraphViewsManager.ts
StephenWeatherford: I'd be fine with that.

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -39,7 +40,9 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 			};
 		});
 
-		await this.collection.bulkWrite(operations);
+		const result: BulkWriteOpResultObject = await this.collection.bulkWrite(operations);
```
StephenWeatherford: Is there a deleted doc count?

PrashanthCorp: There is one. However, this update operation does not delete documents. It looks at each document in the collection view and performs an update _if that document is already present in the collection_ (notice the `upsert` beign false). 
The end result is that there are no deletions. I didn't see the point of consistently printing `Deleted 0 documents`.

PR #374, opened 2018-02-28
Enforce noImplicitThis and noImplicitReturns on src

PrashanthCorp:


src/commands/createCosmosDBAccount.ts
```diff
@@ -70,38 +70,28 @@ async function createResourceGroup(subscriptionNode: IAzureNode): Promise<Resour
                 const resourceManagementClient = new ResourceManagementClient(subscriptionNode.credentials, subscriptionNode.subscription.subscriptionId);
```
StephenWeatherford: I assume you've tested all the changed scenarios?  I wonder if it's not better to keep the current behavior, though, instead of throwing exceptions.

EricJizbaMSFT: This will probably change anyways once we move the ui stuff to the shared package (https://github.com/Microsoft/vscode-azuretools/issues/23) something I'd like to do soon after this release

StephenWeatherford:
@EricJizbaMSFT @PrashanthCorp Eric, how confident are you in letting this get checked in before the release?  I'd prefer to delay.

EricJizbaMSFT:
We have a separate release branch: https://github.com/Microsoft/vscode-cosmosdb/tree/rel/0.5.0

None of the lint stuff will be included

StephenWeatherford:
Ah.  Okay, then.  Is that just this extension that has a separate release branch?

EricJizbaMSFT:
Yeah - we did it specifically because of the lint work

PR #375, opened 2018-02-28
Leverage updated AzureActionHandler

EricJizbaMSFT:
Based on the following PR: https://github.com/Microsoft/vscode-azuretools/pull/88

I don't see any reason to include this in the release branch, so just basing the PR off master


src/CosmosEditorManager.ts
```diff
@@ -110,13 +110,14 @@ export class CosmosEditorManager {
...
-    public async onDidSaveTextDocument(trackTelemetry: () => void, globalState: vscode.Memento, doc: vscode.TextDocument, tree: AzureTreeDataProvider): Promise<void> {
+    public async onDidSaveTextDocument(context: IActionContext, globalState: vscode.Memento, doc: vscode.TextDocument, tree: AzureTreeDataProvider): Promise<void> {
+        context.sendTelemetry = false;
         let filePath = Object.keys(this.fileMap).find((filePath) => path.relative(doc.uri.fsPath, filePath) === '');
         if (!filePath) {
             filePath = await this.loadPersistedEditor(doc.uri, tree);
         }
         if (!this.ignoreSave && filePath) {
-            trackTelemetry();
+            context.sendTelemetry = true;
```
PrashanthCorp: If I understand this correctly, this would send telemetry only on files which we handle. 

@StephenWeatherford is there a similar aim in the pipeline for the arm extension?

StephenWeatherford: I think Jay already fixed that, but I'll have to check.  Eric's changes are just a refactoring, shouldn't modify current behavior.

src/graph/GraphViewServer.ts
```diff
@@ -135,50 +136,51 @@ export class GraphViewServer extends EventEmitter {
     const start = Date.now();
 
     try {
-      await this._actionHandler.callWithTelemetry("cosmosDB.gremlinQuery", async (telemetryProperties, measurements) => {
-        this._pageState = {
+      const graphViewServier: GraphViewServer = this;
```
StephenWeatherford: nit: a common pattern for this is:
  let me = this;
or
  let thisServer = this;

StephenWeatherford: Also, you misspelled server :-)

src/graph/GraphViewServer.ts
```diff
@@ -135,50 +136,51 @@ export class GraphViewServer extends EventEmitter {
...
+      await AzureActionHandler.callWithTelemetry("cosmosDB.gremlinQuery", reporter, async function (this: IActionContext): Promise<void> {
```
StephenWeatherford: I'm liking the "this" thing less now that I see it. :-(  But not worth rehashing the other PR for.

Might consider immediately doing:
  let actionContext = this;

EricJizbaMSFT: Yeah I'm open to changing it, but I agree it's not worth re-hashing right now

PR #376, opened 2018-03-02
EnableGremlin capability to create post-GA graphs, fixes #352

### LINKED ISSUE #352: Graph accounts created in extension are pre-GA versions
Need to create post-GA accounts for performance and other issues

StephenWeatherford:
Unfortunately, this requires a new version of the cosmosdb SDK built using the current swagger, and it has lots of changes.  Not sure how high a risk it is to take this right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -711,7 +711,7 @@
 	},
 	"dependencies": {
 		"antlr4ts": "^0.4.0-alpha.4",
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
Turns out we're broken due to portal changes - graphs created using the extension will not have a working gremlin endpoint without this, so need this fix anyway.

PR #396, opened 2018-03-07
Leverage new ui package for tree telemetry

EricJizbaMSFT:


src/extension.ts
```diff
@@ -31,7 +31,7 @@ import { TableAccountTreeItem } from './table/tree/TableAccountTreeItem';
 export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(new Reporter(context));
 
-	const tree: AzureTreeDataProvider = new AzureTreeDataProvider(new CosmosDBAccountProvider(), 'cosmosDB.loadMore', [new AttachedAccountsTreeItem(context.globalState)]);
```
StephenWeatherford: Ideally we should be making these types of things required parameters.

PR #411, opened 2018-03-16
Markdown spell/lint

StephenWeatherford:


.vscode/settings.json
```diff
@@ -9,5 +9,16 @@
 	"search.exclude": {
 		"out": true // set this to false to include "out" folder in search results
 	},
-	"typescript.tsdk": "./node_modules/typescript/lib" // we want to use the TS server from our node_modules folder to control its version,
+	"typescript.tsdk": "./node_modules/typescript/lib", // we want to use the TS server from our node_modules folder to control its version
+	"markdownlint.config": {},
+	"cSpell.words": [
```
EricJizbaMSFT: What is this setting used for?

StephenWeatherford: It's the speller checker.

EricJizbaMSFT: I'm still confused. For example, are you saying "cosmosdb" is _okay_ or are you saying "cosmosdb" is _not_ okay (aka are you trying to enforce "cosmos db" with a space)?

Also confused by "Intelli"- where would we have that except as the whole word "Intellisense"

PR #412, opened 2018-03-19
Refactor createCosmosDBAccount to use shared wizard

EricJizbaMSFT:
Leverages the shared resource group/location/name steps and refactors other stuff to be actual Wizard steps.

In addition to reducing the amount of copied code, this adds support for the following:
1. Persisting recently used selections
1. Generating a related name for the resource group
1. Consistent use of ignoreFocusOut and UserCancelledError by default


src/extension.ts
```diff
@@ -31,7 +31,9 @@ import { TableAccountTreeItem } from './table/tree/TableAccountTreeItem';
 export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(new Reporter(context));
 
-	const tree: AzureTreeDataProvider = new AzureTreeDataProvider(new CosmosDBAccountProvider(), 'cosmosDB.loadMore', [new AttachedAccountsTreeItem(context.globalState)], reporter);
+	const ui: IAzureUserInput = new AzureUserInput(context.globalState);
+
+	const tree: AzureTreeDataProvider = new AzureTreeDataProvider(new CosmosDBAccountProvider(), 'cosmosDB.loadMore', ui, reporter, [new AttachedAccountsTreeItem(context.globalState)]);
```
PrashanthCorp: Looks like the order of parameters has changed, with `reporter` called before the array. Would this new signature break other calls across the extensions?

EricJizbaMSFT: Yes its a breaking change. That's why we bumped the minor version for the vscode-azureextensionui package from 8 to 9

src/tree/CosmosDBAccountWizard/CosmosDBAccountApiStep.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { AzureWizardStep, IAzureUserInput, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { ICosmosDBWizardContext } from './ICosmosDBWizardContext';
+import { Experience, DBAccountKind } from '../../constants';
+
+export class CosmosDBAccountApiStep extends AzureWizardStep<ICosmosDBWizardContext> {
+    public async prompt(wizardContext: ICosmosDBWizardContext, ui: IAzureUserInput): Promise<ICosmosDBWizardContext> {
+        const picks: IAzureQuickPickItem<string>[] = [
+            { label: Experience.MongoDB, description: '', data: DBAccountKind.MongoDB },
+            { label: Experience.DocumentDB, description: '', data: DBAccountKind.GlobalDocumentDB },
```
StephenWeatherford: Shouldn't this be SQL now?  With a description that it used to be docdb?

src/tree/CosmosDBAccountWizard/CosmosDBAccountApiStep.ts
```diff
@@ -0,0 +1,32 @@
...
+            { label: Experience.Graph, description: '', data: DBAccountKind.GlobalDocumentDB },
```
StephenWeatherford: Likewise Graph and Table are shown differently in the portal now:

image

src/tree/CosmosDBAccountWizard/CosmosDBAccountApiStep.ts
EricJizbaMSFT: I think the whole renaming business is very confusing and I don't want to deal with it 🙃

FIled an issue to address separately: #415

PR #416, opened 2018-03-21
View collection stored procs in the browser tree

### LINKED ISSUE #74: View collection stored procs in the browser tree
These should show up as nodes under each Cosmos DB account (all APIs support them) with the option to click the node and open the associated file.

Note: Persisting these is covered in issue #38

StephenWeatherford:
Fixes #74 

image
image


package.json
```diff
@@ -523,6 +523,14 @@
 					"command": "cosmosDB.refresh",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentDatabase"
 				},
+				{
```
EricJizbaMSFT: nit: You could probably remove the refresh action from the 'cosmosDBDocumentCollection' item since it just has static nodes underneath it

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    public get iconPath(): any {
+        return {
+            light: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'icons', 'theme-agnostic', 'Collection.svg'),
+            dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'icons', 'theme-agnostic', 'Collection.svg')
+        };
+    }
+
+    public get id(): string {
+        return "$Documents";
```
EricJizbaMSFT: nit: Remember id is an optional field - you don't have to specify it in in a case like this where the label should already be unique at this level of the tree

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -0,0 +1,50 @@
...
+        return "$StoredProcedures";
+    }
+
+    public get label(): string {
+        return "Stored Procedures";
+    }
+
+    public get link(): string {
+        return this._collection._self;
+    }
```
PrashanthCorp: I see both `DocDBDocumentsTreeItem` and `DocDBStoredProceduresTreeItem` have the same self-link. If user-defined-functions and triggers are in the pipeline, I'd expect the container treeItem(s) to have the same self-link too (that of the collection). In order to reduce code repetition, would it make sense to have `DocDBDocumentsTreeItem` and `DocDBStoredProceduresTreeItem` inherit this property?

package.json
StephenWeatherford: Yep, thanks. Originally docs were in this node, too.

src/docdb/tree/DocDBDocumentsTreeItem.ts
StephenWeatherford: Abstract in the inherited code, not worth changing.

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
StephenWeatherford: It's calling the parent's method, so not really duplicated code. I think this is better than adding it to the super for this small amount.

package.json
StephenWeatherford: Actually as I user I'd rather have Refresh at all levels for consistency.

Issue #316, opened 2018-01-05
Add tslint


EricJizbaMSFT:
Here's my thoughts on how best to split it up:
[x] 1. [MOVED] Add basic typescript rules (see settings like 'noUnusedLocals' [here](https://github.com/Microsoft/vscode-azurefunctions/blob/master/tsconfig.json))
[x] 1. Add basic tslint (see lint script [here](https://github.com/Microsoft/vscode-azurefunctions/blob/master/package.json#L367))
[x] 1. Add microsoft contributed rules (see [here](https://github.com/Microsoft/vscode-azurefunctions/blob/master/tslint.json))

Remaining rules:
[ ] P1: no-unsafe-any
[ ] P1: strict null checking
[ ] no-floating-promises
[ ] promise-function-async - [Stephen] suggest grandfather in existing code (disable per line) and turning on
[ ] no-string-literal   - [Stephen] suggest turn this off, it's mostly for localization
[ ] member-access   - [Stephen] I think can auto-fix

StephenWeatherford:
> Adding a checklist of remaining rules. Please suggest removing some rules or changing the ordering. I want to decide on what rules to enable for 0.10.0 before we get around to it.

Moved to description so we don't have to search through whole issue.  Added suggestions.

EricJizbaMSFT:
Gonna close this as fixed after merging https://github.com/microsoft/vscode-cosmosdb/pull/1316, https://github.com/microsoft/vscode-cosmosdb/pull/1346, and https://github.com/microsoft/vscode-cosmosdb/pull/1380. Always more we could do, but don't think it's worth having the issue open any more.

PR #426, opened 2018-03-22
Add tslint (with some disabled rules)

StephenWeatherford:
Start of #316

### DIFF
--- a/package.json
+++ b/package.json
@@ -488,14 +488,6 @@
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBGraphAccount"
 				},
 				{
-					"command": "cosmosDB.openInPortal",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBStoredProceduresGroup"
-				},
-				{
-					"command": "cosmosDB.openInPortal",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBStoredProcedure"
-				},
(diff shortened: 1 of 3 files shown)


tslint.json
```diff
@@ -0,0 +1,318 @@
...
+        "no-empty-interfaces": false,
+        "no-missing-visibility-modifiers": false,
+        "no-multiple-var-decl": false,
+        "no-switch-case-fall-through": false,
+        "typeof-compare": false
+    },
+    "rulesDirectory": "node_modules/tslint-microsoft-contrib/",
+    "linterOptions": {
+        "exclude": [
+            "src/diagnostics/**", // should be handled by the Log Points team
```
EricJizbaMSFT: Remove these exclusions - I believe they were only applicable to the app service repo

package.json
```diff
@@ -487,6 +487,14 @@
 					"command": "cosmosDB.openInPortal",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBGraphAccount"
 				},
+				{
+					"command": "cosmosDB.openInPortal",
```
EricJizbaMSFT: IMO its more confusing than valuable if the open in portal links to the Cosmos DB Account instead of directly to the stored procedure

StephenWeatherford: Whoops, shouldn't be in there.

PR #431, opened 2018-03-22
Delinting step 1

StephenWeatherford:
#316 

Some of these rules I'm going to grandfather in because they're no value and some of this code is not well understood by us yet and I'd rather not be messing with it.

### DIFF
--- a/src/graph/client/graphClient.ts
+++ b/src/graph/client/graphClient.ts
@@ -133,6 +133,7 @@ export class GraphClient {
     this.setStateInitial();
 
     this.log(`Listening on port ${port}`);
+    // tslint:disable-next-line:no-http-string
     this._socket = new SocketWrapper(io.connect(`http://localhost:${port}`));
 
     // setInterval(() => {
--- a/src/graph/client/graphTypes.ts
+++ b/src/graph/client/graphTypes.ts
@@ -7,6 +7,8 @@
  * These types are shared between server and client code
  */
 
+// tslint:disable:no-reserved-keywords (This code is deprecated anyway)
+
 interface GraphResults {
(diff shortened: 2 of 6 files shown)


src/mongo/grammar/mongoParser.ts
```diff
@@ -334,6 +334,7 @@ export class mongoParser extends Parser {
 		}
 		return _localctx;
 	}
+	// tslint:disable-next-line:no-banned-terms
```
PrashanthCorp: What's the banned term here?

src/mongo/grammar/mongoParser.ts
```diff
@@ -1022,6 +1023,7 @@ export class CollectionContext extends ParserRuleContext {
 
 export class FunctionCallContext extends ParserRuleContext {
 	public _FUNCTION_NAME: Token;
+	// tslint:disable-next-line:no-banned-terms
 	public arguments(): ArgumentsContext {
```
PrashanthCorp: Is there a list of banned terms I can look at?

src/mongo/grammar/mongoParser.ts
StephenWeatherford: arguments

src/mongo/grammar/mongoParser.ts
StephenWeatherford: This is the best I can find: https://github.com/Microsoft/tslint-microsoft-contrib/wiki/TSLint-and-the-Microsoft-Security-Development-Lifecycle

You could maybe call this a false positive since the scope is not global.

However, I see now this is a generated file, so I'll exclude it.

PR #446, opened 2018-03-26
Re-opening VSCode multiple times now shouldn't break update document

### LINKED ISSUE #370: Failed to update mongo document after closing and re-opening VS Code multiple times
1. Open a mongo document
1. Close vscode
1. Open vscode
1. Open a mongo collection
1. Close vscode
1. Open vscode
1. Try to update the mongo _document_

Expected:
It works

Actual:
It fails with this error:
screen shot 2018-02-23 at 2 57 17 pm

I believe it's because we change the persistence to only include the mongo collection and then we lose persistence of the mongo document.

PrashanthCorp:
Fixes #370

### DIFF
--- a/src/CosmosEditorManager.ts
+++ b/src/CosmosEditorManager.ts
@@ -49,7 +49,7 @@ export class CosmosEditorManager {
             }
         }
         this.fileMap[localDocPath] = editor;
-        const fileMapLabels = {};
+        const fileMapLabels = this._globalState.get(this._persistedEditorsKey);
         Object.keys(this.fileMap).forEach((key) => fileMapLabels[key] = (this.fileMap[key]).id);
         this._globalState.update(this._persistedEditorsKey, fileMapLabels);
         const textEditor = await vscode.window.showTextDocument(document);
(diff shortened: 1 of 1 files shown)


src/CosmosEditorManager.ts
```diff
@@ -49,7 +49,7 @@ export class CosmosEditorManager {
             }
         }
         this.fileMap[localDocPath] = editor;
-        const fileMapLabels = {};
+        const fileMapLabels = this._globalState.get(this._persistedEditorsKey);
```
EricJizbaMSFT: Hahaha didn't we just learn that this can return undefined? 🙃

You could just do `const fileMapLabels = this._globalState.get(this._persistedEditorsKey) || {};`

StephenWeatherford: Yep, you should think about stuff like this always.

BTW, have you searched for other uses of this._globalState.get() that might cause problems?

PrashanthCorp: @EricJizbaMSFT I'm going with [get<T>(key: string, defaultValue: T): T](https://code.visualstudio.com/docs/extensionAPI/vscode-api#Memento) - that looks cleaner and more official to me. 

@StephenWeatherford - Looked at issues filed and StackOverflow, I don't see anything surprising so far.

StephenWeatherford: @PrashanthCorp I was referring to if you've checked all uses of globalState.get() in our code...

PR #448, opened 2018-03-27
Delinting 4

StephenWeatherford:


src/graph/GraphViewServer.ts
```diff
@@ -326,7 +329,7 @@ export class GraphViewServer extends EventEmitter {
           if (err.code === "ENOTFOUND") {
             // Not a valid endpoint
           } else {
-            firstValidError = firstValidError || err;
+            firstValidError = firstValidError || err || Object
```
EricJizbaMSFT: What's the purpose of `|| Object` here? Also - missing a semicolon at the end of the line

src/graph/GraphViewServer.ts
```diff
@@ -282,6 +284,7 @@ export class GraphViewServer extends EventEmitter {
     return message;
   }
 
+  // tslint:disable-next-line:no-any
   private async executeQuery(queryId: number, gremlinQuery: string): Promise<any[]> {
```
EricJizbaMSFT: nit: I don't really care if you use `any[]` or `Object[]`, but I think `executeQuery`, `_executeQueryCore`, and `_executeQueryCoreForEndpoint` should all have the same return type

src/graph/GraphViewServer.ts
```diff
@@ -373,7 +377,7 @@ export class GraphViewServer extends EventEmitter {
       socketError = err;
     }
 
-    return new Promise<[{}[]]>((resolve, reject) => {
+    return new Promise<[Object[]]>((resolve, reject) => {
```
EricJizbaMSFT: nit: Again, I don't care if you use `any[]` or `Object[]`, but I think this should match the return type of its function

src/graph/GraphViewServer.ts
EricJizbaMSFT: If I had to choose, I'd probably pick `any[]` and just disable "no-any" for this file. I don't think `Object[]` is helping us at all and you end up casting it to `any[]` eventually anyways

src/mongo/services/schemaService.ts
```diff
@@ -103,7 +103,7 @@ export default class SchemaService {
 		})
 	}
 
-	private setSchemaForDocument(parent: string, document: any, schema: JSONSchema): void {
+	private setSchemaForDocument(parent: string, document: {}, schema: JSONSchema): void {
```
EricJizbaMSFT: Based on the line right below this, it feels like `document` could be multiple different types, including array. I think `any` is more accurate than `{}`

EricJizbaMSFT: This is another case where I might prefer to just disable "no-any" for the file - especially since we inherited this code and we're pretty unfamiliar with it

src/graph/GraphViewServer.ts
StephenWeatherford: Thanks for catching that.

src/mongo/services/schemaService.ts
StephenWeatherford: Object is what the SDK function that it called was declared as expecting. However, I'm good to just disable it.

Issue #386, opened 2018-03-05
Can only run the first command in a Mongo Scrapbook on Windows


EricJizbaMSFT:
As far as I can tell, this only repros on Windows

Repro steps:
1. Connect to a DB
1. Open a mongo scrapbook
1. Add two commands:
```
db.test.find()
db.test.findOne()
```
1. Put your cursor in the second command
1. Execute command

Expected: Results of findOne are displayed
Actual: Results of find are displayed

EricJizbaMSFT:
This repros on the old Mongo extension, too. Must've always been a problem

StephenWeatherford:
@PrashanthCorp Can you repro and verify if it only repros on Windows?

PR #454, opened 2018-03-29
README fixes/improvements, fixes #453

### LINKED ISSUE #453: Instructions for scrapbook in README are incorrect
"- Click on any DB to open the Mongo shell playground editor" - that's not correct

Also, it doesn't say anything about "connecting" to a database and what that means, which is also very confusing.

StephenWeatherford:


README.md
```diff
@@ -21,27 +21,19 @@
...
+## Mongo "Scrapbooks" (Run Mongo Commands with Rich Intellisense)
 
-## Managing Azure Subscriptions
-
-If you are not signed in to Azure, you will see a "Sign in to Azure..." link. Alternatively, you can select "View->Command Palette" in the VS Code menu, and search for "Azure: Sign In".
-
-![Sign in to Azure](resources/SignIn.gif)
-
-If you don't have an Azure Account, you can sign up for one today for free and receive $200 in credits by selecting "Create a Free Azure Account..." or selecting "View->Command Palette" and searching for "Azure: Create an Account".
+- Attach to a MongoDB database account
```
EricJizbaMSFT: How about "View your MongoDB database account by signing in to Azure or attaching with a connection string"

If you just say "Attach" that leaves out the Azure scenario

README.md
```diff
@@ -21,27 +21,19 @@
...
+- Optionally configure the user setting `mongo.shell.path` to your mongo shell executable path (many of the common commands have built-in support and do not require the Mongo shell to be installed - see [Prerequisites](#prerequisites))
+- Right-click on a database and select "Connect"
+- Click on "New Mongo Scrapbook" in the tree title bar
+- Enter your script, eg: `db.<collectionName>.find()`
```
EricJizbaMSFT: We should change all instances of 'script' to 'command'

Script implies multiple commands, but we only support one command at a time. "Command" is also consistent with our naming inside the extension:
screen shot 2018-03-29 at 9 06 34 am

README.md
```diff
@@ -21,27 +21,19 @@
...
+- IntelliSense (completions) will be provided as you write your script
+- Select the script and press `CMD+"` (Mac) or `CTRL+"` (Windows and Linux) to see the results
```
EricJizbaMSFT: "Select" doesn't seem accurate to me. How about, "With your cursor inside the command, press ..."

Also we really need to address this bug: https://github.com/Microsoft/vscode-cosmosdb/issues/386 otherwise this bullet point only applies for Mac/Linux

PR #456, opened 2018-03-29
Fix handling of CRLF and add some unit tests, fixes #386

### LINKED ISSUE #386: Can only run the first command in a Mongo Scrapbook on Windows

StephenWeatherford:
Also improved some of the error messages for the scrapbook


StephenWeatherford:
Also improved some of the error messages for the scrapbook

src/mongo/commands.ts
```diff
@@ -22,13 +22,14 @@ export class MongoCommands {
...
 		if (activeEditor.document.languageId !== 'mongo') {
+			vscode.window.showInformationMessage("This command can only be run inside of a MongoDB scrapbook (*.mongo)");
 			return;
 		}
 		const selection = activeEditor.selection;
 		const command = MongoCommands.getCommand(activeEditor.document.getText(), selection.start);
 		if (command) {
 			if (!database) {
-				throw new Error('Please connect to the database first');
+				throw new Error('Please select a MongoDB database to run against by selecting it in the explorer and selecting the "Connect" context menu item');
```
EricJizbaMSFT: Just an idea: rather than throwing an error, we could show a warning message that has a 'Select database' button that calls `await vscode.commands.executeCommand('cosmosDB.connectMongoDB');`. It would use the node picker to select the db.

You'd have to make sure the command returned the connectDb so that you had access here, but I'm pretty sure that would be easy to do

test/mongoGetCommand.test.ts
```diff
@@ -0,0 +1,79 @@
+/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT: woo

PR #468, opened 2018-04-02
Delete stored procedures 

### LINKED ISSUE #465: Add ability to delete stored procedures

StephenWeatherford:
Fixes #465

### DIFF
--- a/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
+++ b/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
@@ -40,7 +40,7 @@ export class DocDBStoredProcedureTreeItem implements IAzureTreeItem {
     }
 
     public async deleteTreeItem(_node: IAzureNode): Promise<void> {
-        const message: string = `Are you sure you want to delete collection '${this.label}' and its contents?`;
+        const message: string = `Are you sure you want to delete stored procedure '${this.label}'?`;
         const result = await vscode.window.showWarningMessage(message, DialogBoxResponses.Yes, DialogBoxResponses.Cancel);
         if (result === DialogBoxResponses.Yes) {
             const client = getDocumentClient(this._documentEndpoint, this._masterKey, this._isEmulator);
--- a/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
+++ b/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
@@ -8,7 +8,6 @@ import * as vscode from "vscode";
 import { IAzureTreeItem, IAzureNode, UserCancelledError } from 'vscode-azureextensionui';
 import { ProcedureMeta } from 'documentdb';
 import { DialogBoxResponses } from '../../constants';
-import { DocDBCollectionTreeItem } from './DocDBCollectionTreeItem';
 import { getDocumentClient } from '../getDocumentClient';
 
 /**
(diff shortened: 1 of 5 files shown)


src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -36,4 +39,19 @@ export class DocDBStoredProcedureTreeItem implements IAzureTreeItem {
...
     }
+
+    public async deleteTreeItem(_node: IAzureNode): Promise<void> {
+        const message: string = `Are you sure you want to delete collection '${this.label}' and its contents?`;
+        const result = await vscode.window.showWarningMessage(message, DialogBoxResponses.Yes, DialogBoxResponses.Cancel);
+        if (result === DialogBoxResponses.Yes) {
+            const client = getDocumentClient(this._documentEndpoint, this._masterKey, this._isEmulator);
+            await new Promise((resolve, reject) => {
+                client.deleteStoredProcedure(this.link, function (err) {
+                    err ? reject(err) : resolve();
```
PrashanthCorp: For a destructive operation like this, would it make sense to write the result of this operation to the output,?

StephenWeatherford: Don't think we're doing it for anything else like that.  Will check.

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -15,7 +18,7 @@ export class DocDBStoredProcedureTreeItem implements IAzureTreeItem {
     public static contextValue: string = "cosmosDBStoredProcedure";
     public readonly contextValue: string = DocDBStoredProcedureTreeItem.contextValue;
 
-    constructor(private _procedure: ProcedureMeta) {
+    constructor(private _documentEndpoint: string, private _masterKey: string, private _isEmulator: boolean, private _procedure: ProcedureMeta) {
```
PrashanthCorp: 1. Do procedures have a unique endpoint? If yes, I prefer calling it something like `storedProcEndpoint`, `ProcedureEnpoint`, or something to that effect. :) 
2. Can the variables masterKey, and isEmulator be inferred from the parent?

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -36,4 +38,19 @@ export class DocDBStoredProcedureTreeItem implements IAzureTreeItem {
             dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'icons', 'dark', 'Process_16x.svg')
         };
     }
+
+    public async deleteTreeItem(_node: IAzureNode): Promise<void> {
+        const message: string = `Are you sure you want to delete collection '${this.label}' and its contents?`;
```
EricJizbaMSFT: 'collection' -> 'stored procedure'

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
StephenWeatherford: @fiveisprime Do you want us to print to the output window for every action?  We're currently doing it for deleting accounts but nothing else I think.

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
StephenWeatherford: Might make sense, but not going to worry about it now.

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
StephenWeatherford: fixed

PR #475, opened 2018-04-03
Persist connected Mongo DB

### LINKED ISSUE #3: MongoDB connection is not persisted

EricJizbaMSFT:
Fixes #3


StephenWeatherford:
Thanks!

PR #476, opened 2018-04-03
Handle connection strings with database specified, fixes #473

### LINKED ISSUE #473: Allow access to multi-tenant mongo server
While testing Microsoft/vscode#25332

When you run your database on mlab you get a multi-tenant setup. I.e. a single mongo server has multiple databases but you are only allowed to access the db you own.

The mongoshell handles that case without issues:
mongo mongodb://<dbuser>:<dbpassword>@<server>.mlab.com:55574/<database> works without issues.

When I try to use such a connection string the in extension I get an error since the extension tries to list the databases:

not authorized on admin to execute command { listDatabases: 1 }: MongoError: not authorized on admin to execute command { listDatabases: 1 }

StephenWeatherford:


src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -43,11 +44,24 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
...
         let db: Db | undefined;
         try {
+            let databases: IDatabaseInfo[];
+
             db = await MongoClient.connect(this.connectionString);
-            const result: { databases: IDatabaseInfo[] } = await db.admin().listDatabases();
-            return result.databases
+            let databaseInConnectionString = getDatabaseNameFromConnectionString(this.connectionString);
+            if (databaseInConnectionString && !this.isEmulator) { // emulator violates the connection string format
+                // If the database is in the connection string, that's all we connect to (we might not even have permissions to list databases)
```
EricJizbaMSFT: I'm fine with this as long as the connection strings from Azure do not list a database by default. Is that the case? I just want to make sure Azure Cosmos DB account still list all databases

StephenWeatherford: That's the case for both connected and attached scenarios.

PR #483, opened 2018-04-04
Add DB to id for multi-tenant scenario, fixes #482

### LINKED ISSUE #482: Can't attach to different database names in same server in multi-tenant scenario

StephenWeatherford:
image

### DIFF
--- a/src/tree/AttachedAccountsTreeItem.ts
+++ b/src/tree/AttachedAccountsTreeItem.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import * as keytarType from 'keytar';
-import { MongoClient, ReplSet } from "mongodb";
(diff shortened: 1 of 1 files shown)


src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -6,14 +6,15 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import * as keytarType from 'keytar';
-import { MongoClient, ReplSet } from "mongodb";
+import { MongoClient, ReplSet, connect } from "mongodb";
```
EricJizbaMSFT: Why'd ya add 'connect'? Actually the build might be failing because of this 'unused variable'

StephenWeatherford: It has a mind of its own...

PR #484, opened 2018-04-04
Add scrapbook command name to telemetry

StephenWeatherford:


src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -105,6 +105,7 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
 		if (command.name === 'createCollection') {
 			return reportProgress(this.createCollection(stripQuotes(command.arguments.join(','))).then(() => JSON.stringify({ 'Created': 'Ok' })), 'Creating collection');
 		} else {
+			context.properties["executeInShell"] = "true";
```
EricJizbaMSFT: There's another 'executeCommandInShell' call a few lines up. Make sure to set it there, too

StephenWeatherford: Moved telemetry to 'executeCommandInShell' call itself, thanks.

PR #491, opened 2018-04-04
Fix '(Connected)' label bugs

### LINKED ISSUE #481: Persisted connected node does not display Connected in its label on startup

### LINKED ISSUE #487: Refreshing Mongo Account removes '(Connected)' from database node
Repro Steps:
1. Connect to a mongo database
1. Refresh the mongo account node

Expected:
Database node still has '(Connected)' in label

Actual:
Database node does _not_ have '(Connected)' in label, but it technically still is connected (commands in the scrapbook still work)

EricJizbaMSFT:
Fixes #487 
Fixes #481 

Requires this PR: https://github.com/Microsoft/vscode-azuretools/pull/148

### DIFF
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -1,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IAzureNode } from "vscode-azureextensionui";
+
+/**
+ * Namespace for common variables used throughout the extension. They must be initialized in the activate() method of extension.ts
+ */
+export namespace ext {
+    export let connectedMongoDB: IAzureNode | undefined;
+}
--- a/src/mongo/registerMongoCommands.ts
+++ b/src/mongo/registerMongoCommands.ts
@@ -14,11 +14,14 @@ import { MongoCommands } from "./commands";
 import { MongoDocumentTreeItem } from "./tree/MongoDocumentTreeItem";
 import { MongoCollectionNodeEditor } from "./editors/MongoCollectionNodeEditor";
 import { CosmosEditorManager } from "../CosmosEditorManager";
+import { ext } from "../extensionVariables";
+
+const connectedDBKey: string = 'ms-azuretools.vscode-cosmosdb.connectedDB';
 
 export function registerMongoCommands(context: vscode.ExtensionContext, actionHandler: AzureActionHandler, tree: AzureTreeDataProvider, editorManager: CosmosEditorManager): void {
-    let connectedDb: IAzureParentNode<MongoDatabaseTreeItem> = null;
     let languageClient: MongoDBLanguageClient = new MongoDBLanguageClient(context);
-    const connectedDBKey: string = 'ms-azuretools.vscode-cosmosdb.connectedDB';
+
+    const loadPersistedMongoDBTask: Promise<void> = loadPersistedMongoDB(context, tree);
 
     actionHandler.registerCommand('cosmosDB.createMongoDatabase', async (node?: IAzureParentNode) => {
         if (!node) {
@@ -44,23 +47,20 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
         if (!node) {
             node = <IAzureParentNode<MongoDatabaseTreeItem>>await tree.showNodePicker(MongoDatabaseTreeItem.contextValue);
         }
-        if (connectedDb) {
-            connectedDb.treeItem.isConnected = false;
-            await connectedDb.refresh();
-        }
-        connectedDb = node;
-        await languageClient.connect(connectedDb.treeItem.connectionString, connectedDb.treeItem.databaseName);
-        connectedDb.treeItem.isConnected = true;
-        await node.refresh();
+
+        await languageClient.connect(node.treeItem.connectionString, node.treeItem.databaseName);
         context.globalState.update(connectedDBKey, node.id);
+        ext.connectedMongoDB = node;
+
+        await tree.refresh(node.parent);
     });
     actionHandler.registerCommand('cosmosDB.deleteMongoDB', async (node?: IAzureNode<MongoDatabaseTreeItem>) => {
         if (!node) {
             node = <IAzureNode<MongoDatabaseTreeItem>>await tree.showNodePicker(MongoDatabaseTreeItem.contextValue);
         }
         await node.deleteNode();
-        if (connectedDb && connectedDb.treeItem.id === node.treeItem.id) {
-            connectedDb = null;
+        if (ext.connectedMongoDB && ext.connectedMongoDB.id === node.id) {
+            ext.connectedMongoDB = undefined;
             context.globalState.update(connectedDBKey, undefined);
             languageClient.disconnect();
         }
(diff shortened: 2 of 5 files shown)


src/mongo/registerMongoCommands.ts
```diff
@@ -44,23 +47,20 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
...
-        connectedDb = node;
-        await languageClient.connect(connectedDb.treeItem.connectionString, connectedDb.treeItem.databaseName);
-        connectedDb.treeItem.isConnected = true;
-        await node.refresh();
+
+        await languageClient.connect(node.treeItem.connectionString, node.treeItem.databaseName);
         context.globalState.update(connectedDBKey, node.id);
+        ext.connectedMongoDB = node;
+
+        await tree.refresh(node.parent);
```
StephenWeatherford: Don't think this will handle the case where you're switching the connection from an Azure node to an attached node?

src/mongo/registerMongoCommands.ts
```diff
@@ -86,20 +86,21 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
...
+        await MongoCommands.executeCommandFromActiveEditor(<IAzureParentNode<MongoDatabaseTreeItem>>ext.connectedMongoDB, context.extensionPath, editorManager, tree, this);
     });
 }
 
+async function loadPersistedMongoDB(context: vscode.ExtensionContext, tree: AzureTreeDataProvider): Promise<void> {
+    const persistedNodeId: string | undefined = context.globalState.get(connectedDBKey);
+    if (persistedNodeId) {
+        const persistedNode: IAzureNode | undefined = await tree.findNode(persistedNodeId);
+        if (persistedNode) {
+            await vscode.commands.executeCommand('cosmosDB.connectMongoDB', persistedNode);
```
StephenWeatherford: What if this fails?

src/mongo/registerMongoCommands.ts
EricJizbaMSFT: Good point

src/mongo/registerMongoCommands.ts
EricJizbaMSFT: It feels like worst case you just don't get the 'persisted' behavior and you have to reconnect? Can you think of anything worse? That doesn't seem too bad to me

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -20,24 +21,30 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
 	public readonly databaseName: string;
 
-	public isConnected: boolean = false;
+	private _isConnected: boolean;
 
-	constructor(databaseName: string, connectionString: string) {
+	constructor(databaseName: string, connectionString: string, parentId: string) {
 		this.databaseName = databaseName;
 		this.connectionString = connectionString;
+		this._isConnected = ext.connectedMongoDB && ext.connectedMongoDB.id === `${parentId}/${databaseName}`;
```
StephenWeatherford: Seems like you could get rid of setting _isConnected here if you turned it into a property get. Then also wouldn't need refreshLabel.

src/extensionVariables.ts
```diff
@@ -0,0 +1,13 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IAzureNode } from "vscode-azureextensionui";
+
+/**
+ * Namespace for common variables used throughout the extension. They must be initialized in the activate() method of extension.ts
+ */
+export namespace ext {
+    export let connectedMongoDB: IAzureNode | undefined;
```
StephenWeatherford: Do you actually need to keep a reference to the node instead of just the ID?  Would keeping only the ID break your dependency and allow you to reduce the complexity?

src/mongo/registerMongoCommands.ts
StephenWeatherford: You're awaiting it in execute, so it will fail every time.

src/extensionVariables.ts
EricJizbaMSFT: I need the node to fix your other comment :) (making sure I refresh the "old" node in case you switch from Azure to Attached)

StephenWeatherford: Can you search for the node?

StephenWeatherford: Doesn't matter much if this doesn't reduce complexity.

EricJizbaMSFT: Yes, but that feels more complex and slower than just keeping a reference to the node IMO

src/mongo/registerMongoCommands.ts
EricJizbaMSFT: Can expand on that? I don't know what you mean...are you confusing `cosmosDB.executeMongoCommand` or `MongoCommands.executeCommandFromActiveEditor` with `vscode.commands.executeCommand`

EricJizbaMSFT: Never mind - I get you

EricJizbaMSFT: Fixed

src/mongo/tree/MongoDatabaseTreeItem.ts
EricJizbaMSFT: Done

src/mongo/registerMongoCommands.ts
```diff
@@ -44,23 +48,24 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
...
-        connectedDb = node;
-        await languageClient.connect(connectedDb.treeItem.connectionString, connectedDb.treeItem.databaseName);
-        connectedDb.treeItem.isConnected = true;
-        await node.refresh();
+
+        const oldNode: IAzureNode | undefined = ext.connectedMongoDB;
+        await languageClient.connect(node.treeItem.connectionString, node.treeItem.databaseName);
         context.globalState.update(connectedDBKey, node.id);
+        ext.connectedMongoDB = node;
+        await tree.refresh(node.parent);
```
StephenWeatherford: nit: I thought there was a way to refresh a node's label...?

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -20,20 +21,22 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
 	}
 
 	public get label(): string {
-		if (this.isConnected) {
-			return this.databaseName + " (Connected)";
-		}
 		return this.databaseName;
 	}
 
+	public get description(): string {
```
StephenWeatherford: Might be worth considering to add a fullId property to nodes at some point...

EricJizbaMSFT: There is, but it's only at the 'node' level, not the 'treeItem' level

src/mongo/registerMongoCommands.ts
```diff
@@ -44,23 +48,24 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
...
+        }
     });
     actionHandler.registerCommand('cosmosDB.deleteMongoDB', async (node?: IAzureNode<MongoDatabaseTreeItem>) => {
         if (!node) {
             node = <IAzureNode<MongoDatabaseTreeItem>>await tree.showNodePicker(MongoDatabaseTreeItem.contextValue);
         }
         await node.deleteNode();
-        if (connectedDb && connectedDb.treeItem.id === node.treeItem.id) {
-            connectedDb = null;
+        if (ext.connectedMongoDB && ext.connectedMongoDB.id === node.id) {
```
StephenWeatherford: Also need this in detach.  Although not sure there's anything bad that happens if it's not done?

EricJizbaMSFT: Filed a separate issue: https://github.com/Microsoft/vscode-cosmosdb/issues/492

Don't feel like fixing in this PR and it was always a problem before my changes

src/mongo/tree/MongoDatabaseTreeItem.ts
StephenWeatherford: Ideally there'd be a way to refresh the entire tree's labels only.

EricJizbaMSFT: Fair enough. Doesn't exist yet

EricJizbaMSFT:
Thanks @StephenWeatherford for sticking with me through this PR haha 😊

PR #493, opened 2018-04-05
Fix attached accounts race condition

ejizba:
I saw issues where the connected mongo database didn't always persist for attached accounts. It's because there is a race condition related to loading the persisted accounts.

I suppose there was always a bit of a race condition here. For example, if the user expanded the 'Attached Accounts' node before we load the persisted accounts, they would not see their accounts.

### DIFF
--- a/src/tree/AttachedAccountsTreeItem.ts
+++ b/src/tree/AttachedAccountsTreeItem.ts
@@ -42,6 +42,19 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
         this._loadPersistedAccountsTask = this.loadPersistedAccounts();
     }
 
+    private async getAttachedAccounts(): Promise<IAzureTreeItem[]> {
+        if (!this._attachedAccounts) {
+            try {
+                this._attachedAccounts = await this._loadPersistedAccountsTask;
+            } catch {
+                this._attachedAccounts = [];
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
I would only be nervous if we'd already done a bunch of testing. We have to test anyways, so I don't see the harm in fixing this

nturinski:
I tried my hardest to break it.

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -32,12 +32,27 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
...
 
+    private _loadPersistedAccountsTask: Promise<IAzureTreeItem[]>;
+
     constructor(private readonly _globalState: vscode.Memento) {
         this._keytar = fetchNodeModule('keytar');
-        this.loadPersistedServers();
+        this._loadPersistedAccountsTask = this.loadPersistedAccounts();
+    }
+
+    private async getAttachedAccounts(): Promise<IAzureTreeItem[]> {
```
PrashanthCorp: Nit: Can we move this function to below loadMoreChildren? :P 
If I understand correctly, that's the only call site, and that makes it easier to read. 
Feel free to ignore.

EricJizbaMSFT: It's also called in 'attachAccount' and 'detach'

PR #496, opened 2018-04-05
Fix creating account in command palette

### LINKED ISSUE #495: Properties return undefined when trying to create an account through cmd palette
Repro steps:
1. ) Command palette > Cosmos DB: Delete Stored Procedure or Delete Account
2. ) Select subscription
3.) Select + Create New Account

Action: cosmosDB.deleteStoredProcedure
Error type: TypeError
Error Message: Cannot read property 'properties' of undefined

Version: 0.6.0
OS: win32

EricJizbaMSFT:
And hide stored procedure command since we're not shipping them yet

Fixes #495

### DIFF
--- a/package.json
+++ b/package.json
@@ -605,6 +605,10 @@
 				{
 					"command": "cosmosDB.loadMore",
 					"when": "never"
+				},
+				{
+					"command": "cosmosDB.deleteStoredProcedure",
+					"when": "never"
 				}
 			]
 		},
--- a/package.json
+++ b/package.json
@@ -762,7 +762,7 @@
 	},
 	"dependencies": {
 		"antlr4ts": "^0.4.0-alpha.4",
-		"azure-arm-cosmosdb": "^1.1.0-preview",
+		"azure-arm-cosmosdb": "1.1.0-preview",
 		"azure-arm-resource": "^3.0.0-preview",
 		"azure-graph": "^2.2.0",
 		"copy-paste": "^1.3.0",
--- a/src/tree/CosmosDBAccountProvider.ts
+++ b/src/tree/CosmosDBAccountProvider.ts
@@ -54,12 +54,13 @@ export class CosmosDBAccountProvider implements IChildProvider {
             ],
             wizardContext);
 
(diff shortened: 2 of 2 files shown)


src/tree/CosmosDBAccountProvider.ts
```diff
@@ -54,12 +54,13 @@ export class CosmosDBAccountProvider implements IChildProvider {
             ],
             wizardContext);
 
+        // https://github.com/Microsoft/vscode-azuretools/issues/120
```
StephenWeatherford: I don't understand exactly what this is fixing.

StephenWeatherford: Or at least how it relates to https://github.com/Microsoft/vscode-azuretools/issues/120

EricJizbaMSFT: This fixes "Cannot read property 'properties' of undefined" when actionContext is undefined. The reason its undefined is because its not being passed in to createChild as described in that issue

package.json
```diff
@@ -605,6 +605,10 @@
 				{
 					"command": "cosmosDB.loadMore",
 					"when": "never"
+				},
+				{
+					"command": "cosmosDB.deleteStoredProcedure",
+					"when": "never"
```
PrashanthCorp: Why have the command if we're not shipping stored procedures at this moment?

EricJizbaMSFT: I just did the smallest change I could to hide stored procedures code. I don't wanna completely remove it and change a lot of code

src/tree/CosmosDBAccountProvider.ts
StephenWeatherford: But it's being passed as a dummy, so we won't be getting the proper telemetry, right?

EricJizbaMSFT: Yes that is correct. That's why I filed the issue in the shared package, but I never felt like a few missing telemetry properties was high enough priority to work on (Remember - this only happens if people run the command from the command palette)

PR #498, opened 2018-04-05
Handle errors trying to launch shell, fixes #455 

### LINKED ISSUE #455: If mongo shell not found, the error shown is a timeout

StephenWeatherford:


src/mongo/shell.ts
```diff
@@ -98,6 +100,16 @@ export class Shell {
 				5000);
 			const disposable = this.onResult.event(result => {
 				disposable.dispose();
+
+				if (result.code) {
+					if (result.code === 'ENOENT') {
+						result.message = `Could not find Mongo shell. Make sure it is on your path or you have set the '${ext.settingsKeys.mongoShellPath}' VS Code setting to point to the Mongo shell executable file. Attempted command: "${this.execPath}"`;
+					}
+
+					e(result);
```
EricJizbaMSFT: Do you still need to call `clearTimeout(handler);` at some point?

src/mongo/shell.ts
```diff
@@ -98,6 +100,16 @@ export class Shell {
 				5000);
 			const disposable = this.onResult.event(result => {
 				disposable.dispose();
+
+				if (result.code) {
```
EricJizbaMSFT: What is `result`? Could it ever be undefined? And do we know `code` will always indicate an error or could it be a successful code?

PrashanthCorp: I ran a sample command. It looks like there is no code if the execution is hitherto successful. I might have to look at a more formal documentation. 

These were the results on each trigger of the event. 
- "switched to db trial00" 
- "{ "acknowledged" : true, "matchedCount" : 0, "modifiedCount" : 0 }" 

That said, I'd prefer a `if(result && result.code)` just in case.

src/mongo/shell.ts
StephenWeatherford: Fixed

PR #501, opened 2018-04-05
Use timeout on account name validation, fixes #494

### LINKED ISSUE #494: Can't create an account through the Cosmos Extension
Repro steps:
1.) Either right click a subscription or ctrl + shift+ p and create a new account
2.) Type a valid account name
3.) Hit enter
4.) Nothing happens-- try again
5.) All names appear invalid

Action: cosmosDB.createAccount
Error type: Error
Error Message: {"name":"Error","message":""}

Version: 0.6.0
OS: win32

StephenWeatherford:
Azure is currently giving Server Error 500 when checking for account name availability, if the name is available (after about a minute delay).

Also, VS Code doesn't handle this situation well.

So, after 2 seconds we'll assume the name is available.

### DIFF
--- a/src/utils/inputValidation.ts
+++ b/src/utils/inputValidation.ts
@@ -1,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { valueOnTimeout } from "./timeout";
+
+export const inputValidationTimeoutMs = 2000;
+
+export function validOnTimeout(inputValidation: () => Promise<string | undefined>) {
+    return valueOnTimeout(inputValidationTimeoutMs, undefined, inputValidation);
+}
--- a/src/utils/inputValidation.ts
+++ b/src/utils/inputValidation.ts
@@ -5,8 +5,17 @@
 
 import { valueOnTimeout } from "./timeout";
 
-export const inputValidationTimeoutMs = 2000;
+const inputValidationTimeoutMs = 2000;
 
-export function validOnTimeout(inputValidation: () => Promise<string | undefined>) {
-    return valueOnTimeout(inputValidationTimeoutMs, undefined, inputValidation);
+/**
+ * Intended to be used for VS Code validateInput to protect against long-running validations. If a time-out occurs or the action throws,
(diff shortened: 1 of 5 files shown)


StephenWeatherford:
Adding some unit tests, hold on...

StephenWeatherford:
Ready for review.

test/timeout.test.ts
```diff
@@ -0,0 +1,111 @@
...
+                    resolve();
+                });
+            })
+            assert.equal(executed, true);
+        });
+
+        test("timed out", async () => {
+            let executed = false;
+
+            try {
```
EricJizbaMSFT: Rather than doing this try/catch stuff, you can just use `assert.throws`, which validates that an exception is thrown when a block of code is executed

src/utils/inputValidation.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const inputValidationTimeoutMs = 2000;
+
+/**
+ * Intended to be used for VS Code validateInput to protect against long-running validations. If a time-out occurs or the action throws,
+ * returns undefined (indicating a valid input). Use for optional validations.
+ */
+export function validOnTimeoutOrException(inputValidation: () => Promise<string | undefined>, timeoutMs?: number): Promise<string | undefined> {
+    try {
+        timeoutMs = timeoutMs || inputValidationTimeoutMs;
```
EricJizbaMSFT: Is there any chance someone wants timeoutMs to be 0? If so, this code would ignore that and set the timeout to 2000.

StephenWeatherford: Doesn't seem like a reasonable scenario to me.

test/timeout.test.ts
```diff
@@ -0,0 +1,111 @@
...
+                return new Promise((resolve, reject) => {
+                    executed = true;
+                    resolve();
+                });
+            })
+
+            assert.equal(executed, true);
+        });
+
+        test("executes before time-out", async () => {
```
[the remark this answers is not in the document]
StephenWeatherford: Modified test as had originally intended.

src/utils/inputValidation.ts
EricJizbaMSFT: Okey dokey artichokey

test/timeout.test.ts
StephenWeatherford: Tried that, but it doesn't seem to handle async functions.

PR #503, opened 2018-04-06
findOne : no args

### LINKED ISSUE #502: null ref exception running findOne with no arguments
db.test.findOne()

MongoCollectionTreeItem.ts:
	private async findOne(args?: any[]): Promise<string> {
		if (args && args.length > 2) {
			throw new Error("Too many arguments")
		}
		let result;
		if (**args.length** === 1) {

PrashanthCorp:
Fixes #502

### DIFF
--- a/src/mongo/tree/MongoCollectionTreeItem.ts
+++ b/src/mongo/tree/MongoCollectionTreeItem.ts
@@ -169,7 +169,7 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 		} else if (args.length === 2) {
 			result = await this.collection.findOne(args[0], { fields: args[1] });
 		} else {
-			return Promise.reject(new Error("Too many arguments"));
+			return Promise.reject(new Error("Too many arguments passed to findOne."));
 		}
 		return this.stringify(result);
 	}
@@ -192,12 +192,12 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 	private insertMany(args: any[]): Thenable<string> {
 		// documents = args[0], collectionWriteOptions from args[1]
 		let insertManyOptions: CollectionInsertManyOptions = {};
-		const docsLink: string = "Please see mongo shell documentation. https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany";
+		const docsLink: string = "Please see mongo shell documentation. https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany.";
 		if (!args || args.length === 0) {
-			return Promise.reject(new Error("Too few arguments " + docsLink));
+			return Promise.reject(new Error("Too few arguments passed to insertMany. " + docsLink));
 		}
 		if (args.length > 2) {
-			return Promise.reject(new Error("Too many arguments " + docsLink));
+			return Promise.reject(new Error("Too many arguments passed to insertMany. " + docsLink));
 		} else if (args.length === 2) {
 			if (args[1] && args[1].ordered) {
 				insertManyOptions["ordered"] = args[1].ordered;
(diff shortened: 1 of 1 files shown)


src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -165,9 +165,9 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 			throw new Error("Too many arguments")
 		}
 		let result;
-		if (args.length === 1) {
+		if (args && args.length === 1) {
```
StephenWeatherford: Have you search for all args.length usage in all related files?  I see another possibility in insertMany

PrashanthCorp: We need at least one parameter for the [insertMany](http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html#insertMany) call - reflected in the _mandatory argument array_ to its call. In that case, havign args as undefined would be an error we want to propagate. 
NB: Given [the call](https://github.com/Microsoft/vscode-cosmosdb/blob/b66d52b14e86096809b1bc80ebd8eba3995a396d/src/mongo/tree/MongoCollectionTreeItem.ts#L112), args being undefined would throw an error at the call site itself.

EricJizbaMSFT: You shouldn't have to check 'args' for every 'else' case if you check it in the first 'if'. Something like this would be less bug-prone IMO:
```
if (!args || args.length === 0) {

} else if (args.length === 1) {

} else if (args.length === 2) {

} else {

}
```

EricJizbaMSFT: I don't quite follow what you're saying. Yes an error is expected, but you need to make sure it's the correct error. For example it should be "Too few arguments" instead of "Cannot access 'length' of undefined". Is that the case currently? I feel like the second error would be displayed, which is not what we want

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -161,16 +161,15 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
-		if (args.length === 1) {
+		if (!args || args.length === 0) {
+			result = await this.collection.findOne({});
+		} else if (args.length === 1) {
 			result = await this.collection.findOne(args[0]);
 		} else if (args.length === 2) {
 			result = await this.collection.findOne(args[0], { fields: args[1] });
 		} else {
-			result = await this.collection.findOne({});
+			return Promise.reject(new Error("Too many arguments"));
```
StephenWeatherford: Too many arguments to findOne

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -193,8 +192,12 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 	private insertMany(args: any[]): Thenable<string> {
 		// documents = args[0], collectionWriteOptions from args[1]
 		let insertManyOptions: CollectionInsertManyOptions = {};
+		const docsLink: string = "Please see mongo shell documentation. https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany";
+		if (!args || args.length === 0) {
+			return Promise.reject(new Error("Too few arguments " + docsLink));
```
StephenWeatherford: Missing period.  add name of method.

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -193,8 +192,12 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
+		}
 		if (args.length > 2) {
-			throw new Error("Too many arguments. Please see mongo shell documentation. https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany");
+			return Promise.reject(new Error("Too many arguments " + docsLink));
```
StephenWeatherford: ditto

Issue #408, opened 2018-03-15
Build errors from 'tsc -p ./' aren't displayed in CI


EricJizbaMSFT:
We had a build failure where `gulp package` displayed this:
```
[13:11:30] Using gulpfile ~/Repos/vscode-cosmosdb/gulpfile.js
[13:11:30] Starting 'package'...
Executing prepublish script 'npm run vscode:prepublish'...
[13:11:36] 'package' errored after 5.54 s
[13:11:36] Error: Command failed: npm run vscode:prepublish
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! vscode-cosmosdb@0.4.0 vscode:prepublish: `tsc -p ./`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the vscode-cosmosdb@0.4.0 vscode:prepublish script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/ericjizba/.npm/_logs/2018-03-15T20_11_36_449Z-debug.log

    at formatError (/usr/local/lib/node_modules/gulp/bin/gulp.js:169:10)
    at Gulp.<anonymous> (/usr/local/lib/node_modules/gulp/bin/gulp.js:195:15)
    at emitOne (events.js:115:13)
    at Gulp.emit (events.js:210:7)
    at Gulp.Orchestrator._emitTaskDone (/Users/ericjizba/Repos/vscode-cosmosdb/node_modules/orchestrator/index.js:264:8)
    at /Users/ericjizba/Repos/vscode-cosmosdb/node_modules/orchestrator/index.js:275:23
    at finish (/Users/ericjizba/Repos/vscode-cosmosdb/node_modules/orchestrator/lib/runTask.js:21:8)
    at /Users/ericjizba/Repos/vscode-cosmosdb/node_modules/orchestrator/lib/runTask.js:45:4
    at <anonymous>
```

But if I ran `tsc -p ./` locally, I saw this:
```
src/graph/GraphViewServer.ts(44,20): error TS2503: Cannot find namespace 'SocketIO'.
src/graph/GraphViewServerSocket.ts(10,32): error TS2503: Cannot find namespace 'SocketIO'.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! vscode-cosmosdb@0.4.0 vscode:prepublish: `tsc -p ./`
[...]

PR #515, opened 2018-04-10
Add Mongo code lens provider, Run, Run All, show connected DB and show commands in output

StephenWeatherford:
1) Show executing command in output
2) Run All
3) Code Lens for "Connected to <db>", "Run All" and "Run"
4) Execute Command -> Execute Mongo Command (editor context menu)
5) Commands.ts -> MongoCommands.ts

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,10 @@
 	},
 	"main": "./out/src/extension",
 	"activationEvents": [
+		"onLanguage:mongo",
 		"onView:cosmosDBExplorer",
 		"onCommand:cosmosDB.executeMongoCommand",
+		"onCommand:cosmosDB.executeAllMongoCommands",
 		"onCommand:cosmosDB.attachDatabaseAccount",
 		"onCommand:cosmosDB.attachEmulator",
 		"onCommand:cosmosDB.createAccount",
@@ -113,7 +115,12 @@
 			{
 				"category": "MongoDB",
 				"command": "cosmosDB.executeMongoCommand",
-				"title": "Execute Command"
+				"title": "Execute Mongo Command"
+			},
+			{
+				"category": "MongoDB",
+				"command": "cosmosDB.executeAllMongoCommands",
+				"title": "Execute All Mongo Commands"
 			},
 			{
 				"category": "MongoDB",
@@ -305,6 +312,10 @@
 			"editor/context": [
 				{
 					"command": "cosmosDB.executeMongoCommand",
+					"when": "resourceLangId==mongo"
+				},
+				{
+					"command": "cosmosDB.executeAllMongoCommands",
 					"when": "resourceLangId==mongo"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -115,12 +115,12 @@
 			{
 				"category": "MongoDB",
 				"command": "cosmosDB.executeMongoCommand",
-				"title": "Execute Mongo Command"
+				"title": "Execute MongoDB Command"
 			},
 			{
 				"category": "MongoDB",
 				"command": "cosmosDB.executeAllMongoCommands",
-				"title": "Execute All Mongo Commands"
+				"title": "Execute All MongoDB Commands"
 			},
 			{
 				"category": "MongoDB",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "vscode-cosmosdb",
-	"version": "0.6.0",
+	"version": "0.6.1-alpha",
 	"aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
 	"publisher": "ms-azuretools",
 	"displayName": "Azure Cosmos DB",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -115,7 +115,7 @@ export function activate(context: vscode.ExtensionContext) {
 	actionHandler.registerCommand('cosmosDB.update', (filePath: vscode.Uri) => editorManager.updateMatchingNode(filePath, tree));
 	actionHandler.registerCommand('cosmosDB.loadMore', (node?: IAzureNode) => tree.loadMore(node));
 	actionHandler.registerEvent('cosmosDB.CosmosEditorManager.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async function
-		(this: IActionContext, doc: vscode.TextDocument): Promise<void> { await editorManager.onDidSaveTextDocument(this, context.globalState, doc, tree) });
+		(this: IActionContext, doc: vscode.TextDocument): Promise<void> { await editorManager.onDidSaveTextDocument(this, doc, tree) });
 }
 
 async function getAttachedNode(tree: AzureTreeDataProvider): Promise<IAzureParentNode<AttachedAccountsTreeItem>> {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,6 +27,7 @@ import { DocDBAccountTreeItemBase } from './docdb/tree/DocDBAccountTreeItemBase'
 import { GraphAccountTreeItem } from './graph/tree/GraphAccountTreeItem';
 import { DocDBAccountTreeItem } from './docdb/tree/DocDBAccountTreeItem';
 import { TableAccountTreeItem } from './table/tree/TableAccountTreeItem';
+import { MongoCodeLensProvider } from './mongo/services/MongoCodeLensProvider';
 
 export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(new Reporter(context));
@@ -37,6 +38,9 @@ export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(tree);
 	context.subscriptions.push(vscode.window.registerTreeDataProvider('cosmosDBExplorer', tree));
 
+	const codeLensProvider = new MongoCodeLensProvider();
+	context.subscriptions.push(vscode.languages.registerCodeLensProvider('mongo', codeLensProvider));
+
 	const editorManager: CosmosEditorManager = new CosmosEditorManager(context.globalState);
 
 	context.subscriptions.push(vscodeUtil.getOutputChannel());
@@ -45,7 +49,7 @@ export function activate(context: vscode.ExtensionContext) {
 
 	registerDocDBCommands(actionHandler, tree);
 	registerGraphCommands(context, actionHandler, tree);
-	registerMongoCommands(context, actionHandler, tree, editorManager);
+	registerMongoCommands(context, actionHandler, tree, editorManager, codeLensProvider);
 
 	// Common commands
 	const accountContextValues: string[] = [GraphAccountTreeItem.contextValue, DocDBAccountTreeItem.contextValue, TableAccountTreeItem.contextValue, MongoAccountTreeItem.contextValue];
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,6 @@ import { DocDBAccountTreeItemBase } from './docdb/tree/DocDBAccountTreeItemBase'
 import { GraphAccountTreeItem } from './graph/tree/GraphAccountTreeItem';
 import { DocDBAccountTreeItem } from './docdb/tree/DocDBAccountTreeItem';
 import { TableAccountTreeItem } from './table/tree/TableAccountTreeItem';
-import { MongoCodeLensProvider } from './mongo/services/MongoCodeLensProvider';
 
 export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(new Reporter(context));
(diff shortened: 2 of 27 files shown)


StephenWeatherford:
@fiveisprime 
Question: Should it be "Execute" and "Execute All" to match the current command names?

src/mongo/MongoCommands.ts
```diff
@@ -18,16 +18,61 @@ import { MongoFindOneResultEditor } from './editors/MongoFindOneResultEditor';
...
+		const commands = MongoCommands.getAllCommandsFromActiveEditor();
 		const activeEditor = vscode.window.activeTextEditor;
-		if (activeEditor.document.languageId !== 'mongo') {
-			vscode.window.showInformationMessage("This command can only be run inside of a MongoDB scrapbook (*.mongo)");
-			return;
-		}
 		const selection = activeEditor.selection;
-		const command = MongoCommands.getCommand(activeEditor.document.getText(), selection.start);
+		const command = MongoCommands.findCommandAtPosition(commands, selection.start);
+		return this.executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
```
PrashanthCorp: 1. The order of these arguments is slightly different from the executeCommands function. Would it make sense to make these consistent? 
2. Why is this returned when the other use case (inside executeCommands) awaited? 

As an aside, would it make sense to just have one function, `executeCommands`, with this specific case having a function call - `await MongoCommands.executeCommands([command], vscode.window.activeTextEditor, database, extensionPath, editorManager, tree, context)` where we pass the command in an array? I think there might be lesser code, but it could seem less readable.

StephenWeatherford:
The run button always runs the line of code beneath it.  That's how all similar features I've been work, I don't think that will be confusing.

src/extension.ts
```diff
@@ -37,6 +38,9 @@ export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(tree);
 	context.subscriptions.push(vscode.window.registerTreeDataProvider('cosmosDBExplorer', tree));
 
+	const codeLensProvider = new MongoCodeLensProvider();
+	context.subscriptions.push(vscode.languages.registerCodeLensProvider('mongo', codeLensProvider));
+
```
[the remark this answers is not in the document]
StephenWeatherford: done

src/mongo/MongoCommands.ts
StephenWeatherford: I prefer the readability

src/mongo/MongoCommands.ts
```diff
@@ -18,16 +18,61 @@ import { MongoFindOneResultEditor } from './editors/MongoFindOneResultEditor';
...
+
+	public static async executeCommands(activeEditor: vscode.TextEditor, database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext, commands: MongoCommand[]): Promise<void> {
+		for (let command of commands) {
+			await this.executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+		}
+	}
+
+	public static async executeCommand(activeEditor: vscode.TextEditor, database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext, command: MongoCommand): Promise<void> {
 		if (command) {
+			output.show();
```
EricJizbaMSFT: Not a fan of this output - did you add it for debugging purposes only? A few reasons:
1. The user should know what command we're running and it feels like a bug if we're not running the command they expect
1. There's no way to link this output to its corresponding result.json file
1. We're already popping up the result.json file and I think popping up the output channel as well is too much

I guess I'm okay with this as long as you remove `output.show`.

Another idea is to insert the command to the top of the result.json file as a comment (although comments aren't strictly supported in json files so idk)

src/mongo/registerMongoCommands.ts
```diff
@@ -94,29 +98,49 @@ export function registerMongoCommands(context: vscode.ExtensionContext, actionHa
...
-            const persistedNode: IAzureNode | undefined = await tree.findNode(persistedNodeId);
-            if (persistedNode) {
-                await languageClient.client.onReady();
-                await vscode.commands.executeCommand('cosmosDB.connectMongoDB', persistedNode);
-            }
+            persistedNode = await tree.findNode(persistedNodeId);
         }
     });
+
+    if (persistedNode) {
```
EricJizbaMSFT: Why did you move this code outside of `callWithTelemetryAndErrorHandling`? It was put in there since we await 'loadPersistedMongoDBTask' multiple times and we don't want the error being displayed multiple times

src/mongo/services/MongoCodeLensProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+				command: isInitialized && 'cosmosDB.connectMongoDB'
+			},
+			range: new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 0))
+		});
+
+		if (isConnected) {
+			// Run all
+			lenses.push(<vscode.CodeLens>{
+				command: {
+					title: "Run All",
```
EricJizbaMSFT: Can we pick either 'Execute' or 'Run'? I don't have a preference, but it would be nice to be consistent

package.json
```diff
@@ -113,7 +115,12 @@
 			{
 				"category": "MongoDB",
 				"command": "cosmosDB.executeMongoCommand",
-				"title": "Execute Command"
+				"title": "Execute Mongo Command"
```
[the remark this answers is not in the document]
StephenWeatherford: The Mongo/MongoDB is inconsistent.  For instance, it's "New Mongo Scrapbook", and the name of the "language" is currently "mongo", and the extension is "*.mongo".  I would have chosen MongoDB, but not sure what it should be now.

Added Mongo to the command because it's not very clear that the context menu in the scrapbook refers to this.  There can be lots of other commands in there, too, and simply "Execute Command" is rather vague (I passed it up the first time I saw it).

image

src/mongo/MongoCommands.ts
```diff
@@ -18,16 +18,61 @@ import { MongoFindOneResultEditor } from './editors/MongoFindOneResultEditor';
...
+		return this.executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+	}
+
+	public static getAllCommandsFromActiveEditor(): MongoCommand[] {
+		const activeEditor = vscode.window.activeTextEditor;
+		if (activeEditor) {
+			const commands = MongoCommands.getAllCommandsFromTextDocument(activeEditor.document);
+			return commands;
+		} else {
+			throw new Error(notInScrapbookMessage);
```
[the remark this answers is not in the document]
StephenWeatherford: This is currently there (not my code) because the commands are exposed in the command palette but are always enabled. Excluding them sounds good, except that would make it inaccessible unless I add a keyboard shortcut.

src/mongo/MongoCommands.ts
StephenWeatherford: Personally I much prefer to see what it's doing, especially for Execute All, and if it hits an error.  Fine with not showing output, though.  We might need to refine this behavior in some way. We can revisit after people have played with it more.

src/mongo/registerMongoCommands.ts
StephenWeatherford: Doesn't matter how many times it's awaited, it will only run once.  

CodeLens will show "Initializing" until told setConnectedDatabase called first time, so have to make sure that happens. However, current version doesn't handle errors connecting to DB, I think this next one will be better...

src/mongo/MongoCommands.ts
```diff
@@ -18,16 +18,61 @@ import { MongoFindOneResultEditor } from './editors/MongoFindOneResultEditor';
 import { MongoCommand } from './MongoCommand';
 import { MongoDatabaseTreeItem } from './tree/MongoDatabaseTreeItem';
 
+const output = vscodeUtil.getOutputChannel();
+const notInScrapbookMessage = "This command can only be run inside of a MongoDB scrapbook (*.mongo)";
+
 export class MongoCommands {
+
+	public static async executeAllCommandsFromActiveEditor(database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext): Promise<void> {
```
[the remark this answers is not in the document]
EricJizbaMSFT: FWIW I agree it's generally better to keep functional changes separate from refactoring, but I make an exception when people clean up old code that they're actively changing anyways. It feels like a sustainable way to clean up code as we go rather than filing a bunch of engineering issues. No strong opinion, though - separate PR is fine.

src/mongo/MongoCommands.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I thought it was accessible through code lens?

src/mongo/MongoCommands.ts
EricJizbaMSFT: Good point - but it feels like we should include the command in the error message anyways

src/mongo/MongoCommands.ts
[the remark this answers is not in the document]
StephenWeatherford: "Accessible" accessible - ie through keyboard. Maybe code lens does have this, will have to look it up.  If there is, I have no strong opinion.

fiveisprime:
The code lens text in a new file ends up out of view for me and I have to scroll down to see it. After I scroll it into view, typing shifts everything back up and out of view.

StephenWeatherford:
@fiveisprime I was just noticing that, are you using insiders?

src/mongo/MongoCommands.ts
StephenWeatherford: I still think we need to play with it and make holistic changes in the flow rather than dealing with it in this PR.  The error handling is not new behavior, although execute all is new.

src/mongo/MongoCommands.ts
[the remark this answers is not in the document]
StephenWeatherford: Can't seem to find a shortcut for this in vscode.

fiveisprime:
Yeah, on Insiders.

StephenWeatherford:
Don't think I saw it at work with current, will try when I get in office.

src/mongo/MongoCommands.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Oh that "Accessible". Well I definitely think you should exclude `executeMongoCommand` from the command palette. You lose context of the cursor when you open the command palette so it always runs the first command. Plus it already has a keyboard shortcut.

I guess it might make sense to have `executeAllMongoCommands` in the command palette, but a keyboard shortcut feels more natural to me. No strong opinion on this one. I will point out that whether or not you add a default keyboard shortcut, the user can add a custom keyboard shortcut for literally any command we define.

src/mongo/MongoCommands.ts
EricJizbaMSFT: That's fine - you already removed `output.show()` which was the "new" behavior that I cared about the most

src/mongo/MongoCommands.ts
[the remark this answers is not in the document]
StephenWeatherford: @fiveisprime any suggestions on a keyboard shortcut for execute all?  
Currently CTRL+SHIFT+” is execute.  I would choose CTRL+SHIFT+ENTER but that’s Insert Line Above”.

src/mongo/registerMongoCommands.ts
EricJizbaMSFT: If you await a promise multiple times, you're correct that it only runs once. However, if the result of awaiting the promise is an error, you will get that error thrown every single time you await it. For example you can test this out by putting `throw new Error('Hi Stephen!');` right before `codeLensProvider.setConnectedDatabase(undefined);`. The error will pop up every time you execute a mongo command (not just the first time).

Instead, I suggest moving everything inside `await callWithTelemetryAndErrorHandling(` and using a 'try/finally' to make sure CodeLens always gets initialized

StephenWeatherford: setConnectedDatabase can't fail, but yeah I agree it's better inside the telemetry.

StephenWeatherford:
@fiveisprime https://github.com/Microsoft/vscode/issues/47703

fiveisprime:
@StephenWeatherford 

1. I like `Execute` and `Execute All`
1. How about moving to the left instead of right for Execute All shortcut: `CTRL+SHIFT+:`
1. Thanks for creating that issue

src/mongo/registerMongoCommands.ts
EricJizbaMSFT: "can't fail" <- famous last words 😂

StephenWeatherford: "As currently implemented" (tm)

EricJizbaMSFT:
@StephenWeatherford if you're trying to fix the build, this is the error:
```
src/mongo/services/MongoCodeLensProvider.ts(22,58): error TS6133: 'token' is declared but its value is never read.
```

We have this known issue: https://github.com/Microsoft/vscode-cosmosdb/issues/408 I guess one solution is to add `npm run build` before `gulp package` in the travis config. Technically it's redundant since the package runs the build, but that way we actually get to see the error messages.

PR #522, opened 2018-04-11
Start: Implement strictNullCheck in a few files

PrashanthCorp:


src/utils/timeout.ts
```diff
@@ -26,7 +26,7 @@ export async function valueOnTimeout<T>(timeoutMs: number, timeoutValue: T, acti
  */
 export async function rejectOnTimeout<T>(timeoutMs: number, action: () => Promise<T> | T) {
     return await new Promise<T>(async (resolve, reject) => {
-        let timer: NodeJS.Timer = setTimeout(
+        let timer: NodeJS.Timer | null = setTimeout(
```
StephenWeatherford: Please change null to undefined, which is what vscode prefers for some reason.

PR #525, opened 2018-04-12
Remove duplicate code in editor node labels

PrashanthCorp:
Noticed that the previous implementation was guilty with the strictNullCheck flag. 
Remove almost-duplicate code in those files to reduce number of offending lines. In the process, this function is strictNullCheck compliant. :)

### DIFF
--- a/src/utils/vscodeUtils.ts
+++ b/src/utils/vscodeUtils.ts
@@ -6,6 +6,10 @@
 import * as path from 'path';
 import * as fse from 'fs-extra';
 import * as vscode from 'vscode';
+import { IAzureNode } from 'vscode-azureextensionui';
+import { MongoAccountTreeItem } from '../mongo/tree/MongoAccountTreeItem';
+import { DocDBAccountTreeItem } from '../docdb/tree/DocDBAccountTreeItem';
+import { DocDBAccountTreeItemBase } from '../docdb/tree/DocDBAccountTreeItemBase';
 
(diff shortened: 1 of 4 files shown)


src/utils/vscodeUtils.ts
```diff
@@ -78,3 +81,15 @@ export function fetchNodeModule(moduleName: string) {
     } catch (err) { }
     return null;
 }
+
+export function getLabel(node: IAzureNode): string {
+    let labels = [node.treeItem.label];
+    while (node.parent) {
+        node = node.parent;
+        labels.unshift(node.treeItem.label);
+        if ((node.treeItem instanceof MongoAccountTreeItem) || (node.treeItem instanceof DocDBAccountTreeItemBase)) {
```
StephenWeatherford: Extract into a isAccountTreeItem() method

StephenWeatherford: Ideally we'd have something more generic but not sure it's worth it for this small case.

src/utils/vscodeUtils.ts
```diff
@@ -78,3 +81,15 @@ export function fetchNodeModule(moduleName: string) {
     } catch (err) { }
     return null;
 }
+
+export function getLabel(node: IAzureNode): string {
```
StephenWeatherford: More specific name needed.  Lots of things have labels.

PR #526, opened 2018-04-12
Restricting language services to local files

lostintangent:
In preparation for [Visual Studio Live Share](http://aka.ms/vsls) adding support for "guests" to receive remote language services for MongoDB, this PR simply updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files. This way, when someone has this extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with)

### DIFF
--- a/src/mongo/languageClient.ts
+++ b/src/mongo/languageClient.ts
@@ -2,50 +2,75 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import * as path from 'path';
-import * as nls from 'vscode-nls';
-import { ExtensionContext } from 'vscode';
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
2 questions:
1) Can you not do something in vscode so that you get the correct behavior automatically rather than relying on the extensions to do it?
2) Have you updated the extension develop documentation and samples?

None of this will be obvious from looking at the code.  THanks.

lostintangent:
@StephenWeatherford VS Code will be updating their extensibility docs to discuss custom file system providers and the necessary considerations for extension authors.

At some point, they may change the default behavior be to limit language services to the `file` and `untitled` schemes, but that's not a committed change, with a clear ETA. In the meantime, this PR seems to also make the extension's behavior more explicit.

PR #529, opened 2018-04-12
More PR comments from CodeLens

StephenWeatherford:
Add shortcut for Execute All (SHIFT+CTRL+;)
MongoCommands.ts -> MongoScrapbook.ts, and remove the static class


package.json
```diff
@@ -620,6 +620,14 @@
 				{
 					"command": "cosmosDB.deleteStoredProcedure",
 					"when": "never"
+				},
+				{
+					"command": "cosmosDB.executeAllMongoCommands",
+					"when": "never"
+				},
```
PrashanthCorp: I find it sensible to run all the commands using the command palette. I prefer the redundancy against codelens momentarily failing.
Given that (currently) errors in parsing can prevent codelens from appearing, and that codelens is not very well tested, I'd vote for having this function included. :)

src/mongo/MongoScrapbook.ts
```diff
@@ -0,0 +1,175 @@
...
+
+const output = vscodeUtil.getOutputChannel();
+const notInScrapbookMessage = "You must have a MongoDB scrapbook (*.mongo) open to run a MongoDB command.";
+
+export async function executeAllCommandsFromActiveEditor(database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext): Promise<void> {
+	output.appendLine("Running all commands in scrapbook...")
+	let commands = getAllCommandsFromActiveEditor();
+	await executeCommands(vscode.window.activeTextEditor, database, extensionPath, editorManager, tree, context, commands);
+}
+
```
PrashanthCorp: I see that all the commands have been moved out of the class - wanted to know what the rationale is. :)

StephenWeatherford: The rational is that I don't like static classes. They're a pain to use (have to repeatedly type the class name), and also not necessary in TypeScript.  Using plain functions is simpler, and you can always use a namespace if you want to.  Eric has warmed up to this philosophy.

package.json
StephenWeatherford: We don't need to enable them based on the fact that CodeLens might or might not fail. If that happens, it should be investigated.

You can still do this via the keyboard shortcut.

Personally I don't have a strong opinion on this, though.

EricJizbaMSFT: I strongly feel that "cosmosDB.executeMongoCommand" should _not_ be in the command palette because you lose context of your cursor when you open the command palette and it always runs the first command.

I don't have a strong opinion on "cosmosDB.executeAllMongoCommands" since the cursor doesn't matter in that case.

StephenWeatherford: cosmosDB.executeMongoCommand from the palette does not lose the cursor context - it runs the command the cursor is on.

I've added both commands back but set them to only be visible when editorLangId=='mongo'.

I'll check it in this way but feel free to come by and discuss further.

src/mongo/MongoScrapbook.ts
```diff
@@ -0,0 +1,175 @@
...
+	return await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+}
+
+export async function executeCommandFromText(database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext, commandText: string): Promise<void> {
+	const activeEditor = vscode.window.activeTextEditor;
+	const command = getCommandFromText(commandText, new vscode.Position(0, 0));
+	return await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+}
+
+export function getAllCommandsFromActiveEditor(): MongoCommand[] {
```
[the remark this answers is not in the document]
StephenWeatherford: fixed

src/mongo/MongoScrapbook.ts
```diff
@@ -0,0 +1,175 @@
...
+import { MongoFindResultEditor } from './editors/MongoFindResultEditor';
+import { MongoFindOneResultEditor } from './editors/MongoFindOneResultEditor';
+import { MongoCommand } from './MongoCommand';
+import { MongoDatabaseTreeItem } from './tree/MongoDatabaseTreeItem';
+
+const output = vscodeUtil.getOutputChannel();
+const notInScrapbookMessage = "You must have a MongoDB scrapbook (*.mongo) open to run a MongoDB command.";
+
+export async function executeAllCommandsFromActiveEditor(database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext): Promise<void> {
+	output.appendLine("Running all commands in scrapbook...")
```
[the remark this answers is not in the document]
StephenWeatherford: Separate PR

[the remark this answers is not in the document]
StephenWeatherford: Looked at those with tabs, don't think this is worth messing with. The default VSCode settings at least keep from overwriting willy nilly.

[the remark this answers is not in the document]
EricJizbaMSFT: I don't expect you to go back and change old files from tabs to spaces. However, I do prefer that new files follow our established pattern/preference, which is spaces. Since you already merged this PR it's no longer a "new file" and I agree it's not worth a separate PR.

PR #536, opened 2018-04-17
Implement few tslint rules

PrashanthCorp:
- no unused expressions 
- max-classes-per-file
- no-empty (no empty blocks of code)

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -122,7 +122,7 @@
         // TODO: "interface-name": true,
         "jsdoc-format": true,
         "max-classes-per-file": [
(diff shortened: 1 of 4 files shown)


tslint.json
```diff
@@ -122,7 +122,7 @@
         // TODO: "interface-name": true,
         "jsdoc-format": true,
         "max-classes-per-file": [
-            false, // TODO: true,
+            true, // TODO: true,
```
EricJizbaMSFT: Same comment as before. Remove the "// TODO: true" since you already set it to true

PR #540, opened 2018-04-19
Find should open in next column, fixes #514

### LINKED ISSUE #514: find command in scrapbook doesn't open in right-hand column like other commands do
Run these in a scrapbook:

db.c1.insert({ "key": "value2" })
db.c1.find({"key":"value"});

The first one opens in the right-hand column, but the second doesn't not.  Expecting both to open in right-hand column.

image

StephenWeatherford:


src/CosmosEditorManager.ts
```diff
@@ -37,7 +44,17 @@ export class CosmosEditorManager {
...
 
-    public async showDocument(editor: ICosmosEditor, fileName: string): Promise<void> {
+    public async showDocument(editor: ICosmosEditor, fileName: string, options: ShowEditorDocumentOptions): Promise<void> {
+        let column: vscode.ViewColumn = vscode.ViewColumn.Active;
+        let preserveFocus: boolean = false;
+        if (options && options.showInNextColumn) {
+            preserveFocus = true;
+            const activeEditor = vscode.window.activeTextEditor;
+            if (activeEditor && activeEditor.viewColumn >= vscode.ViewColumn.One) {
+                column = activeEditor.viewColumn < ViewColumn.Three ? activeEditor.viewColumn + 1 : ViewColumn.One;
```
PrashanthCorp: I see an application for using the modulo operator here :  `(activeEditor.viewColumn%3) + 1`

Issue #520, opened 2018-04-11
Replace "DocumentDB" with "SQL"


fiveisprime:
I have confirmation from the Cosmos DB team that this should absolutely be "SQL."

See also: https://docs.microsoft.com/en-us/azure/cosmos-db/faq

StephenWeatherford:
Should we do something like "SQL (DocumentDB)"?

fiveisprime:
Let's just go with "SQL" and we can link to the FAQ in the changelog for clarification. This will be consistent with the portal.

StephenWeatherford:
Dupe of #415

PR #549, opened 2018-04-19
Add (DocumentDB) to SQL description

StephenWeatherford:
Agreed to in #520 but got left out.

### DIFF
--- a/src/experiences.ts
+++ b/src/experiences.ts
@@ -48,10 +48,10 @@ export function getExperienceQuickPick(api: API): IAzureQuickPickItem<Experience
 }
 
 const experiencesArray: Experience[] = [
-    { api: API.DocumentDB, longName: "SQL", shortName: "SQL", kind: DBAccountKind.GlobalDocumentDB },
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
image

src/experiences.ts
```diff
@@ -48,10 +48,10 @@ export function getExperienceQuickPick(api: API): IAzureQuickPickItem<Experience
 }
 
 const experiencesArray: Experience[] = [
-    { api: API.DocumentDB, longName: "SQL", shortName: "SQL", kind: DBAccountKind.GlobalDocumentDB },
+    { api: API.DocumentDB, longName: "SQL", description: "(DocumentDB)", shortName: "SQL", kind: DBAccountKind.GlobalDocumentDB },
```
PrashanthCorp: 'SQL (DocumentDB)' in one place and 'DocumentDB (SQL)' everywhere else seems a bit confusing to the user. 
For the sake of consistency, I think the order here should be swapped, considering what's done in the other PR.

EricJizbaMSFT: I agree with @PrashanthCorp that we should be consistent, but I prefer 'SQL (DocumentDB)' so this PR looks fine to me

StephenWeatherford: I'll change the other PR.

PR #560, opened 2018-04-23
Improve warning messages

EricJizbaMSFT:
1. Make confirmation messages modal (https://github.com/Microsoft/vscode-azuretools/issues/151)
1. Use 'Delete' instead of 'Yes' when asking to delete something
1. Leverage common DialogResponses class

### DIFF
--- a/src/CosmosEditorManager.ts
+++ b/src/CosmosEditorManager.ts
@@ -144,7 +144,7 @@ export class CosmosEditorManager {
             const showSaveWarning: boolean | undefined = vscode.workspace.getConfiguration().get(this.showSavePromptKey);
             if (showSaveWarning !== false) {
                 const message: string = `Saving 'cosmos-editor.json' will update the entity "${editor.label}" to the Cloud.`;
-                const result: MessageItem | undefined = await vscode.window.showWarningMessage(message, { modal: true }, DialogResponses.upload, DialogResponses.alwaysUpload, DialogResponses.cancel);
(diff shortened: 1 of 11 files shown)


src/CosmosEditorManager.ts
```diff
@@ -145,11 +144,11 @@ export class CosmosEditorManager {
             const showSaveWarning: boolean | undefined = vscode.workspace.getConfiguration().get(this.showSavePromptKey);
             if (showSaveWarning !== false) {
                 const message: string = `Saving 'cosmos-editor.json' will update the entity "${editor.label}" to the Cloud.`;
-                const result: MessageItem | undefined = await vscode.window.showWarningMessage(message, DialogBoxResponses.upload, DialogBoxResponses.uploadDontWarn, DialogBoxResponses.Cancel);
+                const result: MessageItem | undefined = await vscode.window.showWarningMessage(message, { modal: true }, DialogResponses.upload, DialogResponses.alwaysUpload, DialogResponses.cancel);
```
EricJizbaMSFT: Do you guys think this one should be modal? It was the only one I was somewhat unsure about since we're preventing the user from continuing their work and I don't know how often they save

StephenWeatherford: I'm inclined to leave it modeless

EricJizbaMSFT: Discussed with @fiveisprime and he doesn't have a strong opinion either - we're fine with leaving it modal-less

nturinski:
Could you localize() that string please?

PR #569, opened 2018-04-28
Document labels: use fields other than _id wherever possible

### LINKED ISSUE #145: Document ID does not refresh on Explorer
If you rename an id, it does not refresh on the explorer unless you manually refresh it.  It should probably refresh itself to update the label on the explorer if the id is changed.

### LINKED ISSUE #381: Mongo ids displayed in the tree give no information
Today the user sees a bunch of ids like this:
screen shot 2018-03-05 at 8 59 16 am

But those aren't helpful at all. I see a few ways to fix (which are not mutually exclusive):
1. Create a setting so that users can specify the property to display in the tree
1. When the user creates a new document, use the value they enter for the id as the "_id" property instead of "id".  If I create a document with id specified as "exampleid", this is what I get:
```
{
  "_id": "5a9d7668dcb45709671aa200",
  "id": "exampleId"
}
```

We could be using "_id" - I know its possible since I can run this command successfully:
```
db.scores.insertOne{
    "_id": "exampleId"
})
```

PrashanthCorp:
Fixes #381 
Fixes #145

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -34,3 +34,5 @@ export const defaultStoredProcedure =
 };` ;
 
 export const emptyPartitionKeyValue = {};
+
+export const documentDefaultFields = ["name", "Name", "ID", "UUID", "id", "_id"];
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -35,4 +35,4 @@ export const defaultStoredProcedure =
 
 export const emptyPartitionKeyValue = {};
 
-export const documentLabelFields = ["name", "Name", "ID", "UUID", "Id", "id", "_id", "uuid"];
+export const documentLabelFields = ["name", "Name", "NAME", "ID", "UUID", "Id", "id", "_id", "uuid"];
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -35,4 +35,4 @@ export const defaultStoredProcedure =
 
 export const emptyPartitionKeyValue = {};
 
-export const documentLabelFields = ["name", "Name", "ID", "UUID", "id", "_id"];
+export const documentLabelFields = ["name", "Name", "ID", "UUID", "Id", "id", "_id", "uuid"];
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -35,4 +35,4 @@ export const defaultStoredProcedure =
 
 export const emptyPartitionKeyValue = {};
 
-export const documentDefaultFields = ["name", "Name", "ID", "UUID", "id", "_id"];
+export const documentLabelFields = ["name", "Name", "ID", "UUID", "id", "_id"];
--- a/src/docdb/editors/DocDBDocumentNodeEditor.ts
+++ b/src/docdb/editors/DocDBDocumentNodeEditor.ts
@@ -24,7 +24,9 @@ export class DocDBDocumentNodeEditor implements ICosmosEditor<RetrievedDocument>
     }
 
     public async update(document: RetrievedDocument): Promise<RetrievedDocument> {
-        return await this._documentNode.treeItem.update(document);
+        const updatedDoc = await this._documentNode.treeItem.update(document);
+        this._documentNode.refresh();
+        return updatedDoc;
     }
 
     public get id(): string {
--- a/src/docdb/editors/DocDBDocumentNodeEditor.ts
+++ b/src/docdb/editors/DocDBDocumentNodeEditor.ts
@@ -25,7 +25,7 @@ export class DocDBDocumentNodeEditor implements ICosmosEditor<RetrievedDocument>
 
     public async update(document: RetrievedDocument): Promise<RetrievedDocument> {
         const updatedDoc = await this._documentNode.treeItem.update(document);
-        this._documentNode.refresh();
+        await this._documentNode.refresh();
         return updatedDoc;
     }
 
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -9,7 +9,7 @@ import { IAzureNode, IAzureTreeItem, UserCancelledError, DialogResponses } from 
 import { RetrievedDocument, DocumentClient } from 'documentdb';
 import { DocDBCollectionTreeItem } from './DocDBCollectionTreeItem';
 import { emptyPartitionKeyValue } from '../../constants';
-import { documentDefaultFields } from '../../constants';
+import { getDocumentTreeItemLabel } from '../../utils/vscodeUtils';
 
 /**
  * Represents a Cosmos DB DocumentDB (SQL) document
@@ -18,26 +18,21 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     public static contextValue: string = "cosmosDBDocument";
     public readonly contextValue: string = DocDBDocumentTreeItem.contextValue;
     public readonly commandId: string = 'cosmosDB.openDocument';
+    public label: string;
 
-    private readonly partitionKeyValue: string | undefined | Object;
-
+    private readonly _partitionKeyValue: string | undefined | Object;
     private _document: RetrievedDocument;
     private _collection: DocDBCollectionTreeItem;
 
     constructor(collection: DocDBCollectionTreeItem, document: RetrievedDocument) {
         this._collection = collection;
         this._document = document;
-        this.partitionKeyValue = this.getPartitionKeyValue();
+        this.label = getDocumentTreeItemLabel(this._document);
+        this._partitionKeyValue = this.getPartitionKeyValue();
     }
 
     public get id(): string {
         return this.document.id;
-    }
-
-    public get label(): string {
-        const presentFields = documentDefaultFields.filter(element => this._document.hasOwnProperty(element));
-        const canonicalField = presentFields.find((element) => this._document[element] && this._document[element].toString() && !this._document[element].toString().startsWith("[object"));
-        return this._document[canonicalField].toString();
     }
 
     public get link(): string {
@@ -60,7 +55,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
         const result = await vscode.window.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
         if (result === DialogResponses.deleteResponse) {
             const client = this._collection.getDocumentClient();
-            const options = { partitionKey: this.partitionKeyValue };
+            const options = { partitionKey: this._partitionKeyValue };
             await new Promise((resolve, reject) => {
                 // Disabling type check in the next line. This helps ensure documents having no partition key value
                 // can still pass an empty object when required. It looks like a disparity between the type settings outlined here
@@ -83,7 +78,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
             throw new Error(`The "_self" and "_etag" fields are required to update a document`);
         }
         else {
-            let options = { accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this.partitionKeyValue };
+            let options = { accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this._partitionKeyValue };
             this._document = await new Promise<RetrievedDocument>((resolve, reject) => {
                 client.replaceDocument(
                     _self,
@@ -98,6 +93,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
                         }
                     });
             });
+            this.label = getDocumentTreeItemLabel(this._document);
             return this.document;
         }
     }
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -35,6 +35,10 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
         return this.document.id;
     }
 
+    public async refreshLabel() {
+        this._label = getDocumentTreeItemLabel(this._document);
+    }
+
     public get link(): string {
         return this.document._self;
     }
@@ -55,7 +59,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     }
 
     public async deleteTreeItem(_node: IAzureNode): Promise<void> {
-        const message: string = `Are you sure you want to delete document '${this._label}'?`;
+        const message: string = `Are you sure you want to delete document '${this.label}'?`;
         const result = await vscode.window.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
         if (result === DialogResponses.deleteResponse) {
             const client = this._collection.getDocumentClient();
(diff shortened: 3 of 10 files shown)


src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -33,7 +34,9 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     }
 
     public get label(): string {
-        return this.document.id;
+        const presentFields = documentDefaultFields.filter(element => this._document.hasOwnProperty(element));
+        const canonicalField = presentFields.find((element) => this._document[element] && this._document[element].toString() && !this._document[element].toString().startsWith("[object"));
```
PrashanthCorp: The three conditions in the find : 
1. Removes values being false-y (null, undefined) 
2. Removes cases where the toString is blank ""
3. Removes cases where `this._document[element]` is an Object which does not have it's own implementation. In such cases, the toString value returned is one of `[object Object]`, `[object null]`, `[object undefined]` etc. 

A simple type check does not work because `typeof <ObjectID>` is `object`  ¯\\_(ツ)_/¯

src/constants.ts
```diff
@@ -32,3 +32,5 @@ var isAccepted = collection.queryDocuments(
 
 if (!isAccepted) throw new Error('The query was not accepted by the server.');
 };`
+
+export const documentDefaultFields = ["name", "Name", "ID", "UUID", "id", "_id"];
```
StephenWeatherford: Are the DBs case-sensitive?

src/docdb/tree/DocDBDocumentTreeItem.ts
StephenWeatherford: typeof == "string"

StephenWeatherford: multiple calls to toString()

src/constants.ts
PrashanthCorp: Yes

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -18,24 +19,22 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     public readonly commandId: string = 'cosmosDB.openDocument';
 
     public readonly partitionKeyValue: string | undefined;
+    public label: string;
```
StephenWeatherford: private _label:string;

public get label() { 
  return this._label;
etc.

ditto for next class

src/constants.ts
StephenWeatherford: documentLabelFields?

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -87,6 +86,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
                         }
                     });
             });
+            this.label = getDocumentTreeItemLabel(this._document);
```
StephenWeatherford: don't you have to call refreshLabel or something?

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,22 @@ export function getNodeEditorLabel(node: IAzureNode): string {
 function isAccountTreeItem(treeItem: IAzureTreeItem): boolean {
     return (treeItem instanceof MongoAccountTreeItem) || (treeItem instanceof DocDBAccountTreeItemBase);
 }
+
+export function getDocumentTreeItemLabel(document: IMongoDocument | RetrievedDocument): string {
+    for (let field of documentDefaultFields) {
+        if (document.hasOwnProperty(field)) {
+            if (document[field]) { //ignore if false-y value: null, undefined, "".
```
StephenWeatherford: let value = document[field];
if (value) ...

Slightly better perf, but  more importantly, less chance of mistakes or botched bug fixes.

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,22 @@ export function getNodeEditorLabel(node: IAzureNode): string {
...
+                if (typeof document[field] === "string") {
+                    return document[field];
+                } else { // A simple instanceOf check doesn't seem to work. I run into this issue detailed here :
+                    //https://libertyseeds.ca/2015/01/03/Mongo-ObjectID-fails-instanceof-type-test-in-nodeunit/
+                    const result = document[field].toString();
+                    if (!result.startsWith("[object ")) {
```
StephenWeatherford: Still not 100% convinced, but I guess we'll see if we hit any issues.

StephenWeatherford:
What's the status of this?

src/docdb/tree/DocDBDocumentTreeItem.ts
PrashanthCorp: The update operation for the collection view in Mongo would have the side-effect of updating a document's label too. In order for that to work, the label field needs a setter. After the offline discussion, it looks like a private+getter+setter is preferred as opposed to public.

src/constants.ts
StephenWeatherford: Add "Id"?

src/docdb/editors/DocDBDocumentNodeEditor.ts
```diff
@@ -24,7 +24,9 @@ export class DocDBDocumentNodeEditor implements ICosmosEditor<RetrievedDocument>
     }
 
     public async update(document: RetrievedDocument): Promise<RetrievedDocument> {
-        return await this._documentNode.treeItem.update(document);
+        const updatedDoc = await this._documentNode.treeItem.update(document);
+        this._documentNode.refresh();
```
StephenWeatherford: Do you need to wait for the promise?

StephenWeatherford: I'd say you should unless there's a reason you specifically think we shouldn't.

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -52,7 +55,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     }
 
     public async deleteTreeItem(_node: IAzureNode): Promise<void> {
-        const message: string = `Are you sure you want to delete document '${this.label}'?`;
+        const message: string = `Are you sure you want to delete document '${this._label}'?`;
```
StephenWeatherford: nit: Since you have a reader for "label"  you should probably still use "label" here.

src/mongo/editors/MongoCollectionNodeEditor.ts
```diff
@@ -38,6 +38,8 @@ export class MongoCollectionNodeEditor implements ICosmosEditor<IMongoDocument[]
             const documentNode = documentNodes.find((node) => node.treeItem.document._id.toString() === updatedDoc._id.toString());
             if (documentNode) {
                 documentNode.treeItem.document = updatedDoc;
+                documentNode.treeItem.label = getDocumentTreeItemLabel(documentNode.treeItem.document);
```
StephenWeatherford: I think the right thing to do here is:

1) put the logic to set _label in refreshLabel()
2) call documentNode.refresh() whenever you want the label to get refreshed

src/constants.ts
StephenWeatherford: ?

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -18,23 +19,24 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
...
+        this._label = getDocumentTreeItemLabel(this._document);
     }
 
     public get id(): string {
         return this.document.id;
     }
 
-    public get label(): string {
-        return this.document.id;
+    public async refreshLabel() {
```
StephenWeatherford: return value
no need to be async

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,25 @@ export function getNodeEditorLabel(node: IAzureNode): string {
 function isAccountTreeItem(treeItem: IAzureTreeItem): boolean {
     return (treeItem instanceof MongoAccountTreeItem) || (treeItem instanceof DocDBAccountTreeItemBase);
 }
+
+export function getDocumentTreeItemLabel(document: IMongoDocument | RetrievedDocument): string {
+    for (let field of documentLabelFields) {
+        if (document.hasOwnProperty(field)) {
+            let value = document[field];
+            if (value) { //ignore if false-y value: null, undefined, "".
```
StephenWeatherford: This will also ignore zero, which I don't think would be desirable.

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,25 @@ export function getNodeEditorLabel(node: IAzureNode): string {
...
+                    //https://libertyseeds.ca/2015/01/03/Mongo-ObjectID-fails-instanceof-type-test-in-nodeunit/
+                    const result = value.toString();
+                    if (!result.startsWith("[object ")) {
+                        return result;
+                    }
+                }
+            }
+        }
+    }
+    return document["_id"].toString();
```
StephenWeatherford: Can _id be undefined or null?

StephenWeatherford:
Have you reviewed this yourself and looked for issues I might find?  There's at least one more issue I haven't mentioned. Thanks.

src/docdb/tree/DocDBDocumentTreeItem.ts
PrashanthCorp: Good observation. 

IAzureTreeItem asks for a `refreshLabel()` that returns a `Promise<void>` Not having it async or a return type of `void` instead of `Promise<void>` gives type errors.

PrashanthCorp: I'll add the return type on the function header to make it clearer. :)

src/mongo/editors/MongoDocumentNodeEditor.ts
```diff
@@ -25,7 +25,9 @@ export class MongoDocumentNodeEditor implements ICosmosEditor<IMongoDocument> {
     }
 
     public async update(document: IMongoDocument): Promise<IMongoDocument> {
-        return await this._documentNode.treeItem.update(document);
+        const updatedDoc = await this._documentNode.treeItem.update(document);
+        this._documentNode.refresh();
```
PrashanthCorp: Forgot to await this.

StephenWeatherford: Good catch

src/mongo/editors/MongoFindOneResultEditor.ts
```diff
@@ -34,12 +34,15 @@ export class MongoFindOneResultEditor implements ICosmosEditor<IMongoDocument> {
 
     public async update(newDocument: IMongoDocument): Promise<IMongoDocument> {
         const node = await this._tree.findNode(this.id);
+        let result: IMongoDocument;
         if (node) {
-            return (<IAzureNode<MongoDocumentTreeItem>>node).treeItem.update(newDocument);
+            result = await (<IAzureNode<MongoDocumentTreeItem>>node).treeItem.update(newDocument);
```
StephenWeatherford: This is the other issue I saw.  Why was the return removed here?

src/utils/vscodeUtils.ts
StephenWeatherford: I'd leave it as _id, since we'll want to change documentLabelFields to be user-settable anyway.  Looks like previous code had return document._id.toString(). Still, I'd rather be conservative and not assume this won't give a null ref exception.

I think that String(document._id) is safe for null/undefined?  Can you verify? Thx.

src/mongo/editors/MongoFindOneResultEditor.ts
PrashanthCorp: My bad. We need to kick off the refresh before returning, but after updating the value. 
Should return the result.

src/utils/vscodeUtils.ts
PrashanthCorp: Good catch. 

Looking at the other [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), I think we'll be covered after acconting for 0.

StephenWeatherford: In this case I'd suggest the following:

I'd suggest this:
(note: you're effectively checking that it's not an object with your value.toString().startsWith("[object ")) below.  I'm not sure this idea behavior when the property to display is an object or array, but it seems reasonable to skip it and try something else.  But typeof is a lot better than the toString.startsWith()

// If doesn't have value, or is null, ObjectId or other object type, try a different property
if (typeof value !== 'undefined' && typeof value !== 'object') {
    return String(value);
}

I think that handles it all, right?

PrashanthCorp: Yeah. It relies on null having type object (slightly obfuscated). If that's well known, then I'll go ahead with it.

src/mongo/editors/MongoFindOneResultEditor.ts
```diff
@@ -38,10 +38,11 @@ export class MongoFindOneResultEditor implements ICosmosEditor<IMongoDocument> {
         if (node) {
             result = await (<IAzureNode<MongoDocumentTreeItem>>node).treeItem.update(newDocument);
             node.refresh();
+        } else {
+            // If the node isn't cached already, just update it to Mongo directly (without worrying about updating the tree)
+            const db = await this._databaseNode.treeItem.getDb();
+            result = await MongoDocumentTreeItem.update(db.collection(this._collectionName), newDocument);
```
PrashanthCorp: Felt this control flow looked clearer. If you prefer a return statement after each block (functionally the same), I'll do that.

StephenWeatherford: I'm fine with either one, really depends on the rest of the code as to which is easier to read/maintain (e.g. how many control flows, how big the then case is, etc.)

src/utils/vscodeUtils.ts
StephenWeatherford: Unfortunately, typeof null = "object" in JS.  Well-known "bug" that can't be fixed now.

src/constants.ts
```diff
@@ -34,3 +34,5 @@ export const defaultStoredProcedure =
 };` ;
 
 export const emptyPartitionKeyValue = {};
+
+export const documentLabelFields = ["name", "Name", "ID", "UUID", "id", "_id"];
```
StephenWeatherford: Please add "Id", "uuid"

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,17 @@ export function getNodeEditorLabel(node: IAzureNode): string {
 function isAccountTreeItem(treeItem: IAzureTreeItem): boolean {
     return (treeItem instanceof MongoAccountTreeItem) || (treeItem instanceof DocDBAccountTreeItemBase);
 }
+
+export function getDocumentTreeItemLabel(document: IMongoDocument | RetrievedDocument): string {
+    for (let field of documentLabelFields) {
+        if (document.hasOwnProperty(field)) {
+            let value = document[field];
+            if (value || value === 0) { //ignore if false-y value: null, undefined, "".
```
StephenWeatherford: You were supposed to remove this line:

 if (value || value === 0)

src/utils/vscodeUtils.ts
```diff
@@ -101,3 +104,17 @@ export function getNodeEditorLabel(node: IAzureNode): string {
...
+                if (value != 'undefined' && typeof value !== 'object') {
```
StephenWeatherford: typeof value

StephenWeatherford:
@PrashanthCorp CLA will pass if you push a new commit

PrashanthCorp:
Ping.

Issue #591, opened 2018-05-01
"Cannot read property 'errors' of undefined" trying to execute scrapbook with error


StephenWeatherford:
1) Put this into a scrapbook:

adb.testdrop()

2) SHIFT+CTRL+'
!) Cannot read property 'errors' of undefined

StephenWeatherford:
#596 is a partial fix - just gets rid of the null ref error. Still need to be able to surface these errors as part of #471.

StephenWeatherford:
Fixed as part of improved error handling.

Issue #471, opened 2018-04-02
Parser - Handle error nodes better


PrashanthCorp:
[Currently](https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/mongo/grammar/visitors.ts#L67), we mainly print an error message with the location of the encountered error. 

A richer experience could have text highlighting, or something _at the position of erroneous text_. 

@fiveisprime  thoughts?

StephenWeatherford:
@PrashanthCorp I think the next step is to select the error region in the text document, and see how well that works.  That might be good enough.

StephenWeatherford:
Yes, we need to stop throwing in the visit, and instead implement something that can return the errors.

StephenWeatherford:
Red squiggles would be ideal.

StephenWeatherford:
I'm taking this one, got tired of current situation.
#652

PR #596, opened 2018-05-01
Don't reference badCommand.errors when !badCommand

StephenWeatherford:


StephenWeatherford:
Partial fix for #591 - just gets rid of the null ref error.  Still need to be able to surface these errors as part of #471.

StephenWeatherford:
visitErrorNode(node: ErrorNode): MongoCommand[] {
		const position = new vscode.Position(node._symbol.line - 1, node._symbol.charPositionInLine); // Symbol lines are 1 indexed. Position lines are 0 indexed
		const text = node.text;
		const badCommand = this.commands.find((command) => command.range && command.range.contains(position));
		if (badCommand) {
			// Need a place to hang errors that occur when no command is actually recognized
			badCommand.errors = badCommand.errors || [];
			badCommand.errors.push({ position: position, text: text });

PR #597, opened 2018-05-02
Pass partitionKey values appropriate to type of collection and document

### LINKED ISSUE #544: Fail to delete document with error for SQL accounts
Repro steps:
Right click one document and select 'Delete Document', then check whether the document can be deleted or not.

Action: cosmosDB.deleteDocDBDocument
Error type: BadRequest
Error Message: The partition key supplied in x-ms-partitionkey header has fewer components than defined in the the collection.

More Info:
This issue doesn't reproduce in Storage Explorer. The same document can be deleted successfully in Storage Explorer.

Version: 0.6.1-alpha
OS: win32

PrashanthCorp:
Fixes #544.

### DIFF
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -25,7 +25,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     constructor(collection: DocDBCollectionTreeItem, document: RetrievedDocument) {
         this._collection = collection;
         this._document = document;
-        this.partitionKeyValue = this.getPartitionKeyValue() || "{}";
+        this.partitionKeyValue = this.getPartitionKeyValue();
     }
 
     public get id(): string {
@@ -56,9 +56,14 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
         const result = await vscode.window.showWarningMessage(message, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
         if (result === DialogResponses.deleteResponse) {
             const client = this._collection.getDocumentClient();
-            const options = { partitionKey: this.partitionKeyValue }
+            const options = { partitionKey: this.partitionKeyValue || {} }
             await new Promise((resolve, reject) => {
-                client.deleteDocument(this.link, options, function (err) {
+                // Disabling type check in the next line. This helps ensure documents having no partition key value
+                // can still pass an empty object. It looks like a disparity between the type settings  out lined
+                // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/01e0ffdbab16b15c702d5b8c87bb122cc6215a59/types/documentdb/index.d.ts#L72
+                // and the workaround outlined at https://github.com/Azure/azure-documentdb-node/issues/222#issuecomment-364286027
+                // tslint:disable-next-line:no-any
+                client.deleteDocument(this.link, <any>options, function (err) {
                     err ? reject(err) : resolve();
                 });
             });
--- a/src/docdb/tree/DocDBDocumentTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentTreeItem.ts
@@ -8,6 +8,7 @@ import * as path from 'path';
 import { IAzureNode, IAzureTreeItem, UserCancelledError, DialogResponses } from 'vscode-azureextensionui';
 import { RetrievedDocument, DocumentClient } from 'documentdb';
 import { DocDBCollectionTreeItem } from './DocDBCollectionTreeItem';
+import { emptyPartitionKeyValue } from '../../constants';
 
 /**
  * Represents a Cosmos DB DocumentDB (SQL) document
@@ -17,7 +18,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     public readonly contextValue: string = DocDBDocumentTreeItem.contextValue;
     public readonly commandId: string = 'cosmosDB.openDocument';
 
-    public readonly partitionKeyValue: string | undefined | Object;
+    private readonly partitionKeyValue: string | undefined | Object;
 
     private _document: RetrievedDocument;
     private _collection: DocDBCollectionTreeItem;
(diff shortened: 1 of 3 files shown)


src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -103,8 +112,8 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
         let value;
         for (let field of fields) {
             value = value ? value[field] : this.document[field];
-            if (!value) {
-                break;
+            if (!value) { //Partition Key exists, but this document doesn't have a value
+                return {};
```
PrashanthCorp: Not a fan of writing a comment here. Was wondering if it makes more sense to have the following constants: 
```
fixedSizeCollectionPartitionKeyValue = undefined;
UnlimitedCollectionNoPartitionKeyValue= {};
```

The downside being it's only used in one place (here).

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -74,11 +79,15 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
...
+            let options = { accessCondition: { type: 'IfMatch', condition: newData._etag } };
+            if (this.partitionKeyValue) { //empty object is truth-y
+                options["partitionKey"] = this.partitionKeyValue;
+            }
             this._document = await new Promise<RetrievedDocument>((resolve, reject) => {
                 client.replaceDocument(
                     _self,
                     newData,
-                    { accessCondition: { type: 'IfMatch', condition: newData._etag }, partitionKey: this.partitionKeyValue },
+                    options,
```
StephenWeatherford: Why'd you remove the access condition?

src/docdb/tree/DocDBDocumentTreeItem.ts
StephenWeatherford: FWIW, don't confuse the idea of fixed vs unlimited with the idea of having a partition key - they might be the same right now in Azure, but not necessarily always.

StephenWeatherford: Have you thought about whether a partition key with no value is the same as a partition key with {} as the value?

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -74,11 +79,15 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
             throw new Error(`The "_self" and "_etag" fields are required to update a document`);
         }
         else {
+            let options = { accessCondition: { type: 'IfMatch', condition: newData._etag } };
+            if (this.partitionKeyValue) { //empty object is truth-y
```
StephenWeatherford: Is it necessary to avoid assigning the partition key value if it's falsy?  How you considered whether 0, empty string, null, etc. are valid partition keys?

src/docdb/tree/DocDBDocumentTreeItem.ts
PrashanthCorp: Yes. Here's what I tested: 
1.Create a doc in the portal, with the partition key value of {}, no quotes. 
2. Open it in the extension. Doc opens fine, empty object has no quotes. 
3. Add another field and update. Open another document and open this document again. Works fine. 
4. Remove the partition key, update. Open another document and open this document again. Works fine. 
5. Add back the empty object partition key value. Update. Open another document and open this document again.  Works fine. 

My hypothesis is that at the server side, documents with either value ({} or no value) are placed in the same partition, and hence it has had no problem locating the document.  But the API expects a value, and the 'no partition key' value happens to be {}. 


Hope this answers the question.

src/docdb/tree/DocDBDocumentTreeItem.ts
PrashanthCorp: This is the error returned for a fixed size collection, if I pass 0, "" or {} as the partitionKey for a collection with no partition key (aka fixed size at this point): 
`Error: Partition key provided either doesn't correspond to definition in the collection or doesn't match partition key field values specified in the document.`

Currently, this is how the API likes to receive partitionKeys: 
1. Key exists for collection, document has a value - give it the value, 
2. Key exists for collection, document has no value - give it an empty object, as outlined [in the node-documentdb workaround](https://github.com/Azure/azure-documentdb-node/issues/222)
3. No key exists (_currently_ corresponds to fixed size) do not give it a partition key value.

StephenWeatherford: Again, I'm not sure why the "if" is even necessary. If there's no partition key, partitionKeyValue will be undefined, and set as undefined into the options.  I wouldn't expect that to be a problem (although I have seen code where it is).

src/docdb/tree/DocDBDocumentTreeItem.ts
StephenWeatherford: Not fully happy with setting the partition key value to {} when it has no value.  It feels like a hack (although I realize it's Azure that's enforcing it).  Personally it seems like returning undefined makes more sense, and then converting on the update/delete side.  Otherwise you're fixing up a problem at update/delete time by making a chance in code farther away.

But not sure what other effects either of these choices causes - is partitionKey used in other places where the undefined vs {} might be an issue?

If you do go with {} here, then I agree having a constant for it would be a good idea.

src/docdb/tree/DocDBDocumentTreeItem.ts
PrashanthCorp: I placed it there due a typescript compile issue (concerning the type of the partitionKey field). I'll move it back and <any> cast it. Seems better that way.

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -103,7 +101,7 @@ export class DocDBDocumentTreeItem implements IAzureTreeItem {
     private getPartitionKeyValue(): string | undefined | Object {
         const partitionKey = this._collection.partitionKey;
         if (!partitionKey) { //Fixed collections -> no partitionKeyValue
-            return undefined;
+            return 0;
```
StephenWeatherford: ?

Undefined seems the appropriate return value.

PR #600, opened 2018-05-02
Better error message for missing connection string

### LINKED ISSUE #599: Failed to load attached account
screen shot 2018-05-02 at 9 51 19 am

I'm not sure how my attached account got into this state, but the connection string was missing from keytar. I feel like we just need a better error message if the value returned by 'getPassword' [here](https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/tree/AttachedAccountsTreeItem.ts#L239) is undefined/null/empty string

StephenWeatherford:
Fixes #599

### DIFF
--- a/src/mongo/tree/MongoAccountTreeItem.ts
+++ b/src/mongo/tree/MongoAccountTreeItem.ts
@@ -46,6 +46,10 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
         try {
             let databases: IDatabaseInfo[];
 
+            if (!this.connectionString) {
+                throw new Error('Missing connection string');
+            }
+
             db = await MongoClient.connect(this.connectionString);
             let databaseInConnectionString = getDatabaseNameFromConnectionString(this.connectionString);
             if (databaseInConnectionString && !this.isEmulator) { // emulator violates the connection string format


src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -46,6 +46,10 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
         try {
             let databases: IDatabaseInfo[];
 
+            if (!this.connectionString) {
+                throw new Error('Missing connection string');
```
EricJizbaMSFT: How about 'Failed to retrieve connection string.'

src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -46,6 +46,10 @@ export class MongoAccountTreeItem implements IAzureParentTreeItem {
         try {
```
EricJizbaMSFT: Technically, the same problem could happen for non-Mongo accounts

StephenWeatherford: For other accounts, you get this:
image

Fixing it to be nicer requires deeper changes than I think are worth the risk unless we get feedback that this is happening a lot.

EricJizbaMSFT: Fine with me. I think the effort for this fix accurately matched the priority of the bug

PR #617, opened 2018-05-07
Implement tslint: no-var-keyword. Replace var -> let wherever applicable.

PrashanthCorp:


PrashanthCorp:
1. Had a rel/0.8.0 branch because I didn't feel these changes should go into master immediately. If there's any P1's that warrant a 0.7.1, I didn't want these pushes to add to it. 

2. Apologies for the WIP tag. I realized I had more changes to make after the push, so I added the tag. I have no more changes to make, so I removed it.

EricJizbaMSFT:
I see what you're saying, but I think you have it backwards. You should always use the master branch for your normal day-to-day work (including this stuff). If we need a 0.7.1 release, we will create a 'rel/0.7.1' branch directly from https://github.com/Microsoft/vscode-cosmosdb/tree/v0.7.0

PR #620, opened 2018-05-08
Implement tslint rule 'semicolon' -

PrashanthCorp:
Ensures properties/expressions are separated by semicolon.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -251,7 +251,7 @@
         ],
         "react-tsx-curly-spacing": true,
         "semicolon": [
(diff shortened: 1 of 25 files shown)


tslint.json
```diff
@@ -251,7 +251,7 @@
         ],
         "react-tsx-curly-spacing": true,
         "semicolon": [
-            false, // TODO: true,
+            true, // TODO: true,
```
EricJizbaMSFT: You can remove the comment "// TODO: true" since you fixed this

PR #624, opened 2018-05-09
Handle unexpected defaultExp values when labelling account node

PrashanthCorp:


src/tree/CosmosDBAccountProvider.ts
```diff
@@ -74,7 +74,17 @@ export class CosmosDBAccountProvider implements IChildProvider {
     private async initChild(client: CosmosDBManagementClient, databaseAccount: DatabaseAccount): Promise<IAzureTreeItem> {
         const defaultExperience = <API>databaseAccount.tags.defaultExperience;
         const resourceGroup: string = azureUtils.getResourceGroupFromId(databaseAccount.id);
-        const label: string = `${databaseAccount.name} (${getExperience(defaultExperience).shortName})`;
+        let accountKind;
+        try {
+            accountKind = getExperience(defaultExperience).shortName;
```
EricJizbaMSFT: I don't like this because getExperience could throw a RangeError in a bunch of different places (aka it's error-prone). Just change getExperience so that it doesn't throw a RangeError

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -227,8 +227,8 @@ export class AttachedAccountsTreeItem implements IAzureParentTreeItem {
                     // Default to Mongo if the value is a string for the sake of backwards compatiblity
                     // (Mongo was originally the only account type that could be attached)
```
EricJizbaMSFT: Your branches aren't correct. It should something like "ps/fixRangeError" merging into "rel/0.7.1"

"rel/0.7.1" should be a clean branch off of "v0.7.0"

EricJizbaMSFT: Also you might as well do the changelog/version bump in this PR

PR #637, opened 2018-05-22
 Filter out invalid accounts under a subscription

### LINKED ISSUE #628: An invalid Database Account should not prevent valid Accounts from being displayed
See [this bug](https://github.com/Microsoft/vscode-cosmosdb/issues/623) where a single invalid Database Account was preventing _any_ database account from being shown. The specific error in that bug was already fixed, but other errors could still occur.

We had a similar issue in App Service that was already fixed:
Issue: https://github.com/Microsoft/vscode-azureappservice/issues/371
PR with fix: https://github.com/Microsoft/vscode-azureappservice/pull/373

PrashanthCorp:
Fixes #628

### DIFF
--- a/src/tree/CosmosDBAccountProvider.ts
+++ b/src/tree/CosmosDBAccountProvider.ts
@@ -30,22 +30,19 @@ export class CosmosDBAccountProvider implements IChildProvider {
     public async loadMoreChildren(node: IAzureNode): Promise<IAzureTreeItem[]> {
         const client = new CosmosDBManagementClient(node.credentials, node.subscriptionId);
         const accounts: DatabaseAccountsListResult = await client.databaseAccounts.list();
-        let validAccounts = await Promise.all(
+        let accountTreeItems = [];
+        await Promise.all(
             accounts.map(async (databaseAccount: DatabaseAccount) => {
                 try {
-                    let result = await this.initChild(client, databaseAccount);
-                    return result;
+                    let account = await this.initChild(client, databaseAccount);
+                    accountTreeItems.push(account);
                 } catch (e) {
-                    const output = util.getOutputChannel();
-                    output.appendLine(e);
-                    output.show();
-                    vscode.window.showInformationMessage(`Failed to pick laod invalid account. See output window for more details.`);
-                    return null;
+                    accountTreeItems.push(<IAzureTreeItem>{ id: e.toString(), label: e.toString() });
                 }
 
             })
         );
-        return validAccounts.filter((element) => !!element);
+        return accountTreeItems;
     }
 
     public async createChild(node: IAzureNode, showCreatingNode: (label: string) => void, actionContext?: IActionContext): Promise<IAzureTreeItem> {
--- a/src/tree/CosmosDBAccountProvider.ts
+++ b/src/tree/CosmosDBAccountProvider.ts
@@ -103,10 +103,9 @@ export class CosmosDBAccountProvider implements IChildProvider {
                     return new GraphAccountTreeItem(databaseAccount.id, label, databaseAccount.documentEndpoint, gremlinEndpoint, keyResult.primaryMasterKey, isEmulator);
                 }
                 case "DocumentDB":
+                default:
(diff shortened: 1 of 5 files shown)


src/tree/CosmosDBAccountProvider.ts
```diff
@@ -30,10 +30,19 @@ export class CosmosDBAccountProvider implements IChildProvider {
...
         const client = new CosmosDBManagementClient(node.credentials, node.subscriptionId);
         const accounts: DatabaseAccountsListResult = await client.databaseAccounts.list();
+        let accountTreeItems = [];
+        await Promise.all(
+            accounts.map(async (databaseAccount: DatabaseAccount) => {
+                try {
+                    let account = await this.initChild(client, databaseAccount);
+                    accountTreeItems.push(account);
+                } catch (e) {
+                    accountTreeItems.push(<IAzureTreeItem>{ id: e.toString(), label: e.toString() });
```
StephenWeatherford: label: That's what parseError is for.  e could be anything - a string, a number, even null

id: If you get the same error for multiple nodes, you'll get duplicate ID's, which I think causes vscode to throw.  Does Eric's PR have this issue?

src/tree/CosmosDBAccountProvider.ts
```diff
@@ -14,7 +14,7 @@ import { DatabaseAccountsListResult, DatabaseAccount, DatabaseAccountListKeysRes
 import { TryGetGremlinEndpointFromAzure } from '../graph/gremlinEndpoints';
 import { ICosmosDBWizardContext } from './CosmosDBAccountWizard/ICosmosDBWizardContext';
 import { CosmosDBAccountNameStep } from './CosmosDBAccountWizard/CosmosDBAccountNameStep';
-import * as vscodeUtil from '../utils/vscodeUtils';
+import * as util from '../utils/vscodeUtils';
```
StephenWeatherford: why?

src/tree/CosmosDBAccountProvider.ts
EricJizbaMSFT: Just leave the id out - it's optional.

But even then, it would be weird to have multiple labels with the same message. I would prefer if the label was still the account name and the description was 'invalid'. If you want to display the error message, I would make it a child node.

(If we don't even know the account name, I just wouldn't display the node in the tree. I don't think there's much we can do there)

src/tree/CosmosDBAccountProvider.ts
```diff
@@ -37,7 +37,8 @@ export class CosmosDBAccountProvider implements IChildProvider {
                     let account = await this.initChild(client, databaseAccount);
                     accountTreeItems.push(account);
                 } catch (e) {
-                    accountTreeItems.push(<IAzureTreeItem>{ id: e.toString(), label: e.toString() });
+                    const err = parseError(e);
+                    accountTreeItems.push(<IAzureTreeItem>{ id: err.toString(), label: err.toString() });
```
StephenWeatherford: Uh, no...  parseError(e).message

src/tree/CosmosDBAccountProvider.ts
```diff
@@ -38,7 +38,8 @@ export class CosmosDBAccountProvider implements IChildProvider {
                     accountTreeItems.push(account);
                 } catch (e) {
                     const err = parseError(e);
-                    accountTreeItems.push(<IAzureTreeItem>{ id: err.toString(), label: err.toString() });
+                    //tslint:disable-next-line:no-non-null-assertion
+                    accountTreeItems.push(<IAzureTreeItem>{ label: databaseAccount!.name, description: "Invalid: " + err.message });
```
StephenWeatherford: Can you paste a screenshot?  I'm not sure the "Invalid:" will be necessary, and it will reduce space.

PrashanthCorp: image

StephenWeatherford: Yeah, I'd suggest getting rid of "Invalid:"

PR #638, opened 2018-05-23
[SQL]: Allow setting a partitionkey value during document creation

PrashanthCorp:


src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -77,4 +88,20 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
 
         throw new UserCancelledError();
     }
+
+    // Create a nested Object given the partition key path and value
+    private createPartitionPathObject(partitionKey, partitionKeyValue): Object {
```
StephenWeatherford: types for params

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -77,4 +88,20 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
...
+        //remove leading slash
+        partitionKey = partitionKey.slice(1);
```
StephenWeatherford: Is this guaranteed to always start with a slash?

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -61,9 +61,20 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
 
         if (docID || docID === "") {
             docID = docID.trim();
+            let body = { 'id': docID };
+            const partitionKey: string | undefined = this._collection.partitionKey && this._collection.partitionKey.paths[0];
+            if (partitionKey) {
+                const partitionKeyValue: string = await vscode.window.showInputBox({
+                    prompt: `The partition key is ${partitionKey}. Enter a value for the partition key`,
```
StephenWeatherford: suggest: Enter a value for the partition key (${partitionKey)}

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -61,9 +61,20 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
 
         if (docID || docID === "") {
             docID = docID.trim();
+            let body = { 'id': docID };
+            const partitionKey: string | undefined = this._collection.partitionKey && this._collection.partitionKey.paths[0];
```
StephenWeatherford: paths[0] is checking that the first array value has a value, but will throw if the array is empty.  Did you mean && this._collection.partitionKey.paths.length
?

src/docdb/tree/DocDBDocumentsTreeItem.ts
StephenWeatherford: Not sure we should assign a key value if the user enters ENTER.  What do you think?

PrashanthCorp: It currently creates the nested object with an empty string. 

Looks like empty strings aren't the same as no partition key. Thanks for the catch!

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -65,12 +65,14 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
...
                 const partitionKeyValue: string = await vscode.window.showInputBox({
-                    prompt: `The partition key is ${partitionKey}. Enter a value for the partition key`,
+                    prompt: `Enter a value for the partition key ("${partitionKey}")`,
                     ignoreFocusOut: true
                 });
-                // We cannot pass a partition key value during document creation.
-                // We need to present the partitionKey value as part of the document contents
-                Object.assign(body, this.createPartitionPathObject(partitionKey, partitionKeyValue));
+                if (partitionKeyValue) {
+                    // We cannot pass a partition key value during document creation.
```
StephenWeatherford: Nit: comment is confusing - this *is* document creation.

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -90,9 +92,11 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
     }
 
     // Create a nested Object given the partition key path and value
-    private createPartitionPathObject(partitionKey, partitionKeyValue): Object {
+    private createPartitionPathObject(partitionKey: string, partitionKeyValue: string): Object {
         //remove leading slash
-        partitionKey = partitionKey.slice(1);
+        if (partitionKey[0] === '/') {
```
StephenWeatherford: You'd also need to check that it's not empty

PR #652, opened 2018-05-31
Show errors in the error list

StephenWeatherford:
This displays scrapbook parse errors in the Errors pane.  Right now it's pretty ugly due to various error-handling issues, but I'm enabling it now because it makes it a lot easier to see those other problems.  We can easily disable before shipping if we need to but hopefully the others will be fixed first.

image

Depends on https://github.com/Microsoft/vscode-azuretools/pull/189

### DIFF
--- a/src/mongo/registerMongoCommands.ts
+++ b/src/mongo/registerMongoCommands.ts
@@ -18,7 +18,7 @@ import { reporter } from "../utils/telemetry";
 import { MongoCodeLensProvider } from "./services/MongoCodeLensProvider";
 import { ext } from "../extensionVariables";
 import { executeCommandFromText, executeCommandFromActiveEditor, executeAllCommandsFromActiveEditor, getAllErrorsFromTextDocument } from "./MongoScrapbook";
-import TelemetryReporter from "../../../vscode-azuretools/ui/node_modules/vscode-extension-telemetry";
+import TelemetryReporter from "vscode-extension-telemetry";
 
 const connectedDBKey: string = 'ms-azuretools.vscode-cosmosdb.connectedDB';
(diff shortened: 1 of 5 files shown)


src/mongo/registerMongoCommands.ts
```diff
@@ -148,3 +163,23 @@ function setConnectedNode(node: IAzureNode | undefined, codeLensProvider: MongoC
     let dbName = node && node.treeItem.label;
     codeLensProvider.setConnectedDatabase(dbName);
 }
+
+function updateErrorsInActiveDocument(context: vscode.ExtensionContext): void {
+    callWithUnxpectedErrorTelemetry(
```
EricJizbaMSFT: See my comments in this PR: https://github.com/Microsoft/vscode-azuretools/pull/189, but I'm not sure I see the value in callWithUnxpectedErrorTelemetry.

I would prefer to follow the pattern of  [other events](https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L117), which use `actionHandler.registerEvent` and `this.suppressTelemetry`

StephenWeatherford:
Yes. please see also my comments in the description.



Sent via the Samsung Galaxy S7, an AT&T 4G LTE smartphone

PrashanthCorp:
It looks like the `callWithUnxpectedErrorTelemetry` call is missing an import, leading to the failing build.

StephenWeatherford:
> It looks like the callWithUnxpectedErrorTelemetry call is missing an import, leading to the failing build.

That's because the PR for the common package has to be merged first as I mentioned above with the dependency note.

PR #654, opened 2018-06-04
Put in place triple-equals, restrict-plus-operands; ignore no-unnecessary-local

PrashanthCorp:


src/graph/GraphViewServer.ts
```diff
@@ -436,6 +436,7 @@ export class GraphViewServer extends EventEmitter {
   private handleQueryMessage(queryId: number, gremlin: string) {
     this.log(`Query requested: queryId=${queryId}, gremlin="${gremlin}"`);
 
+    //tslint:disable-next-line:no-floating-promises
     this.queryAndShowResults(queryId, gremlin);
```
PrashanthCorp: `handleQueryMessage` is not async, and is used in the GraphViewServerSocket. 
@StephenWeatherford should this be awaited?

tslint.json
```diff
@@ -162,7 +162,7 @@
         "no-suspicious-comment": true,
         "no-typeof-undefined": true,
         "no-unnecessary-field-initialization": true,
-        // TODO: "no-unnecessary-local-variable": true,
+        "no-unnecessary-local-variable": false, //changed
```
PrashanthCorp: Not enforcing this, as per discussion with @StephenWeatherford

src/graph/GraphViewServer.ts
StephenWeatherford: Why are you thinking this one should be left alone?

StephenWeatherford: Could you add return value to handleQueryMessage?

Makes sense, can't handle this synchronously.

FYI, that would normally be a red flag - it means we could get overlapping calls.  Luckily this is already being handled at least somewhat - I associate each query with an ID, and the client ignores results that are not from the most recent query, so we're probably fine here.

StephenWeatherford:
You gonna merge this?

PrashanthCorp:
Merging. I expected some merge conflicts & build failures for the other PR's when this is merged, but this has probably waited enough

Issue #659, opened 2018-06-06
Cannot read property 'text' of undefined


StephenWeatherford:
1) Type this in connected scrapbook:

db.c1

2) Type "period"
!) 	at DefaultErrorStrategy.recoverInline (C:\Repos\vscode-cosmosdb\node_modules\antlr4ts\DefaultErrorStrategy.js:425:15)
	at mongoParser.match (C:\Repos\vscode-cosmosdb\node_modules\antlr4ts\Parser.js:143:34)
	at mongoParser.functionCall (c:\Repos\vscode-cosmosdb\src\mongo\grammar\mongoParser.ts:323:37)
	at mongoParser.command (c:\Repos\vscode-cosmosdb\src\mongo\grammar\mongoParser.ts:234:14)
	at mongoParser.commands (c:\Repos\vscode-cosmosdb\src\mongo\grammar\mongoParser.ts:165:15)
	at getAllCommands (c:\Repos\vscode-cosmosdb\src\mongo\MongoScrapbook.ts:127:65)
	at getAllCommandsFromTextDocument (c:\Repos\vscode-cosmosdb\src\mongo\MongoScrapbook.ts:70:9)
	at Object.getAllErrorsFromTextDocument (c:\Repos\vscode-cosmosdb\src\mongo\MongoScrapbook.ts:26:17)
	at updateErrorsInActiveScrapbook (c:\Repos\vscode-cosmosdb\src\mongo\registerMongoCommands.ts:208:22)
	at Object.<anonymous> (c:\Repos\vscode-cosmosdb\src\mongo\registerMongoCommands.ts:187:13)
	at Generator.next (<anonymous>)
	at __awaiter (C:\Repos\vscode-cosmosdb\out\src\mongo\registerMongoCommands.js:11:71)
	at Promise (<anonymous>)
	at __awaiter (C:\Repos\vscode-cosmosdb\out\src\mongo\registerMongoCommands.js:7:12)
	at Object.<anonymous> (C:\Repos\vscode-cosmosdb\out\src\mongo\registerMongoCommands.js:180:16)
	at Object.<anonymous> (c:\Repos\vscode-azuretools\ui\src\AzureActionHandler.ts:50:37)
	at Object.<anonymous> (c:\Repos\vscode-azuretools\ui\src\callWithTelemetryAndErrorHandling.ts:35:47)
	at Generator.next (<anonymous>)
[...]

PR #660, opened 2018-06-07
Cannot read property text of undefined

### LINKED ISSUE #609: CodeLens disappears/reappears as you type commands
1) New mongo scrapbook
2) type "d"
!) CodeLens disappears/reappears
3) finish typing "db.c1."
!) CodeLens disappears
4) type "f"
!) CodeLens reappears

StephenWeatherford:
Fixes #609, #659 

Also adds telemetry for provideCodeLenses (failures only).

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -167,7 +167,7 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 
 	visitFunctionCall(ctx: mongoParser.FunctionCallContext): MongoCommand[] {
 		if (ctx.parent instanceof mongoParser.CommandContext) {
-			this.commands[this.commands.length - 1].name = ctx._FUNCTION_NAME.text;
+			this.commands[this.commands.length - 1].name = (ctx._FUNCTION_NAME && ctx._FUNCTION_NAME.text) || "";
 		}
 		return super.visitFunctionCall(ctx);
 	}
--- a/src/mongo/services/MongoCodeLensProvider.ts
+++ b/src/mongo/services/MongoCodeLensProvider.ts
@@ -5,6 +5,8 @@
 
 import * as vscode from "vscode";
 import { getAllCommandsFromTextDocument } from "../MongoScrapbook";
+import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
+import TelemetryReporter from "vscode-extension-telemetry";
 
 export class MongoCodeLensProvider implements vscode.CodeLensProvider {
 	private _onDidChangeEmitter = new vscode.EventEmitter<void>();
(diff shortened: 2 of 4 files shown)


src/mongo/services/MongoCodeLensProvider.ts
```diff
@@ -13,55 +15,66 @@ export class MongoCodeLensProvider implements vscode.CodeLensProvider {
...
 		this._connectedDatabaseInitialized = true;
 		this._onDidChangeEmitter.fire();
 	}
 
 	public provideCodeLenses(document: vscode.TextDocument, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.CodeLens[]> {
-		let isInitialized = this._connectedDatabaseInitialized;
-		let isConnected = !!this._connectedDatabase;
-		let database = isConnected && this._connectedDatabase;
-		let lenses: vscode.CodeLens[] = [];
+		const me = this;
```
PrashanthCorp: Is `me` and abbreviation for something? I'm not too sure if the `this` variable stands for anything aside from MongoCodeLensProvider.

src/mongo/MongoScrapbook.ts
```diff
@@ -167,7 +167,7 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 
 	visitFunctionCall(ctx: mongoParser.FunctionCallContext): MongoCommand[] {
 		if (ctx.parent instanceof mongoParser.CommandContext) {
-			this.commands[this.commands.length - 1].name = ctx._FUNCTION_NAME.text;
+			this.commands[this.commands.length - 1].name = (ctx._FUNCTION_NAME && ctx._FUNCTION_NAME.text) || "";
```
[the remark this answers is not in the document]
StephenWeatherford: I thought you originally wrote this...?

It's probably just an auto-generated name, I doubt it's intended to be private.

src/mongo/services/MongoCodeLensProvider.ts
StephenWeatherford: This is a disadvantage of the current implementation of IActionContext.  Since we're getting an action context passed in as "this", we can't use the normal arrow functions and this functionality of TypeScript.  I'm falling back to a standard JavaScript pattern - same the pointer to the class instance to "me" so we can access it instead of "this".

https://stackoverflow.com/questions/13364503/javascript-me-this-why

src/mongo/MongoScrapbook.ts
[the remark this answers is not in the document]
PrashanthCorp: Stems from here: https://github.com/Microsoft/vscode-cosmosdb/commit/a6af2cbbf6e98f19936b584d12f7f9b056d36fb4, inherited code. 

Makes sense.

src/mongo/services/MongoCodeLensProvider.ts
StephenWeatherford: (TypeScript fixes "this" in arrow functions)

PrashanthCorp: Thanks for the link. I've faced this problem, but I usually name it according to class. Didn't know it had a name. 

If I understand correctly, your way of doing it is a nicer, standardized way of `var that = this` to be used inside a callback/function scope.

I don't fully agree on the name `me`, but I wouldn't want to force naming a hardly used variable. 
Suggestion: Can you annotate the type of `me` so that it's clearer what to expect from `me`?

StephenWeatherford: this vs me is a matter of preference but I've seen me a lot and I think it's more descriptive.

StephenWeatherford: Added:

const me: MongoCodeLensProvider = this;

PR #685, opened 2018-06-15
Handle single quotes in scrapbook

PrashanthCorp:


grammar/mongo.g4
```diff
@@ -33,7 +33,7 @@ arguments:
 
 argument: literal | objectLiteral | arrayLiteral;
 
-objectLiteral: '{' propertyNameAndValueList? ','? '}';
+objectLiteral: '{' propertyNameAndValueList? '}';
```
StephenWeatherford: Don't think you want this change - this removes the ability to have trailing commas, right?

src/mongo/MongoScrapbook.ts
```diff
@@ -181,7 +181,10 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 				if (!lastCommand.arguments) {
```
StephenWeatherford: This code already existed, but I don't understand how lastCommand is guaranteed to exist.

StephenWeatherford: Let's create argumentObjects/arguments as [] when command visited.

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +207,46 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 	protected defaultResult(_node: ParseTree): MongoCommand[] {
 		return this.commands;
 	}
+
+	private parseContext(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext): Object {
```
StephenWeatherford: Better name.  argumentToObject?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +207,46 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+		let parsedObject: Object = {};
+		if (!ctx || ctx.childCount === 0) { //Base case and malformed statements
+			return parsedObject;
+		}
+		// In a well formed expression, Argument and propertyValue tokens should have exactly one child, from their definitions in mongo.g4
+		// The only difference in types of children between PropertyValue and argument tokens is the functionCallContext that isn't handled at the moment.
+		let child: ParseTree = ctx.children[0];
```
StephenWeatherford: I assume you'll need to add support for multiple children...?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +207,46 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+					const propertyValue = <mongoParser.PropertyValueContext>propertyAssignment.children[2];
+					parsedObject[stripQuotes(propertyName.text)] = this.parseContext(propertyValue);
+				}
+			}
+		}
+		else if (child instanceof mongoParser.ArrayLiteralContext) {
+			let elementList = <mongoParser.ElementListContext>child.children.find((context) => context instanceof mongoParser.ElementListContext);
+			let propertyValues = elementList.children.filter((node) => node instanceof mongoParser.PropertyValueContext);
+			parsedObject = propertyValues.map(this.parseContext.bind(this));
+		}
```
StephenWeatherford: else {
  console.assert("Unrecognized child type in parse tree");
}

src/mongo/MongoScrapbook.ts
PrashanthCorp: I tried to stay away from having argument in the name since it could be an argument or Property value.

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +207,46 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+
+	private parseContext(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext): Object {
+		let parsedObject: Object = {};
+		if (!ctx || ctx.childCount === 0) { //Base case and malformed statements
+			return parsedObject;
+		}
+		// In a well formed expression, Argument and propertyValue tokens should have exactly one child, from their definitions in mongo.g4
+		// The only difference in types of children between PropertyValue and argument tokens is the functionCallContext that isn't handled at the moment.
+		let child: ParseTree = ctx.children[0];
+		if (child instanceof mongoParser.LiteralContext) {
```
StephenWeatherford: I'd create some util funtions:
  filterType<T>(array, callback) -> filter objects which are instanceof T
  findType<T>(array, callback) { return filterType<T>(array, callback)[0]; }

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +207,46 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+			let unquoted = stripQuotes(child.text); // LiteralContext tokens encode numbers, booleans and undefined as "\"123\"", "\"true\"", "\"undefined\"", etc.
```
StephenWeatherford: I think you'll need to handle strings differently, right?  E.g. how handle true vs "true"?

src/mongo/grammar/mongoLexer.ts
```diff
@@ -114,7 +114,8 @@ export class mongoLexer extends Lexer {
 	private STRING_LITERAL_sempred(_localctx: RuleContext, predIndex: number): boolean {
 		switch (predIndex) {
 		case 0:
-			return !this.isExternalIdentifierText(this.text);
+			return !this.isExternalIdentifierText(this.text)
```
StephenWeatherford: whoops

src/mongo/MongoScrapbook.ts
PrashanthCorp: From my understanding, 
the token has 0 children if it is an empty string. 
1 child for each TerminalNode (a comma, a dot, a bracket/brace, etc.)
1 child for each know token it matches. 

I'll keep aside the case of 0 children moving forward (That's an empty string, handled a few lines above this code). 

For [an argument](https://github.com/Microsoft/vscode-cosmosdb/blob/f3641d4a88ebe8fe06120a9d2531e9e75e2e5633/grammar/mongo.g4#L34): If matched correctly, it will have only one child, which will be a literal, ObjectLiteral, or ArrayLiteral 
For [a propertyValue](https://github.com/Microsoft/vscode-cosmosdb/blob/f3641d4a88ebe8fe06120a9d2531e9e75e2e5633/grammar/mongo.g4#L47): If matched correctly, it should have only once child, which will be literal, ObjectLiteral, ArrayLiteral  or FunctionCall. 

The respective types would have the suffix of `Context` to the identifier on the g4.

src/mongo/MongoScrapbook.ts
PrashanthCorp: I'm finding implementing it to be more code than it's worth. Let's sync up tomorrow about this.

src/mongo/MongoScrapbook.ts
```diff
@@ -224,13 +225,13 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 			}
 		}
 		else if (child instanceof mongoParser.ObjectLiteralContext) {
-			let propertyNameAndValue = <mongoParser.PropertyNameAndValueListContext>child.children.find((node) => node instanceof mongoParser.PropertyNameAndValueListContext);
+			let propertyNameAndValue = randomUtils.findType(child.children, mongoParser.PropertyNameAndValueListContext);
```
StephenWeatherford: randomUtils is not "random utilities", but "utilities for randomness"

StephenWeatherford: Likely need to create a new util file.

src/mongo/MongoScrapbook.ts
StephenWeatherford: contextToObject?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
 	}
+
+	private parseContext(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext): Object {
+		let parsedObject: Object = {};
+		if (!ctx || ctx.childCount === 0) { //Base case and malformed statements
+			return parsedObject;
+		}
+		// In a well formed expression, Argument and propertyValue tokens should have exactly one child, from their definitions in mongo.g4
+		// The only difference in types of children between PropertyValue and argument tokens is the functionCallContext that isn't handled at the moment.
+		let child: ParseTree = ctx.children[0];
```
StephenWeatherford: When are you going to deal with function calls?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+			parsedObject = propertyValues.map(this.parseContext.bind(this));
+		} else {
+			console.assert("Unrecognized child type in parse tree.");
+		}
+		// tslint:disable-next-line:no-suspicious-comment
+		//TODO: else if(child instanceof mongoParser.FunctionCallContext)
+
+		return parsedObject;
+	}
+	private _parserNonStringTokenTypes = [mongoParser.mongoParser.NullLiteral, mongoParser.mongoParser.BooleanLiteral, mongoParser.mongoParser.NumericLiteral];
```
StephenWeatherford: This name doesn't adequately describe what this is (plenty of other non-string token types).  Personally I'd move this to the function where it's used and make it a const.

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+		// The only difference in types of children between PropertyValue and argument tokens is the functionCallContext that isn't handled at the moment.
+		let child: ParseTree = ctx.children[0];
+		if (child instanceof mongoParser.LiteralContext) {
+			let text = child.text;
+			let tokenType = child.start.type;
+			if (tokenType === mongoParser.mongoParser.StringLiteral) {
+				parsedObject = stripQuotes(text);
+			} else if (this._parserNonStringTokenTypes.indexOf(tokenType) > -1) {
+				parsedObject = JSON.parse(text);
+			}
```
StephenWeatherford: Throw otherwise?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+				for (let propertyAssignment of propertyAssignments) {
+					const propertyName = <mongoParser.PropertyNameContext>propertyAssignment.children[0];
+					const propertyValue = <mongoParser.PropertyValueContext>propertyAssignment.children[2];
+					parsedObject[stripQuotes(propertyName.text)] = this.parseContext(propertyValue);
+				}
+			}
+		}
+		else if (child instanceof mongoParser.ArrayLiteralContext) {
+			let elementList = randomUtils.findType(child.children, mongoParser.ElementListContext);
+			let propertyValues = randomUtils.filterType(elementList.children, mongoParser.PropertyValueContext);
```
StephenWeatherford: elementItems?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+			parsedObject = propertyValues.map(this.parseContext.bind(this));
+		} else {
+			console.assert("Unrecognized child type in parse tree.");
```
StephenWeatherford: This assert will always succeed (look up its definition).

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,51 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+		}
+		// tslint:disable-next-line:no-suspicious-comment
+		//TODO: else if(child instanceof mongoParser.FunctionCallContext)
```
StephenWeatherford: Otherwise throw?

src/mongo/MongoScrapbook.ts
PrashanthCorp: Does it look like `utils/array.ts` would be a good place?

StephenWeatherford:
Yes

src/mongo/MongoScrapbook.ts
PrashanthCorp: Does `nonStringLiteral` sound appropriate?

src/mongo/MongoScrapbook.ts
StephenWeatherford: yes

grammar/mongo.g4
```diff
@@ -73,7 +73,7 @@ BooleanLiteral: 'true' | 'false';
 NumericLiteral: '-'? DecimalLiteral;
 
 DecimalLiteral:
-	DecimalIntegerLiteral '.' DecimalDigit* ExponentPart?
+	DecimalIntegerLiteral '.' DecimalDigit+ ExponentPart?
```
StephenWeatherford: Why this change?

src/mongo/MongoScrapbook.ts
```diff
@@ -204,4 +206,53 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
...
+		if (child instanceof mongoParser.LiteralContext) {
+			let text = child.text;
+			let tokenType = child.start.type;
+			const nonStringLiterals = [mongoParser.mongoParser.NullLiteral, mongoParser.mongoParser.BooleanLiteral, mongoParser.mongoParser.NumericLiteral];
+			if (tokenType === mongoParser.mongoParser.StringLiteral) {
+				parsedObject = stripQuotes(text);
+			} else if (nonStringLiterals.indexOf(tokenType) > -1) {
+				parsedObject = JSON.parse(text);
+			} else {
+				throw new Error(`Can not identify token. Token text: ${text}`);
```
StephenWeatherford: hit: cannot is one word

PrashanthCorp:
I realized while testing that on the way to typing out 1.324, "1." would give an error. 1. is not json parsable. So this change enables the squigglies to point at an error instead of the error popup.
________________________________

grammar/mongo.g4
StephenWeatherford: > I realized while testing that on the way to typing out 1.324, "1." would give an error. 1. is not json parsable. So this change enables the squigglies to point at an error instead of the error popup.

okay

test/mongoGetCommand.test.ts
```diff
@@ -336,4 +336,49 @@ suite("scrapbook parsing Tests", () => {
         assert.deepEqual(err.range.start.line, 0);
         assert.deepEqual(err.range.start.character, 26);
     });
+    test("test function call with single quotes", () => {
```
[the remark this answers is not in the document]
StephenWeatherford: Also, have you played around with typing into a .mongo file yet, and see if you hit any exceptions or assertions?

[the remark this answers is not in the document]
PrashanthCorp: > Empty Object
Seems covered [here](https://github.com/Microsoft/vscode-cosmosdb/blob/9d72e66c28a76adf618e7e223c51597768de34a2/test/mongoGetCommand.test.ts#L127)

Have commits for the others mentioned. Just pushed them

grammar/mongo.g4
```diff
@@ -86,7 +86,7 @@ DB: 'db';
 // Don't declare LR/CRLF tokens - they'll interfere with matching against LineTerminator LF: '\n';
 // CRLF: '\r\n';
 
-STRING_LITERAL: ((~[",\\ \t\n\r:.;(){}\-]) | STRING_ESCAPE)+ {!this.isExternalIdentifierText(this.text)
+STRING_LITERAL: ((~["',\\ \t\n\r:.;(){}\-]) | STRING_ESCAPE)+ {!this.isExternalIdentifierText(this.text)
```
StephenWeatherford: This is questionable.

src/mongo/MongoScrapbook.ts
```diff
@@ -178,10 +181,9 @@ class MongoScriptDocumentVisitor extends MongoVisitor<MongoCommand[]> {
 			let functionCallContext = argumentsContext.parent;
 			if (functionCallContext && functionCallContext.parent instanceof mongoParser.CommandContext) {
 				const lastCommand = this.commands[this.commands.length - 1];
-				if (!lastCommand.arguments) {
-					lastCommand.arguments = [];
-				}
-				lastCommand.arguments.push(ctx.text);
+				const argAsObject = this.contextToObject(ctx);
+				lastCommand.argumentObjects.push(argAsObject);
+				lastCommand.arguments.push(JSON.stringify(argAsObject));
```
StephenWeatherford: I thought we were trying to avoid this?

test/mongoGetCommand.test.ts
```diff
@@ -319,21 +333,86 @@ suite("scrapbook parsing Tests", () => {
...
+        let command = getCommandFromText(text, new Position(0, 0));
+        assert.deepEqual(command.argumentObjects, [{ name: 1010101 }]);
+    });
+    test("test function call boolean", () => {
+        let text = `db.test1.insertMany({'name': false})`;
+        let command = getCommandFromText(text, new Position(0, 0));
+        assert.deepEqual(command.argumentObjects, [{ name: false }]);
+    });
+    test("test function call token inside quotes", () => {
+        let text = `db.test1.insertMany({'name': 'false'})`;
```
StephenWeatherford: How is this a function call token?

test/mongoGetCommand.test.ts
```diff
@@ -319,21 +333,86 @@ suite("scrapbook parsing Tests", () => {
...
+    /* This test will fail. Unquoted strings need to be fixed.
+    test("test function call with no quotes", () => {
+        let arg0 = `{name: 'First'}`;
+        let text = `db.test1.insertMany(${arg0})`;
+        let command = getCommandFromText(text, new Position(0, 0));
+        const argument = command.argumentObjects[0];
+        assert.deepEqual(argument, { name: "First" });
+    });
+    */
+
```
StephenWeatherford: Please test these strings:

"That's all"
'That"s all'
"Hello \"there\""
'Hello \'there\''

PrashanthCorp:
@StephenWeatherford 
A few users mentioned using aggregations. Hence I ran through the example mentioned in [this](https://docs.mongodb.com/manual/reference/operator/aggregation/match/) mongo docs page, mirroring #688 and #691 using single quotes in a few places.  
2. Aggregation works (For cosmos accounts, you need to set up the preview feature). 
1. Performed a find after the above test, using only single quotes. Worked for me. 

NB: Have a few commits I'm pushing now. 
Apart from tests, here's what they handle: 
1. Handling empty arrays in a parameter (ArrayLiteralContext tokens won't have children). This ensures users typing in a command don't see an error. 
2. I set the argument to be an empty array by default [here](https://github.com/Microsoft/vscode-cosmosdb/pull/685/files#diff-66772e7323fd96e096d49333cb563090R160). This causes a regression in collection views and find queries with no parameters, detailed in #715. Made a change there.

Issue #696, opened 2018-06-19
Connecting to Emulator running in Docker container doesn't parse properly


bfarmilo:
Related to  #324 

I can't seem to connect to an Emulator running in a Docker container (rather than localHost).

Using 'connect to Emulator' fails (since the emulator is not running on localhost).
Connecting via a connection string fails since the emulator connection string includes the key which contains '/' characters, causing the connection string parsing to fail.

I'm happy to try to help resolve this; it would be great if someone can point me in a direction though (either where I can specify a custom IP address for the emulator connection, probably the preferred solution, or properly parse the connectionstring to ignore '/' characters that are part of the key).

Apologies if not clear, will provide a few screen shots to illustrate.

1) Attach an Emulator (*127.0.0.1 not found, no surprise there as it's not running at that location)
![image](https://user-images.githubusercontent.com/15174544/41618814-3157aa1a-73d2-11e8-992f-61800870959f.png)

2) Connect using a connection string (* note key includes fwd-slashes)
![image](https://user-images.githubusercontent.com/15174544/41618883-64d45b90-73d2-11e8-9656-54fec062ba29.png)

3) Result after connecting with connection string (*a portion of the key parsed as the database, rather than the database I actually want to connect to, I assume this is why the 'attach emulator' feature was added -- to deal with this :)
![image](https://user-images.githubusercontent.com/15174544/41618975-9df0ea06-73d2-11e8-9a8b-e9fa8acdce7e.png)

PrashanthCorp:
Thanks for contributing! Really appreciate your help in fixing this! We didn't foresee this usecase. 

Just a reminder:  Is the address on the 2nd screenshot private information?

fiveisprime:
I'm confused, why can't you just forward the mongo container's port then connect to that? You shouldn't need to make any changes to the extension to connect to mongo in a container unless I'm missing something.

I typically do this for my own projects, start the container using `$ docker run --rm -d -p 27017:27017 mongo:latest` then connect to localhost:27017` as you normally would.

bfarmilo:
Hi  - I'm using a Docker Windows container, as far as I can tell the only way I can connect to the container is to use the IP address that's assigned when I start the container. If there's a way to do this without having to inspect the container each time, though, please let me know! My working solution feels like a nasty hack.

From: https://github.com/Azure/azure-cosmos-db-emulator-docker

> The response looks similar to the following:
>
```
Starting Emulator
[...]

PR #697, opened 2018-06-19
added config setting for IP

bfarmilo:
to be used when running emulator in docker container rather than localhost. See the issue I just reported #696


StephenWeatherford:
Closing since this is no longer needed (see solution in #696).  Thanks for be willing to help fix the issue, though.

Issue #687, opened 2018-06-15
update anything from scrapbook got timeout


lipt0n:
Repro steps:
open scrapbook
connect
update something
observe !

Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "use meteor"

Version: 0.7.2
OS: linux

example:
this works:
db.timesheets.find( {
        "workers._id":"iKSHSGcw2iCrNyjxf",
        "workers.days.year":"2018",
        "workers.days.month":"06"
      })

this got timeout every time

db.timesheets.update( {
        "workers._id":"iKSHSGcw2iCrNyjxf",
        "year":"2018",
        "month":"06"
      },{
        "$set":{
          "workers.$.days.0.something":"yupy!"
        }
      })

also
db.timesheets.update( {
        "year":"2018",
        "month":"06"
      },{
        "$set":{
          "workers.0.days.0.something":"yupy!"
        }
      })

got timeout...so probably any update ?

PrashanthCorp:
Hi Lipton, 
Thanks for trying out the extension and filing the issue! 


I ran the following two commands: 

```
db.timesheets.update( {
"workers._id":"iKSHSGcw2iCrNyjxf",
"year":"2018",
"month":"06"
},{
"$set":{
"workers.$.days.0.something":"yupy!"
}
});
```
This command gave me an error regarding the name of the field. It looks like having the $ in `workers.$.days.0.something` made it an invalid name. 


```
db.timesheets.update( {
"year":"2018",
"month":"06"
},{
"$set":{
"workers.0.days.0.something":"yupy!"
}
});
```
This command seemed to run without a timeout - I got back a result of how many documents were updated, etc. 


1. Please let me know if changing the field works. I couldn't find any links on $ being a valifd charcter for a field name. 
2. Please provide us a few more details to help me repro the last command.

lipt0n:
workers.$.days.0.something shouldn't be invalid, this is how You address element in array matching "workers._id":"iKSHSGcw2iCrNyjxf",

StephenWeatherford:
@lipt0n I'm guessing you're using the Cosmos emulator, right?  There's a known issue with some commands using the emulator (#508), which we hope to have fixed soon.

PrashanthCorp:
Hi Lipton, the merged fix was published today (version 0.8.0). Hope this fixes your issue. :-)
Let us know how the fix holds up with your use-case.

Issue #613, opened 2018-05-04
Timed out executing MongoDB command "db.version("


StephenWeatherford:
1) new scrapbook
2) type:

    db.version(

3) Execute
!) After a minute, you get a timeout error
EXPECTED: An actual error, immediately

StephenWeatherford:
Expecting: error message

Issue #508, opened 2018-04-09
Emulator: Running scrapbook commands utilizing the shell gives timeout


PrashanthCorp:
Before the function is executed, the shell creation runs a `use <dbname>` that times out in case of an emulator.

StephenWeatherford:
It's not just the emulator.  db.version() gives this, too

StephenWeatherford:
I take it back, I think it is only for emulator.  Somehow I think the emulator got connected as my db, and I can't repro anymore with a regular MongoDB connected.

StephenWeatherford:
The problem is here:

		try {
			this.mongoShell.stdin.write(script, 'utf8');
			this.mongoShell.stdin.write(os.EOL);
			this.mongoShell.stdin.write(executionId, 'utf8');
			this.mongoShell.stdin.write(os.EOL);
		} catch (error) {
			window.showErrorMessage(error.toString());
		}

PR #726, opened 2018-06-27
[Delayed] Improve error-handling of Mongo shell commands

StephenWeatherford:
Part of #508, #613, #687.  With this change, we now see the true error that's occurring on the process stdout:

> db.version()
> Error: FailedToParse: Bad digit "C" while parsing C2y6yDjf5
> try 'c:\Program Files\MongoDB\Server\3.6\bin\mongo.exe --help' for more information

### DIFF
--- a/src/mongo/shell.ts
+++ b/src/mongo/shell.ts
@@ -17,7 +17,7 @@ type CommandResult = {
 
 export class Shell {
 
-	private executionId: number = 0;
+	private executionId: number = 1234;
 	private disposables: IDisposable[] = [];
 
 	private onResult: EventEmitter<CommandResult> = new EventEmitter<CommandResult>();
@@ -57,27 +57,30 @@ export class Shell {
 			this.fireOnResult(error);
 		});
 		once(this.mongoShell, 'exit', (exitCode: number) => {
-			this.fireOnResult({ stringResult: "", exitCode });
+			console.error("outputdata:" + outputData);
+			console.error("stderr:" + stderrData);
+			this.fireOnResult({ stringResult: undefined, exitCode });
 		});
 
-		let buffers: string[] = [];
-		on(this.mongoShell.stdout, 'data', (b: Buffer) => {
-			let data: string = b.toString();
+		let outputData: string = "";
+		let stderrData: string = "";
+		on(this.mongoShell.stdout, 'data', (buffer: Buffer) => {
+			let data: string = buffer.toString();
 			const delimiter = `${this.executionId}${os.EOL}`;
 			if (data.endsWith(delimiter)) {
-				const result: string = buffers.join('') + data.substring(0, data.length - delimiter.length);
-				buffers = [];
+				const result: string = outputData + data.substring(0, data.length - delimiter.length);
+				outputData = "";
 				this.fireOnResult({ stringResult: result });
 			} else {
-				buffers.push(data);
+				outputData += data;
 			}
 		});
 
-		on(this.mongoShell.stderr, 'data', (stderrOutput: Buffer) => {
-			this.fireOnResult(new Error(stderrOutput.toString()));
+		on(this.mongoShell.stderr, 'data', (buffer: Buffer) => {
+			stderrData += buffer.toString();
 		});
-		once(this.mongoShell.stderr, 'close', () => {
-			// Do nothing
+		once(this.mongoShell.stderr, 'end', () => {
+			this.fireOnResult(new Error(stderrData));
 		});
 	}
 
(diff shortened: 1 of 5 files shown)


src/mongo/shell.ts
```diff
@@ -41,28 +52,33 @@ export class Shell {
...
 				buffers = [];
-				this.onResult.fire({
-					exitCode: void 0,
-					result,
-					stderr: void 0
-				});
+				this.fireOnResult({ stringResult: result });
 			} else {
-				buffers.push(b);
+				buffers.push(data); // asdf was buffer
```
EricJizbaMSFT: what's this comment mean?

src/mongo/shell.ts
```diff
@@ -72,43 +88,59 @@ export class Shell {
...
 			this.mongoShell.stdin.write(script, 'utf8');
 			this.mongoShell.stdin.write(os.EOL);
 			this.mongoShell.stdin.write(executionId, 'utf8');
 			this.mongoShell.stdin.write(os.EOL);
 		} catch (error) {
-			window.showErrorMessage(error.toString());
+			// Generally if writing to the process' stdin fails it has already exited
+			// with an error, and we will get notification via its stdout. So hold on to
+			// this and only return it if we don't see any other notifications after timeout.
+			writeError = error;
```
EricJizbaMSFT: Can you give an example of what this error would look like? I'm just confused why we would need to wait for the timeout to occur before throwing it.

src/mongo/shell.ts
```diff
@@ -5,23 +5,30 @@
 import * as cp from 'child_process';
 import * as os from 'os';
 import { IDisposable, toDisposable } from '../utils/vscodeUtils';
-import { EventEmitter, window } from 'vscode';
+import { EventEmitter } from 'vscode';
 import { ext } from '../extensionVariables';
 
+type CommandResult = {
```
StephenWeatherford: This is the crux of the fix...  Previously all errors were processed just like the actual data.  I've fixed all error code so that the processing can correctly tell what are errors and what are results.

src/mongo/shell.ts
```diff
@@ -72,43 +88,62 @@ export class Shell {
...
 				5000);
-			const disposable = this.onResult.event(result => {
+
+			// Handle results
+			const disposable = this.onResult.event((result: CommandResult) => {
+				// We will only process the first time onResult is fired
 				disposable.dispose();
 
-				if (result && result.code) {
+				if (result.stringResult) {
```
EricJizbaMSFT: So if 'stringResult' is defined - that means there are no errors?

src/mongo/shell.ts
```diff
@@ -41,28 +52,36 @@ export class Shell {
...
+				outputData += data;
 			}
 		});
 
-		on(this.mongoShell.stderr, 'data', result => this.onResult.fire(result));
-		once(this.mongoShell.stderr, 'close', result => this.onResult.fire(result));
+		on(this.mongoShell.stderr, 'data', (buffer: Buffer) => {
+			stderrData += buffer.toString();
+		});
+		once(this.mongoShell.stderr, 'end', () => {
```
PrashanthCorp: Looks like the event type was modified from `close` to `end`. What's the effect of this?

src/mongo/shell.ts
```diff
@@ -5,23 +5,30 @@
...
+	stringResult?: string;
+	exitCode?: number;
+	code?: string;
+	message?: string;
+};
+
 export class Shell {
 
-	private executionId: number = 0;
+	private executionId: number = 1234;
```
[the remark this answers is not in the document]
PrashanthCorp: Seeing where it's used, I don't quite know how to check if it's unique.

src/mongo/shell.ts
```diff
@@ -72,43 +91,64 @@ export class Shell {
...
 				5000);
-			const disposable = this.onResult.event(result => {
+
+			// Handle results
+			const disposable = this.onResult.event((result: CommandResult) => {
+				// We will only process the first time onResult is fired
 				disposable.dispose();
 
-				if (result && result.code) {
+				if (result.stringResult) {
```
PrashanthCorp: I'm unable to follow when `result.code` (the original line) would be defined - I don't see any mention of an object having the `code` field defined.

src/mongo/shell.ts
```diff
@@ -72,43 +91,64 @@ export class Shell {
...
+					let lines = result.stringResult.split(os.EOL).filter(line => !!line && line !== 'Type "it" for more');
+					lines = lines[lines.length - 1] === 'Type "it" for more' ? lines.splice(lines.length - 1, 1) : lines;
```
PrashanthCorp: Given that you filter out `Type "it" for more` one line above, in what scenario would the true branch execute here?

src/mongo/shell.ts
```diff
@@ -72,43 +91,64 @@ export class Shell {
...
+					executed = true;
+					let text = lines.join(os.EOL);
+					resolve(text);
+				} else {
 					if (result.code === 'ENOENT') {
 						result.message = `Could not find Mongo shell. Make sure it is on your path or you have set the '${ext.settingsKeys.mongoShellPath}' VS Code setting to point to the Mongo shell executable file. Attempted command: "${this.execPath}"`;
+					} else if (result.exitCode) {
+						result.message = result.message || `Mongo shell exited with code ${result.exitCode}`;
```
PrashanthCorp: I'm trying to understand the use of || here. Please point out any errors in my understanding: 
1. The exitCode would be a defaultvalue, say 0, if the operation is successful. In this case, `result.message` is guaranteed to exist.
2. On any other exit code, we know that result.message doesn't exist, hence we output the exitCode.

I don't see where result.message is being defined, hence the perplexity on my end.

StephenWeatherford:
This class probably needs to be completely rewritten...

PR #727, opened 2018-06-28
Handle passwords with slashes in connection strings

### LINKED ISSUE #696: Connecting to Emulator running in Docker container doesn't parse properly

StephenWeatherford:
Fixes #696

### DIFF
--- a/src/mongo/mongoConnectionStrings.ts
+++ b/src/mongo/mongoConnectionStrings.ts
@@ -3,6 +3,18 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+//let matches = connectionString.match(/[a-zA-Z]+:\/\/(?:[^:]+:[^@]+@)?[^/]+(\/([^/?]+))/);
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
Closing in favor of #731

PR #733, opened 2018-06-29
Perform null check before assigning functionCallContext name.

### LINKED ISSUE #689: Error while typing out a command in between two legal commands
Repro: 
1. Have these commands typed out : 
a) `db.heroes.find({"address.City": "Goooooothammmmmmmmm"
});`
b)`db.heroes.drop();`

2. Start typing a command in between these commands. 
a)`db` -> Fine
b)`db.heroes` -> Fine
c)`db.herores.` -> Error before intellisense: "TypeError: Cannot read property 'text' of undefined"

Error stems from visitFunctionCall that tries to look up `ctx._FUNCTION_NAME.text` of the string "" (the empty string after `db.heroes`).

PrashanthCorp:
Fixes #689

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -201,7 +201,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 	}
 
 	visitFunctionCall(ctx: mongoParser.FunctionCallContext): MongoCommand[] {
-		if (ctx.parent instanceof mongoParser.CommandContext) {
+		if (ctx.parent instanceof mongoParser.CommandContext && ctx._FUNCTION_NAME) {
 			this.commands[this.commands.length - 1].name = (ctx._FUNCTION_NAME && ctx._FUNCTION_NAME.text) || "";
 		}
 		return super.visitFunctionCall(ctx);
--- a/test/mongoGetCommand.test.ts
+++ b/test/mongoGetCommand.test.ts
@@ -696,13 +696,12 @@ suite("scrapbook parsing Tests", () => {
     });
 
 
-    //This test will fail. See https://github.com/Microsoft/vscode-cosmosdb/issues/689
-    // test("test incomplete function call - replicate user typing - no function call yet", () => {
-    //     let text = `db.test1.`;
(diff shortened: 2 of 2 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -201,7 +201,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 	}
 
 	visitFunctionCall(ctx: mongoParser.FunctionCallContext): MongoCommand[] {
-		if (ctx.parent instanceof mongoParser.CommandContext) {
+		if (ctx.parent instanceof mongoParser.CommandContext && ctx._FUNCTION_NAME) {
 			this.commands[this.commands.length - 1].name = (ctx._FUNCTION_NAME && ctx._FUNCTION_NAME.text) || "";
```
nturinski: nit: I don't think you need to check ctx._FUNCTION_NAME on line 205 if you're doing it on 204.

nturinski: What does `(ctx._FUNCTION_NAME && ctx._FUNCTION_NAME.text)` resolve to anyway?  Are those 2 strings?

src/mongo/MongoScrapbook.ts
```diff
@@ -201,7 +201,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 	}
 
 	visitFunctionCall(ctx: mongoParser.FunctionCallContext): MongoCommand[] {
-		if (ctx.parent instanceof mongoParser.CommandContext) {
+		if (ctx.parent instanceof mongoParser.CommandContext && ctx._FUNCTION_NAME) {
```
StephenWeatherford: Not sure we should be avoiding creating a command here.  Need to play around with it.

src/mongo/MongoScrapbook.ts
PrashanthCorp: `ctx._FUNCTION_NAME` is a Token that the parser identifies. 
`ctx._FUNCTION_NAME.text` is the string the token points to. 

Should resolve to a string, else a false-y value.

PrashanthCorp:
I'm closing this. 

It looks like #660  fixes the issue anyway. 
The corresponding test was not present in the PR.

PR #748, opened 2018-07-03
Add user agent to docdb and mongo calls

StephenWeatherford:
I had to revert this PR yesterday due to a dependency issue - mongo language server can't load vscode and therefore can't depend on vscode-azureextensionui.

The first commit in this PR is the same as yesterday, please review the second one, which is the fix.  Thx.


src/mongo/services/languageService.ts
```diff
@@ -3,6 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// NOTE: This file may not take a dependencey on vscode or anything that takes a dependency on it (such as vscode-azureextensionui)
```
PrashanthCorp: dependency

PR #758, opened 2018-07-09
Implement some more tslint rules for 0.9.0

PrashanthCorp:
Enforce:  curly, eofline, no-single-line-block-comment, (no) trailing comma, no-console, switch-default, class-name

Ignore: no-use-before-declare, func-name, export-name, interface-name

### DIFF
--- a/src/graph/GraphViewServer.ts
+++ b/src/graph/GraphViewServer.ts
@@ -4,15 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { EventEmitter } from 'events';
-import * as http from 'http';
-import * as vscode from 'vscode';
-import * as io from 'socket.io';
-import { GraphConfiguration } from './GraphConfiguration';
 import * as gremlin from "gremlin";
+import * as http from 'http';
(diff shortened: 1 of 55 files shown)


src/graph/GraphViewServer.ts
```diff
@@ -97,6 +97,7 @@ export class GraphViewServer extends EventEmitter {
         0, // dynamnically pick an unused port
         () => {
           this._port = this._httpServer.address().port;
+          //tslint:disable-next-line:no-console
```
StephenWeatherford: We should probably just disable this rule

src/mongo/services/languageService.ts
```diff
@@ -33,7 +33,7 @@ export class LanguageService {
 			return {
 				capabilities: {
 					textDocumentSync: this.textDocuments.syncKind, // Tell the client that the server works in FULL text document sync mode
-					completionProvider: { triggerCharacters: ['.'] },
```
StephenWeatherford: I've got mixed feelings about this rule.   Either way is fine with me.

PR #761, opened 2018-07-10
Add regex support for find commands in the scrapbook

PrashanthCorp:
- Add capability to the parser to recognize regular expressions, and basic error handling (wrong flags)
- Modify contextToObject to recognize and handle Regex tokens
- Use argumentObjects instead of arguments in the find operation


PrashanthCorp:
There are a few more considerations w.r.t EJSON compatibility. Moreover, EJSON already supports this well, [as detailed here](https://docs.mongodb.com/manual/reference/mongodb-extended-json/#regular-expression).

grammar/mongo.g4
```diff
@@ -72,6 +73,10 @@ BooleanLiteral: 'true' | 'false';
 
 NumericLiteral: '-'? DecimalLiteral;
 
+RegexLiteral: '/' (~[/])+ '/' (RegexFlag)*;
```
StephenWeatherford: what about escaped '/'s?

src/mongo/MongoScrapbook.ts
```diff
@@ -240,6 +240,11 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 			const nonStringLiterals = [mongoParser.mongoParser.NullLiteral, mongoParser.mongoParser.BooleanLiteral, mongoParser.mongoParser.NumericLiteral];
 			if (tokenType === mongoParser.mongoParser.StringLiteral) {
 				parsedObject = stripQuotes(text);
+			} else if (tokenType === mongoParser.mongoParser.RegexLiteral) {
+				let params: string[] = child.text.split('/');
```
StephenWeatherford: won't work for escaped '/'s

grammar/mongo.g4
StephenWeatherford: both \/ and [*/*]

src/mongo/MongoScrapbook.ts
StephenWeatherford: You should have the parser do this processing for you...  Have a specific token or parse element for just the portion inside the slashes.

test/mongoGetCommand.test.ts
```diff
@@ -640,6 +640,30 @@ suite("scrapbook parsing Tests", () => {
         }
     });
 
+    //Examples inspired from https://docs.mongodb.com/manual/reference/operator/query/regex/
```
StephenWeatherford: You don't have any interesting regular expressions here. e.g. using ^()[]$+* etc.  Even just one with something like:

  /^(hello)*[^0-9]+|(there)$/

Notice you also need escaped '/' tests.

StephenWeatherford:
Updates?

test/mongoGetCommand.test.ts
```diff
@@ -640,6 +640,30 @@ suite("scrapbook parsing Tests", () => {
...
+        assert.deepEqual(generatedRegExp.source, "789$");
+    });
+
+    test("test regular expressions - Intellisense - flag contains invalid option", () => {
+        let text = `db.test1.beep.find({ sku: { $regex: /789$/q } })`;
+        let command = getCommandFromTextAtLocation(text, new Position(0, 0));
+        assert.deepEqual(command.errors[0].message, "mismatched input '}' expecting '('");
+        assert.deepEqual(command.errors[0].range.start.character, 44);
+    });
+
```
StephenWeatherford: Let's see some more unusual regexes then I'll approve.

PR #763, opened 2018-07-10
Replace contents of cosmos-result.json with blank text after opening.

### LINKED ISSUE #676: When we open a cosmos-results.json, we display previous contents before updating with current results
If there's an error getting those results, then the previous results are what the user sees.

E.g.:
db.items.find({ "_id": { "$oid": "5b0ed0dd03baa16259000020" } })
 with 0.7.1 will display previous file contents then give error:

unknown operator: $oid

PrashanthCorp:
Fixes #676

### DIFF
--- a/src/CosmosEditorManager.ts
+++ b/src/CosmosEditorManager.ts
@@ -76,6 +76,7 @@ export class CosmosEditorManager {
         this._globalState.update(this._persistedEditorsKey, fileMapLabels);
 
         const textEditor = await vscode.window.showTextDocument(document, column, preserveFocus);
+        await vscodeUtils.writeToEditor(textEditor, "");
         const data = await editor.getData();
         await this.updateEditor(data, textEditor, editor);
     }
--- a/src/CosmosEditorManager.ts
+++ b/src/CosmosEditorManager.ts
@@ -75,9 +75,8 @@ export class CosmosEditorManager {
         Object.keys(this.fileMap).forEach((key) => fileMapLabels[key] = (this.fileMap[key]).id);
         this._globalState.update(this._persistedEditorsKey, fileMapLabels);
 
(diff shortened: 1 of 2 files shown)


src/CosmosEditorManager.ts
```diff
@@ -76,6 +76,7 @@ export class CosmosEditorManager {
         this._globalState.update(this._persistedEditorsKey, fileMapLabels);
 
         const textEditor = await vscode.window.showTextDocument(document, column, preserveFocus);
+        await vscodeUtils.writeToEditor(textEditor, "");
```
EricJizbaMSFT: Shouldn't this use `this.updateEditor`?

PrashanthCorp: `this.updateEditor` also tries to convert the passed data to a string using each editor's method. I feel that's unnecessary in this case - we only want to clear the contents of what's printed (recovered by VSCode).

EricJizbaMSFT: I'm more worried about the logic that ignores the save. I don't want to accidentally update someones document in Azure to an empty string

PrashanthCorp: FWIW, I tested the following: 
1. Run a `find` query
2. Delete a character in result.json to make it dirty.
3. Run another find query. 

I get a warning regarding the overwrite, courtesy [this line](https://github.com/Microsoft/vscode-cosmosdb/blob/7b49e1df853098632ec4d174af29f38faf27ef57/src/CosmosEditorManager.ts#L67). 

I don't feel strongly about this, I can use `this.updateEditor` if required.

StephenWeatherford: As discussed, if there query throws with an error, you're left with a blank, unsaved document.  The important part is to not end up bringing the document forward if it's going to be displaying old data, so *not* bringing it up in the case of an error is fine.

PR #765, opened 2018-07-12
executeCommands: catch and rethrow error with relevant command.

### LINKED ISSUE #606: Mongo command should be included in error message during "executeAll"
I got the below error message and it took me a little bit to figure out which command was throwing the error. I would prefer a message that included the db and command name (although not the full command with arguments, because that would be too much). For example: `Error while running "db.test.insertMany": E11000 duplicate key error collection: test.scores Failed _id or unique key constraint`

Action: cosmosDB.executeAllMongoCommands
Error type: 11000
Error Message: E11000 duplicate key error collection: test.scores Failed _id or unique key constraint

Version: 0.6.1-alpha
OS: darwin

PrashanthCorp:
Fixes #606

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -8,7 +8,7 @@ import { CommonTokenStream } from 'antlr4ts/CommonTokenStream';
 import { ErrorNode } from 'antlr4ts/tree/ErrorNode';
 import { ParseTree } from 'antlr4ts/tree/ParseTree';
 import * as vscode from 'vscode';
-import { AzureTreeDataProvider, IActionContext, IAzureParentNode } from 'vscode-azureextensionui';
+import { AzureTreeDataProvider, IActionContext, IAzureParentNode, parseError } from 'vscode-azureextensionui';
 import { CosmosEditorManager } from '../CosmosEditorManager';
 import { ext } from '../extensionVariables';
 import { filterType, findType } from '../utils/array';
@@ -77,8 +77,9 @@ async function executeCommands(activeEditor: vscode.TextEditor, database: IAzure
 		try {
 			await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
 		} catch (e) {
-			e.message = `${command.text.split('(')[0]}, line ${command.range.start.line + 1}, character ${command.range.start.character + 1}: ${e.message}`;
-			throw e;
+			const err = parseError(e);
+			err.message = `${command.text.split('(')[0]}, ${command.range.start.line + 1}:${command.range.start.character + 1} - ${err.message}`;
+			throw err;
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -74,7 +74,12 @@ export function getAllCommandsFromTextDocument(document: vscode.TextDocument): M
 
 async function executeCommands(activeEditor: vscode.TextEditor, database: IAzureParentNode<MongoDatabaseTreeItem>, extensionPath, editorManager: CosmosEditorManager, tree: AzureTreeDataProvider, context: IActionContext, commands: MongoCommand[]): Promise<void> {
 	for (let command of commands) {
-		await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+		try {
+			await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
+		} catch (e) {
```
StephenWeatherford: In JavaScript anything can be thown, so e.message will not work.  use parseError

src/mongo/MongoScrapbook.ts
```diff
@@ -74,7 +74,12 @@ export function getAllCommandsFromTextDocument(document: vscode.TextDocument): M
...
+			e.message = `${command.text.split('(')[0]}, line ${command.range.start.line + 1}, character ${command.range.start.character + 1}: ${e.message}`;
```
StephenWeatherford: The reason for using split is unclear (and very suspect).

StephenWeatherford: Show an example output please.

StephenWeatherford: Don't need to be so specific with "line" and "column" - I'd suggest xx:yy format like compilers do.  But you should review with Matt.

PrashanthCorp: It arose out of `command.collection` not always being defined. I wanted the command prefix that would hold for both of : 
1. `db.collection1.insertMany`
2. `db.getCollectionInfos()`

It seems like printing the substring till the first open bracket is a simple way to extract a well-represented prefix for both cases. 

I myself don't find it cleanest or the most obvious, I'm open to suggestions.

PrashanthCorp: Here: 

db.testcase0.insertMany, line 0, character 0: E11000 duplicate key error collection: trial00.testcase0 Failed _id or unique key constraint

src/mongo/MongoScrapbook.ts
PrashanthCorp: Thanks for the catch!

src/mongo/MongoScrapbook.ts
```diff
@@ -77,8 +77,9 @@ async function executeCommands(activeEditor: vscode.TextEditor, database: IAzure
 		try {
 			await executeCommand(activeEditor, database, extensionPath, editorManager, tree, context, command);
 		} catch (e) {
-			e.message = `${command.text.split('(')[0]}, line ${command.range.start.line + 1}, character ${command.range.start.character + 1}: ${e.message}`;
-			throw e;
+			const err = parseError(e);
```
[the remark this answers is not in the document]
PrashanthCorp: This is the output now thrownm with parseError :
db.testcase0.insertMany, 1:1 - E11000 duplicate key error collection: trial00.testcase0 Failed _id or unique key constraint

parseError throws an `IParseError`, guaranteed to have a message field, which is what's displayed when we throw an error. Am I missing something?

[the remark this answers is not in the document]
StephenWeatherford: But it's not an error.  new Error(parseError(e).message)

StephenWeatherford:
image

PR #772, opened 2018-07-20
Mongo Document treeItem: show "null" if _id is null

PrashanthCorp:
vscode has a problem with treeItems having no id field. 

It is possible to insert a document with the _id field as _null_. The following commands executes successfully. 
```
db.coll.insertOne({_id: null})
```
(_id can have any data type except for arrays)

This PR handles this edge-case on our end.

### DIFF
--- a/src/mongo/tree/MongoDocumentTreeItem.ts
+++ b/src/mongo/tree/MongoDocumentTreeItem.ts
@@ -34,7 +34,7 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
     }
 
     get id(): string {
-        return this.document._id.toString();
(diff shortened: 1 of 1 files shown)


src/mongo/tree/MongoDocumentTreeItem.ts
```diff
@@ -34,7 +34,7 @@ export class MongoDocumentTreeItem implements IAzureTreeItem {
     }
 
     get id(): string {
-        return this.document._id.toString();
+        return (this.document._id && this.document._id.toString()) || "null";
```
StephenWeatherford: return String(this.document._id);

Issue #614, opened 2018-05-04
Error Message: m.text.split is not a function


StephenWeatherford:
Repro steps:
1) in scrapbook, run:

   db.c1.insert()

Action: cosmosDB.executeMongoCommand
Error type: Error
Error Message: m.text.split is not a function

Version: 0.7.0
OS: win32

PrashanthCorp:
Insert requires a parameter (the documents to be inserted) to run correctly. Following the stack trace, the error is thrown at [this line in the library](https://github.com/mongodb/node-mongodb-native/blob/985485069e41885d49d0e44617d81738b1f53348/lib/collection.js#L504) which we catch and resolve as an error [here](https://github.com/Microsoft/vscode-cosmosdb/blob/bb17955e18ed0f4405232908afcded0a1dfaba35/src/mongo/tree/MongoCollectionTreeItem.ts#L144). 

I feel an error is being thrown as intended, but it's non-informative. If you feel the error message is not informative enough, that can be fixed,

StephenWeatherford:
The original exception is "TypeError: Cannot read property '_id' of undefined".  Sometimes it contains a bit better info.  But it's not getting handled correctly, and a new exception occurs here:

export async function writeToEditor(editor: vscode.TextEditor, data: string): Promise<void> {
    await editor.edit((editBuilder: vscode.TextEditorEdit) => {

So the first issue is that the second exception should not be happening, we should be showing the first.

EricJizbaMSFT:
Re-opening this as it doesn't seem truly fixed to me. It feels like one confusing error message was switched for a different confusing error message ("_id" isn't even a required parameter when inserting something). I would expect an error like "insert requires at least one parameter"

PrashanthCorp:
Fixed via #780. Closing this. 

Current error for reference: 
![image](https://user-images.githubusercontent.com/30420973/45970458-6525dc80-bfeb-11e8-89af-66bf9bafa129.png)

PR #778, opened 2018-07-27
Fix symptom of #614. Throw error if user calls insert with no argument

PrashanthCorp:


PrashanthCorp:
Needs some refactoring. Closing

PR #780, opened 2018-08-01
Refactor executeCommand to pre-emptively throw errors for parameter mismatch

PrashanthCorp:


src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -171,84 +167,75 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
 
 	//tslint:disable:no-any
-	private insertMany(args: any[]): Thenable<string> {
-		// documents = args[0], collectionWriteOptions from args[1]
+	private insertMany(documents: any[], options?: any): Thenable<string> {
 		let insertManyOptions: CollectionInsertManyOptions = {};
 		const docsLink: string = "Please see mongo shell documentation. https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany.";
-		if (!args || args.length === 0) {
+		if (!documents) { // this code should not be hit
 			return Promise.reject(new Error("Too few arguments passed to insertMany. " + docsLink));
```
PrashanthCorp: Using an assert here throws an error inside a promise. 
Is there a way to return a rejected promise in case an assertion fails?

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -25,6 +25,7 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 	private _cursor: Cursor | undefined;
 	private _hasMoreChildren: boolean = true;
 	private _batchSize: number = DefaultBatchSize;
+	private readonly deferToShell = null;
```
StephenWeatherford: I'm finding this more confusing than simply commenting on the return values using `/**    @returns` etc.  Plus you don't seem to have used it.

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -101,46 +102,41 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
-				}
-				if (name === 'insertOne') {
-					return reportProgress(this.insertOne(argument ? parseJSContent(argument) : undefined), 'Inserting document');
-				}
-				if (name === 'deleteOne') {
-					return reportProgress(this.deleteOne(argument ? parseJSContent(argument) : undefined), 'Deleting document');
-				}
-				if (name === 'deleteMany') {
-					return reportProgress(this.deleteMany(argument ? parseJSContent(argument) : undefined), 'Deleting documents');
+			// tslint:disable-next-line:no-any
```
StephenWeatherford: Are the arguments truly any?  I thought they were all objects at least...?

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -101,46 +102,41 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
+			type MongoFunction = (...args: any[]) => Thenable<string>;
+			let functions: { [functionName: string]: [MongoFunction, string, number, number, number] } = {
+				// format: command name (from the argument) : corresponding function call[0],
+				// text to show during the operation[1], min #args[2], #max args[3], #args this function handles[4]
```
StephenWeatherford: easier to read as:
```
0: corresponding function call
1: min #args
```
etc


Also explain what "#args this function handles" means, it's not very clear

src/mongo/tree/MongoCollectionTreeItem.ts
StephenWeatherford: Make it a const (may need to pull out of class) and add a comment as to what it means

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -101,46 +102,41 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
...
+			};
+
+			if (name in functions) {
+				let functionData = functions[name];
+				if (parameters.length < functionData[2]) { //has less than the min allowed
+					return Promise.reject(new Error(`Too few arguments passed to command ${name}.`));
 				}
-				if (name === 'remove') {
-					return reportProgress(this.remove(argument ? parseJSContent(argument) : undefined), 'Removing');
+				if (parameters.length > functionData[3]) { //has more than the max allowed
```
StephenWeatherford: these types of things are always easier to read (and self-documenting) if at the beginning of the block you do this:
```
  let funcCall = functionData[0];
  let whatever = functionData[1];
  let whatever = functionData[2];
  let maxParams = functionData[3]
```
etc

PrashanthCorp:
Ping.

PR #782, opened 2018-08-03
Update CosmosDBAccountProvider.ts

yujhongmicrosoft:
Hello, the next release for the azure account extension has sovereign support, and to work with those changes this extension needs to include the base url when creating the cosmosdbManagementClient.
I have gone through testing; please let me know what needs to be done to get these changes merged in.

Thank you!

### DIFF
--- a/src/tree/CosmosDBAccountProvider.ts
+++ b/src/tree/CosmosDBAccountProvider.ts
@@ -7,7 +7,6 @@ import { CosmosDBManagementClient } from 'azure-arm-cosmosdb';
 import { DatabaseAccount, DatabaseAccountListKeysResult, DatabaseAccountsListResult } from 'azure-arm-cosmosdb/lib/models';
 import * as vscode from 'vscode';
 import { AzureWizard, IActionContext, IAzureNode, IAzureTreeItem, IChildProvider, LocationListStep, parseError, ResourceGroupListStep } from 'vscode-azureextensionui';
-import { getCosmosDBManagementClient } from '../docdb/getCosmosDBManagementClient';
 import { DocDBAccountTreeItem } from "../docdb/tree/DocDBAccountTreeItem";
 import { API, getExperience } from '../experiences';
 import { TryGetGremlinEndpointFromAzure } from '../graph/gremlinEndpoints';
@@ -28,7 +27,7 @@ export class CosmosDBAccountProvider implements IChildProvider {
     }
 
     public async loadMoreChildren(node: IAzureNode): Promise<IAzureTreeItem[]> {
-        const client = getCosmosDBManagementClient(node.credentials, node.subscriptionId);
+        const client = new CosmosDBManagementClient(node.credentials, node.subscriptionId, node.environment.resourceManagerEndpointUrl);
         const accounts: DatabaseAccountsListResult = await client.databaseAccounts.list();
         let accountTreeItems = [];
         await Promise.all(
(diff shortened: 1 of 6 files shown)


yujhongmicrosoft:
Hello, the next release for the azure account extension has  sovereign support, and to work with those changes this extension needs  to include the base url when creating the **cosmosDBManagementClient.** I have gone through testing; please let me know what needs to be done to get these changes merged in. Thank you!
--

src/tree/CosmosDBAccountProvider.ts
```diff
@@ -28,7 +27,7 @@ export class CosmosDBAccountProvider implements IChildProvider {
     }
 
     public async loadMoreChildren(node: IAzureNode): Promise<IAzureTreeItem[]> {
-        const client = getCosmosDBManagementClient(node.credentials, node.subscriptionId);
+        const client = new CosmosDBManagementClient(node.credentials, node.subscriptionId, node.environment.resourceManagerEndpointUrl);
```
PrashanthCorp: Hi, here are a couple of things I have noticed: 
1. The change effectively removes the userAgent. The userAgent helps _Azure_ collect the telemetry, not us. The intent of this is not very clear.  
2. This change affects _all clients created_, not just sovereign accounts. 

I would suggest  the following: 
```
export function getCosmosDBManagementClient(credentials: ServiceClientCredentials, subscriptionId: string, baseUri: string, isSovereign: boolean): CosmosDBManagementClient {  
//difference in function definition - added baseUri and a boolean flag. 
}
----

public loadMoreChildren(....){
//logic to check if it is a sovereign account
getCosmosDBManagementClient(node.credentials, node.subscriptionId, , node.environment.resourceManagerEndpointUrl, account.isSovereign);
....
}
```

StephenWeatherford: @yujhongmicrosoft It looks like from https://github.com/Microsoft/vscode-azureappservice/pull/554/files that you simply intended to add an extra argument to the new CosmosDBManagementClient call. Please add this argument to getCosmosDBManagementClient instead of replacing it with a call to `new CosmosDBManagementClient ()`

PrashanthCorp:
Looks good to me. I'll wait for @StephenWeatherford's thoughts before merging.

src/tree/CosmosDBAccountProvider.ts
```diff
@@ -28,7 +28,7 @@ export class CosmosDBAccountProvider implements IChildProvider {
     }
 
     public async loadMoreChildren(node: IAzureNode): Promise<IAzureTreeItem[]> {
-        const client = getCosmosDBManagementClient(node.credentials, node.subscriptionId);
+        const client = getCosmosDBManagementClient(node.credentials, node.subscriptionId, node.environment.resourceManagerEndpointUrl);
```
StephenWeatherford: Why do we only pass this in sometimes?

StephenWeatherford: If it should be passed in always, please change getCosmosDBManagementClient to require the argument.

yujhongmicrosoft: @StephenWeatherford  any updates?

PrashanthCorp:
There seems to be a build error. 

It looks like you're passing the wrong number of arguments into `getCosmosDBManagementClient` in one of the calls.

src/tree/CosmosDBAccountProvider.ts
StephenWeatherford: I've made some fixes.  Haven't been able to actually test it against a sovereign cloud, but theoretically it should be fine to check in now.  Have you done any testing?

yujhongmicrosoft: @StephenWeatherford yes I had done some testing before, but I realized I did not commit that last change that you fixed 👍 @PrashanthCorp I have tested the current branch and everything works with sovereigns. Let me know if there's anything else you need.

yujhongmicrosoft:
@PrashanthCorp  any other changes needed?

PR #800, opened 2018-08-28
showExplorer on cmd palette: Resolve collectionTreeItem to GraphTreeItem

### LINKED ISSUE #799: Error when trying to open GraphExplorer
Repro steps:
<Enter steps to reproduce issue>
I connected to my subscription and then went through selection boxes to select the graph database and the collection. In the end, I got below error. The query interface didn't open.

Action: cosmosDB.openGraphExplorer
Error type: TypeError
Error Message: node.treeItem.showExplorer is not a function

Version: 0.8.0
OS: win32

PrashanthCorp:
Fixes #799

### DIFF
--- a/src/graph/registerGraphCommands.ts
+++ b/src/graph/registerGraphCommands.ts
@@ -39,9 +39,15 @@ export function registerGraphCommands(context: vscode.ExtensionContext, tree: Az
         await node.deleteNode();
     });
     registerCommand('cosmosDB.openGraphExplorer', async (node: IAzureNode<GraphTreeItem>) => {
-        if (!node) {
-            node = <IAzureNode<GraphTreeItem>>await tree.showNodePicker(GraphCollectionTreeItem.contextValue);
+        let treeItem: GraphTreeItem;
+        if (node) {
+            treeItem = node.treeItem;
         }
-        await node.treeItem.showExplorer(graphViewsManager);
+        if (!node) {
+            let collectionNode = <IAzureNode<GraphCollectionTreeItem>>await tree.showNodePicker(GraphCollectionTreeItem.contextValue);
+            let children = await collectionNode.treeItem.loadMoreChildren(collectionNode, false);
+            treeItem = <GraphTreeItem>(children[0]);
+        }
+        await treeItem.showExplorer(graphViewsManager);
     });
 }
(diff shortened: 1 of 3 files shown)


src/graph/registerGraphCommands.ts
```diff
@@ -39,9 +39,15 @@ export function registerGraphCommands(context: vscode.ExtensionContext, tree: Az
         await node.deleteNode();
     });
     registerCommand('cosmosDB.openGraphExplorer', async (node: IAzureNode<GraphTreeItem>) => {
+        let treeItem: GraphTreeItem;
+        if (node) {
+            treeItem = node.treeItem;
+        }
         if (!node) {
```
StephenWeatherford: else if

src/graph/registerGraphCommands.ts
```diff
@@ -39,9 +39,15 @@ export function registerGraphCommands(context: vscode.ExtensionContext, tree: Az
...
-            node = <IAzureNode<GraphTreeItem>>await tree.showNodePicker(GraphCollectionTreeItem.contextValue);
+            let collectionNode = <IAzureNode<GraphCollectionTreeItem>>await tree.showNodePicker(GraphCollectionTreeItem.contextValue);
+            let children = await collectionNode.treeItem.loadMoreChildren(collectionNode, false);
+            treeItem = <GraphTreeItem>(children[0]);
```
StephenWeatherford: I believe the right way to do this is:

1) pass in GraphTreeItem to showNodePicker
2) Add it to GraphAccountTreeItem.isAncestorOf
3) Implement GraphCollectionTreeItem.pickTreeItem\

It's possible this might not work because of the fact that we don't want to show the final leaf node, but I think it will work.

src/graph/tree/GraphCollectionTreeItem.ts
```diff
@@ -75,4 +78,21 @@ export class GraphCollectionTreeItem implements IAzureTreeItem {
             throw new UserCancelledError();
         }
     }
+
+    public pickTreeItem?(expectedContextValue: string): IAzureTreeItem | undefined {
+        switch (expectedContextValue) {
+            case DocDBDocumentsTreeItem.contextValue:
```
StephenWeatherford: Why are DocDBDocument{,s}TreeItem here?

PR #828, opened 2018-09-17
fix #813

### LINKED ISSUE #813: [Suggestion]Keep consistent when adding a collection between VS Code and portal
Repro steps:
1. Right click on a SQL account -> Select 'Open In Portal' -> Select 'Add Collection'.
2. Drill down from the same SQL account in Step1 ->  Create a new database -> Right click on this new database -> Select 'Create Collection'.

Actual:
In VS Code, it prompts user to enter a collection name for this new collection.
In portal, it prompts user to enter a collection id for this new collection.

Suggestion:
Keep consistent when adding a collection between VS Code and portal.
capture3

tec-goblin:
CosmosDb collections have Ids, while MongoDb ones have names :)


tec-goblin:
Right! I guess I should have read that instead of monkeying the portal :) https://docs.microsoft.com/en-us/windows/desktop/uxguide/text-ui

PR #831, opened 2018-09-18
Allow user to import documents from workspace into Mongo and SQL collections

PrashanthCorp:


StephenWeatherford:
@PrashanthCorp Is this ready to review?

PrashanthCorp:
Yes.

src/extension.ts
```diff
@@ -145,6 +158,106 @@ async function copyConnectionString(node: IAzureNode<MongoAccountTreeItem | DocD
...
+	return documents;
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocuments(nodes: vscode.Uri[]): Promise<[any[], string[]]> {
+	let documents = [];
+	let errors = {};
+	for (let node of nodes) {
+		const document = (await vscode.workspace.openTextDocument(node));
+		const text = document.getText();
```
EricJizbaMSFT: I think this feature has to account for the fact that users might import a lot of files. I don't think reading the files through VS Code is the most efficient (although I haven't done any testing). I would prefer if you do `fse.readFile`

src/extension.ts
```diff
@@ -145,6 +158,106 @@ async function copyConnectionString(node: IAzureNode<MongoAccountTreeItem | DocD
...
+		let parsed;
+		try {
+			parsed = JSON.parse(text);
+		} catch (e) {
+			const err = parseError(e);
+			const fileName = node.path.split('/').pop();
+			errors[fileName] = err;
+			ext.outputChannel.appendLine(`${fileName}:\n${err}`);
+			await vscode.window.showTextDocument(document);
```
EricJizbaMSFT: What's the thoughts for showing all these files? This could get unwieldy if I'm uploading a bunch of files and several of them have errors. For example, the VS Code explorer lets me select any file type and anything that's not 'json' will fail:
screen shot 2018-09-27 at 11 02 12 am 2

src/extension.ts
```diff
@@ -145,6 +158,106 @@ async function copyConnectionString(node: IAzureNode<MongoAccountTreeItem | DocD
...
+	}
+	return [documents, Object.keys(errors)];
+}
+
+// tslint:disable-next-line:no-any
+async function insertDocumentsIntoDocdb(collectionNode: IAzureParentNode<DocDBCollectionTreeItem>, documents: any[], nodes: vscode.Uri[]): Promise<string> {
+	let result;
+	let ids = [];
+	const collectionTreeItem = (<DocDBCollectionTreeItem>collectionNode.treeItem);
+	const documentsTreeItem: DocDBDocumentsTreeItem = <DocDBDocumentsTreeItem>(await collectionTreeItem.loadMoreChildren(collectionNode, false))[0];
```
EricJizbaMSFT: loadMoreChildren is not meant to be called directly and could have weird effects on the tree. Call `collectionNode.getCachedChildren` instead. I am refactoring the tree to hopefully make this more clear (but don't worry about the refactor for now)

src/extension.ts
```diff
@@ -145,6 +158,106 @@ async function copyConnectionString(node: IAzureNode<MongoAccountTreeItem | DocD
...
+	let i = 0;
+	for (i = 0; i < documents.length; i++) {
+		let document: NewDocument = documents[i];
+		if (!documentsTreeItem.documentHasPartitionKey(document)) {
+			throw new Error(`Error in file ${vscode.workspace.asRelativePath(nodes[i])}. Please ensure every document has a partition key path for the collection you choose to import into.`);
```
EricJizbaMSFT: Can we do this error handling as a part of the parsing error handling? Again - I want these errors to show up to the user as soon as possible, rather than waiting for us to parse everything before I get a notification saying I have to start over again

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -35,13 +35,13 @@ export class MongoCollectionTreeItem implements IAzureParentTreeItem {
 		}
 	}
 
-	public async update(documents: IMongoDocument[]): Promise<IMongoDocument[]> {
+	public async update(documents: IMongoDocument[], flag = false): Promise<IMongoDocument[]> {
```
EricJizbaMSFT: Why the changes to this file? I don't see this being called anywhere in this PR

src/extension.ts
```diff
@@ -145,6 +158,106 @@ async function copyConnectionString(node: IAzureNode<MongoAccountTreeItem | DocD
...
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocumentsForErrors(nodes: vscode.Uri[]): Promise<any[]> {
+	const parseResult = await parseDocuments(nodes);
+	const documents = parseResult[0];
+	const errors: string[] = parseResult[1];
+	if (errors.length > 0) {
+		ext.outputChannel.show();
+		throw new Error(`Errors found in the following documents: ${errors.join(',')}.\nPlease fix these and try again.`);
```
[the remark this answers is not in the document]
PrashanthCorp: In my first prototype, I had the same preference. The first time I demo'ed it, the feedback was pretty strong in favor of making the command transactional/all-or-nothing. So it only made sense to collect all the errors and inform the user once about all of them - that reduces the number of times the user needs to fix errors. 

This is specifically why I didn't go down the path of having a Skip - Skip all - Cancel flow you mention. 

IIRC, the thought process was that if you're importing documents to a database, showing individual errors with skipping requires the user to keep track of failed imports.

src/mongo/tree/MongoCollectionTreeItem.ts
PrashanthCorp: This must have come from an earlier commit which was later undone. I'll fix this.

src/extension.ts
PrashanthCorp: Nice catch. I remember testing the multiselect behavior, and it looked like it would not show unless all the files were .json. 
I tested it some more just now, and it looks like the group of files does not determine context menu as much as _which file_ was right-clicked. I right-clicked on the .cs file for the screenshot.
image

So my code checks for the clicked file, but made an assumption that all files should support the context item => all are jsons.

src/extension.ts
PrashanthCorp: I understand your concern, and I initially agreed with you. The feedback from the first demo however was contrary to what you ask. 

I also feel showing the user all the errors is more beneficial to the user that the speed of showing the first error. This would ensure they start over again a fewer times.

src/extension.ts
PrashanthCorp: I've filed an issue for the symptom you mentioned (selecting a non JSON file) [here](https://github.com/Microsoft/vscode/issues/59782). I'm handling this behavior in the next push. 

I understand where the issue with too many errors. It looks like vscode would overwrite the previously opened document (it is temporary, so the tab-title is italicized). I might scrap opening the documents  entirely, let me play with that.

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+            ignoredUris.push(uri);
+            return false;
+        }
+    });
+    if (ignoredUris.length) {
+        ext.outputChannel.appendLine(`The following selected files are not json: ${ignoredUris.map(uri => uri.fsPath).join(',')}. \nWe are ignoring these.`);
+        ext.outputChannel.show();
+    }
+    const documents = await withProgress(parseDocumentsForErrors(uris), "Parsing documents...", vscode.ProgressLocation.Notification);
```
EricJizbaMSFT: Parsing documents should happen after the user has gone through all UI. This lets them cancel before any long running notification and it avoids delays between input steps.

Specifically, this should be after `showNodePicker`

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,127 @@
...
+// tslint:disable-next-line:no-any
+async function insertDocumentsIntoDocdb(collectionNode: IAzureParentNode<DocDBCollectionTreeItem>, documents: any[], nodes: vscode.Uri[]): Promise<string> {
+    let result;
+    let ids = [];
+    const documentsTreeItem: DocDBDocumentsTreeItem = <DocDBDocumentsTreeItem>(await collectionNode.getCachedChildren()[0].treeItem);
+    let i = 0;
+    for (i = 0; i < documents.length; i++) {
+        let document: NewDocument = documents[i];
+        if (!documentsTreeItem.documentHasPartitionKey(document)) {
+            throw new Error(`Error in file ${vscode.workspace.asRelativePath(nodes[i])}. Please ensure every document has a partition key path for the collection you choose to import into.`);
```
EricJizbaMSFT: Regardless of our disagreements on how to handle errors (throw the first error or list all errors) - this error handling needs to be consistent with the error handling in `parseDocumentsForErrors`. In this case you're throwing only the first error, but in the other case you're listing out all errors.

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,127 @@
...
+    return jsonDocuments.map(choice => choice.uri);
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocumentsForErrors(nodes: vscode.Uri[]): Promise<any[]> {
+    const parseResult = await parseDocuments(nodes);
+    const documents = parseResult[0];
+    const errors: string[] = parseResult[1];
+    if (errors.length > 0) {
+        throw new Error(`Errors found in the following documents: ${errors.join(',')}.\nPlease fix these and try again.`);
```
EricJizbaMSFT: I wouldn't list out all errors here. It's probably way too much text. I would just direct them to the output channel if they need more details

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,127 @@
...
+            }
+            const err = parseError(e);
+            const fileName = uri.path.split('/').pop();
+            errors[fileName] = err;
+            ext.outputChannel.appendLine(`${fileName}:\n${err.message}`);
+            ext.outputChannel.show();
+        }
+        if (parsed) {
+            if (Array.isArray(parsed)) {
+                documents = documents.concat(parsed);
```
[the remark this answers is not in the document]
StephenWeatherford: This is the scenario where they've pointed you to a set of documents in JSON?  Seems like importing everything would be the right thing, no?  We can ask Matt.

It might make sense to confirm the import with a number of document that will be imported so they're not susprised…?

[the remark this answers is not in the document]
StephenWeatherford: Let's review the UI/experience first before making changes like this, though.

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (uri.fsPath.endsWith('.json')) {
+            return true;
+        } else {
+            ignoredUris.push(uri);
+            return false;
+        }
+    });
+    if (ignoredUris.length) {
+        ext.outputChannel.appendLine(`The following selected files are not json: ${ignoredUris.map(uri => uri.fsPath).join(',')}. \nWe are ignoring these.`);
+        ext.outputChannel.show();
```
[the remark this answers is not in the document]
PrashanthCorp: I demoed/talked to Matt about this - he was fine with showing the output in this case.

src/commands/importDocuments.ts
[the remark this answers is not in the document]
PrashanthCorp: Talked to Matt about this - we don't intend to throw errors. It is perfectly valid JSON, and if it has an array of objects, we just add them.

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+import { ext } from '../extensionVariables';
+import { MongoCollectionTreeItem } from '../mongo/tree/MongoCollectionTreeItem';
+
+export async function importDocuments(tree: AzureTreeDataProvider, uris: vscode.Uri[] | undefined, collectionNode: IAzureParentNode<MongoCollectionTreeItem | DocDBCollectionTreeItem> | undefined): Promise<void> {
+    if (!uris) {
+        uris = await askForDocuments();
+    }
+    let ignoredUris: vscode.Uri[] = []; //account for https://github.com/Microsoft/vscode/issues/59782
+    uris = uris.filter((uri) => {
+        if (uri.fsPath.endsWith('.json')) {
```
EricJizbaMSFT: Should this feature support jsonc files?

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+        filters: {
+            "JSON": ["json"]
+        },
+        defaultUri: vscode.Uri.file(vscode.workspace.rootPath)
+    });
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocumentsForErrors(nodes: vscode.Uri[]): Promise<any[]> {
+    const parseResult = await parseDocuments(nodes);
```
EricJizbaMSFT: nit: These three lines can just be `const [documents, hasErrors] = await parseDocuments(nodes);`

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (hasErrors) {
+        throw new Error(`Errors found in some documents.\nPlease see the output, fix these and try again.`);
+    }
+    return documents;
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocuments(uris: vscode.Uri[]): Promise<[any[], boolean]> {
+    let documents = [];
+    let errors = {};
```
EricJizbaMSFT: nit: No need to keep track of errors anymore. Just create a boolean `let hasErrors: boolean = false;` and change `errors[uri.path] = err;` below to `hasErrors = true`

Also this logic can probably be merged with `errorFoundFlag`

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+    let errorFoundFlag: boolean = false;
+    for (let uri of uris) {
+        const text: string = await fse.readFile(uri.fsPath, 'utf-8');
+        let parsed;
+        try {
+            parsed = JSON.parse(text);
+        } catch (e) {
+            if (!errorFoundFlag) {
```
EricJizbaMSFT: You never set `errorFoundFlag` to true, so this gets output multiple times

```
Errors found in documents listed below. Please fix these.
/Users/ericjizba/TestRepos/py1/doc1.json:
Unexpected end of JSON input
Errors found in documents listed below. Please fix these.
/Users/ericjizba/TestRepos/py1/doc2.json:
Unexpected end of JSON input
```

I would prefer if the output looked like this:
```
Errors found in documents listed below. Please fix these.
/Users/ericjizba/TestRepos/py1/doc1.json: Unexpected end of JSON input
/Users/ericjizba/TestRepos/py1/doc2.json: Unexpected end of JSON input
```

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+    if (ignoredUris.length) {
+        ext.outputChannel.appendLine(`The following selected files are not json: ${ignoredUris.map(uri => uri.fsPath).join(',')}. \nIgnoring these.`);
+        ext.outputChannel.show();
+    }
+    if (!collectionNode) {
+        collectionNode = <IAzureParentNode<MongoCollectionTreeItem | DocDBCollectionTreeItem>>await tree.showNodePicker([MongoCollectionTreeItem.contextValue, DocDBCollectionTreeItem.contextValue]);
+    }
+
+    const documents = await vscode.window.withProgress(
```
EricJizbaMSFT: The title should be 'Importing documents...' and it should also wrap the actual import code (not just the parsing code)

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,134 @@
...
+    return documents;
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocuments(uris: vscode.Uri[]): Promise<[any[], boolean]> {
+    let documents = [];
+    let errors = {};
+    let errorFoundFlag: boolean = false;
+    for (let uri of uris) {
+        const text: string = await fse.readFile(uri.fsPath, 'utf-8');
```
EricJizbaMSFT: Actually I think we want to use `fse.readJson` instead of `fse.readFile` and `JSON.parse`

For example, change the encoding on a file from 'utf-8' to something else (I did 'utf-8 with bom') from the status bar in VS Code:
screen shot 2018-10-04 at 10 25 42 am 2

fse.readJson will automatically detect the encoding and parse appropriately (based on my testing), but the current code will fail with an error like this:
screen shot 2018-10-04 at 10 27 37 am

src/commands/importDocuments.ts
PrashanthCorp: I remember python allows this, but didn't realize TS would be fine with this too. Thanks for the headsup!

src/commands/importDocuments.ts
PrashanthCorp: Good question. I expect using fse.readJson might curtail this to only support json. Let me test this.

PrashanthCorp: fse.readJSOn does not  support jsonc.

package.json
```diff
@@ -62,6 +62,7 @@
 		"onCommand:cosmosDB.deleteDocDBDocument",
```
StephenWeatherford: Where are the tests?  I see neither CTI testcases nor integration/unit tests.  Ideally there would be a simple testcase or two for CTI and some integration tests.

package.json
```diff
@@ -62,6 +62,7 @@
...
 		"onCommand:cosmosDB.deleteGraphDatabase",
```
StephenWeatherford: I got this message:

image

package.json
```diff
@@ -62,6 +62,7 @@
...
 		"onCommand:cosmosDB.deleteGraph",
```
StephenWeatherford: image
Command is somewhat vague when out of context.  How about "Import into MongoDB/CosmosDB Collection"?

package.json
```diff
@@ -62,6 +62,7 @@
...
+		"onCommand:cosmosDB.importDocument",
```
StephenWeatherford: You should add this to the editor context menu as well:
image

StephenWeatherford: See docker package.json for how to do this.

package.json
StephenWeatherford: I guess that's then confusing when on the collection node:
image
Maybe we need two commands so they can have different wording?

package.json
StephenWeatherford: I seem to always get this when imported into mongodb.

src/commands/importDocuments.ts
```diff
@@ -0,0 +1,145 @@
...
+            "JSON": ["json"]
+        },
+        defaultUri: vscode.Uri.file(vscode.workspace.rootPath)
+    });
+}
+
+// tslint:disable-next-line:no-any
+async function parseDocuments(uris: vscode.Uri[]): Promise<any[]> {
+    let documents = [];
+    let errors = {};
```
EricJizbaMSFT: nit: Remove `errors`. It's not used anymore

package.json
PrashanthCorp: Filing an issue for this.

PR #834, opened 2018-09-21
Support using ObjectID as a constructor in scrapbook.

### LINKED ISSUE #769: Understand ObjectId and other common function calls
- [ ] ObjectID
- [ ] ISODate

PrashanthCorp:
Fixes #769 

I will push this after the regex PR.


StephenWeatherford:
@PrashanthCorp  Could you fix conflicts (and regenerate grammar) after you merge your regex fix so I can play with this?  Thx.

PrashanthCorp:
Done with the merge. Let's wait for Travis to give the go-ahead before asking you to play with it.

StephenWeatherford:
Could you merge so I can see this without the regex changes? Thx.

src/mongo/MongoScrapbook.ts
```diff
@@ -294,12 +285,62 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 				parsedObject = [];
 			}
 		} else if (child instanceof mongoParser.FunctionCallContext || child instanceof ErrorNode) {
-			return {};
+			if (child.text.startsWith("ObjectId")) {
```
StephenWeatherford: What's child.text usually?  Don't like matching without word boundaries.

src/mongo/MongoScrapbook.ts
```diff
@@ -294,12 +285,62 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+				parsedObject = this.objectIdContextToObject(ctx, child.text);
+			} else {
+				return {};
```
StephenWeatherford: Should be doing an error instead of return {}.  Does your ISODate work fix that?

src/mongo/MongoScrapbook.ts
StephenWeatherford: Don't rely so much on the text - you're already using a parser, so take advantage of that.  function name is child.children[0].text (just make sure to do non-null checks)

src/mongo/MongoScrapbook.ts
```diff
@@ -294,12 +285,62 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+			}
 		} else {
 			return {};
 		}
 
 		return parsedObject;
 	}
 
+	private objectIdContextToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, tokenText: string): Object {
+		let opening = tokenText.indexOf('(');
```
StephenWeatherford: Again, use the parser output rather than parsing it again.

src/mongo/MongoScrapbook.ts
```diff
@@ -294,12 +285,62 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+		let closing = tokenText.indexOf(')');
+		let hexID: string;
+		let tokenObject: Object = {};
+		if (closing === opening + 1) { // usage : ObjectID()
+			tokenObject = new ObjectID();
+		} else {
+			if (tokenText[opening + 1] === "\"") {
```
StephenWeatherford: What about single quotes?

PrashanthCorp:
Ping.

src/mongo/MongoScrapbook.ts
```diff
@@ -7,8 +7,9 @@ import { ANTLRInputStream as InputStream } from 'antlr4ts/ANTLRInputStream';
 import { CommonTokenStream } from 'antlr4ts/CommonTokenStream';
```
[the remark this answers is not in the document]
StephenWeatherford: Waiting for reply...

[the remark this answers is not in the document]
PrashanthCorp: Yes, with two caveats: 
While both the [BSON ObjectId docs])(https://docs.mongodb.com/manual/reference/method/ObjectId/) and the ObjectId constructor suggest that 12 bytes are okay, the shell is not happy with this. 
image

A document with a 12-byte ObjectId in a field gets converted to 24-hex (not surprising), and can be found. The hexId in `db.people.find({"name": ObjectId("abcdef123456")})` gets converted to `616263646566313233343536`. Inserting the above document and searching for it works. 

As to other cases I've tested: 
1. Errors while typing show up in problems and disappear once the id is legit. 
2. It consistently converts ids having special characters - for eg., `!@#$%^&*()-_` can be inserted and found. 
3. A multi-line split like 
```
{"name": 
ObjectId(

"abcdef123456")

})
```
works. Splitting midway inside "abcdef123456" does not work: The resultant string, say "abcdef1\r\n23456" is longer than 12 bytes. Removing 2 of the characters would still work.

PR #840, opened 2018-09-28
Added get database command.

Omelianenko:
Need this for Connections node in App Service


src/extension.ts
```diff
@@ -122,14 +124,17 @@ export function activate(context: vscode.ExtensionContext) {
...
-			(this: IActionContext, event: vscode.ConfigurationChangeEvent): Promise<void> {
+		async function (this: IActionContext, event: vscode.ConfigurationChangeEvent): Promise<void> {
 			this.properties.isActivationEvent = "true";
 			this.suppressErrorDisplay = true;
 			if (event.affectsConfiguration(ext.settingsKeys.documentLabelFields)) {
 				await vscode.commands.executeCommand("cosmosDB.refresh");
 			}
 		});
+	registerCommand('cosmosDB.api.getDatabase', async () => {
+		const AttachedDatabaseTreeItemContextvalue: string = 'cosmosDBAttachDatabaseAccount';
```
EricJizbaMSFT: I don't think 'cosmosDBAttachDatabaseAccount' should be necessary?

Omelianenko: You're right, thanks!

PR #845, opened 2018-10-03
Browse to mongo shell

### LINKED ISSUE #844: We should improve the experience around the mongo.shell.path setting
1) Use file open dialog instead of textbox when asking
2) Auto-fix if directory specified

StephenWeatherford:
Fixes #844

### DIFF
--- a/src/mongo/tree/MongoDatabaseTreeItem.ts
+++ b/src/mongo/tree/MongoDatabaseTreeItem.ts
@@ -3,8 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as fse from 'fs-extra';
 import { Collection, Db } from 'mongodb';
+import opn = require('opn');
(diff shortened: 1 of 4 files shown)


src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -130,28 +137,73 @@ export class MongoDatabaseTreeItem implements IAzureParentTreeItem {
...
+				}
+
+				throw new UserCancelledError();
+			}
+		} else {
+			// User has specified the path or command.  Sometimes they set the folder instead of a path to the file, let's check that and auto fix
+			if (await fse.pathExists(shellPathSetting)) {
+				let stat = await fse.stat(shellPathSetting);
+				if (stat.isDirectory()) {
+					this._cachedShellPathOrCmd = path.join(shellPathSetting, mongoExecutableFileName);
```
PrashanthCorp: Wouldn't we want to give the user a heads-up that we're fixing a setting?

PR #847, opened 2018-10-04
Added api.getConnectionString command

Omelianenko:
Model of command that returns the connection string

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
 	"activationEvents": [
 		"onLanguage:mongo",
 		"onView:cosmosDBExplorer",
-		"onCommand:cosmosDB.executeMongoCommand",
-		"onCommand:cosmosDB.executeAllMongoCommands",
+		"onCommand:cosmosDB.api.getConnectionString",
+		"onCommand:cosmosDB.api.getDatabase",
 		"onCommand:cosmosDB.attachDatabaseAccount",
 		"onCommand:cosmosDB.attachEmulator",
 		"onCommand:cosmosDB.createAccount",
@@ -62,6 +62,8 @@
 		"onCommand:cosmosDB.deleteDocDBDocument",
 		"onCommand:cosmosDB.deleteGraphDatabase",
 		"onCommand:cosmosDB.deleteGraph",
+		"onCommand:cosmosDB.executeAllMongoCommands",
+		"onCommand:cosmosDB.executeMongoCommand",
 		"onCommand:cosmosDB.newMongoScrapbook",
 		"onCommand:cosmosDB.update",
 		"onCommand:cosmosDB.openDocument",
@@ -327,6 +329,11 @@
 				"category": "Cosmos DB",
 				"command": "cosmosDB.openCollection",
 				"title": "Open Collection"
+			},
+			{
+				"category": "Cosmos DB",
+				"command": "cosmosDB.api.getConnectionString",
+				"title": "Get connection string"
 			},
 			{
 				"category": "Cosmos DB",
@@ -658,6 +665,10 @@
 				},
 				{
 					"command": "cosmosDB.loadMore",
+					"when": "never"
+				},
+				{
+					"command": "cosmosDB.api.getConnectionString",
 					"when": "never"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -599,22 +599,6 @@
 				},
 				{
 					"command": "cosmosDB.copyConnectionString",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServerAttached"
-				},
-				{
-					"command": "cosmosDB.copyConnectionString",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBGraphAccountAttached"
-				},
-				{
-					"command": "cosmosDB.copyConnectionString",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentServerAttached"
-				},
-				{
-					"command": "cosmosDB.copyConnectionString",
-					"when": "view == cosmosDBExplorer && viewItem == cosmosDBTableAccountAttached"
-				},
-				{
-					"command": "cosmosDB.copyConnectionString",
 					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -598,6 +598,26 @@
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBTableAccount"
 				},
 				{
+					"command": "cosmosDB.copyConnectionString",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServerAttached"
+				},
+				{
+					"command": "cosmosDB.copyConnectionString",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBGraphAccountAttached"
+				},
+				{
+					"command": "cosmosDB.copyConnectionString",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBDocumentServerAttached"
+				},
+				{
+					"command": "cosmosDB.copyConnectionString",
+					"when": "view == cosmosDBExplorer && viewItem == cosmosDBTableAccountAttached"
+				},
+				{
+					"command": "cosmosDB.copyConnectionString",
+					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
+				},
+				{
 					"command": "cosmosDB.refresh",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServer"
 				},
--- a/package.json
+++ b/package.json
@@ -598,10 +598,6 @@
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBTableAccount"
 				},
 				{
-					"command": "cosmosDB.copyConnectionString",
-					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
-				},
-				{
 					"command": "cosmosDB.refresh",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBMongoServer"
 				},
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,10 +135,9 @@ export function activate(context: vscode.ExtensionContext) {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
 	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
-		const possibleContextValues = { mongoDb: 'mongoDb' };
 		const node = (<IAzureNode>await tree.findNode(treeItemId));
 		switch (node.treeItem.contextValue) {
-			case possibleContextValues.mongoDb:
+			case MongoDatabaseTreeItem.contextValue:
 				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
 			default:
 				throw new Error("Not implemented yet. Works only with Mongo.");
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,7 +135,10 @@ export function activate(context: vscode.ExtensionContext) {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
 	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
-		const node = (<IAzureNode>await tree.findNode(treeItemId));
+		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
+		if (node === undefined) {
+			throw new Error("Coudn't find node with provided Id.");
+		}
 		switch (node.treeItem.contextValue) {
 			case MongoDatabaseTreeItem.contextValue:
 				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -136,14 +136,14 @@ export function activate(context: vscode.ExtensionContext) {
 	});
 	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
 		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
-		if (node === undefined) {
-			throw new Error("Couldn't find the database node in Cosmos DB with provided Id.");
+		if (!node) {
+			throw new Error(`Couldn't find the database node in Cosmos DB with provided Id: ${treeItemId}`);
 		}
 		switch (node.treeItem.contextValue) {
 			case MongoDatabaseTreeItem.contextValue:
 				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
 			default:
-				throw new Error("Not implemented yet. For now works only with Mongo.");
+				throw new Error('Not implemented yet. For now works only with Mongo.');
 		}
 	});
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,7 +135,7 @@ export function activate(context: vscode.ExtensionContext) {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
 	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
-		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
+		const node = await tree.findNode(treeItemId);
 		if (!node) {
 			throw new Error(`Couldn't find the database node in Cosmos DB with provided Id: ${treeItemId}`);
 		}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -137,13 +137,13 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
 		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
 		if (node === undefined) {
-			throw new Error("Coudn't find node with provided Id.");
+			throw new Error("Couldn't find the database node in Cosmos DB with provided Id.");
 		}
 		switch (node.treeItem.contextValue) {
 			case MongoDatabaseTreeItem.contextValue:
 				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
 			default:
-				throw new Error("Not implemented yet. Works only with Mongo.");
+				throw new Error("Not implemented yet. For now works only with Mongo.");
 		}
 	});
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -134,6 +134,16 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getDatabase', async () => {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const possibleContextValues = { mongoDb: 'mongoDb' };
+		const node = (<IAzureNode>await tree.findNode(treeItemId));
+		switch (node.treeItem.contextValue) {
+			case possibleContextValues.mongoDb:
+				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
+			default:
+				throw new Error("Not implemented yet. Works only with Mongo.");
+		}
+	});
 }
 
 async function getAttachedNode(tree: AzureTreeDataProvider): Promise<IAzureParentNode<AttachedAccountsTreeItem>> {
--- a/src/mongo/mongoConnectionStrings.ts
+++ b/src/mongo/mongoConnectionStrings.ts
@@ -3,19 +3,55 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-export function getDatabaseNameFromConnectionString(connectionString: string): string | undefined {
-    //  Connection strings follow the following format (https://docs.mongodb.com/manual/reference/connection-string/):
-    // mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
+export function parseConnectionString(connectionString: string): string[] | undefined {
+    // Connection strings follow the following format (https://docs.mongodb.com/manual/reference/connection-string/):
+    //   mongodb[+srv]://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
     // Some example connection strings:
-    //   mongodb://dbuser:dbpassword@dbname.mlab.com:14118/dbname
-    //   mongodb://db1.example.net:27017,db2.example.net:2500/?replicaSet=test
-    //   mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/
+    //   mongodb://dbuser:dbpassword@dbname.mlab.com:14118
+    //   mongodb+srv://db1.example.net:27017,db2.example.net:2500/?replicaSet=test
+    //   mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/database?ssh=true
+    // Regex splits into five parts:
+    //   Full match
+    //   mongodb[+srv]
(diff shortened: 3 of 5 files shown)


src/extension.ts
```diff
@@ -134,6 +134,15 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getDatabase', async () => {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const node = (<IAzureNode>await tree.findNode(treeItemId));
```
EricJizbaMSFT: Don't cast this to `IAzureNode`. The type of `IAzureNode | undefined` is correct. You need to handle the case where it's undefined (aka it didn't find the node). Maybe just throw an error

PrashanthCorp: I'll add that throwing an error would show up to the user as stemming from the Cosmos extension instead of the caller function. The report error button would also submit an error to cosmos - so we'll have to look here to see how the feature is doing.

package.json
```diff
@@ -660,6 +667,10 @@
 					"command": "cosmosDB.loadMore",
 					"when": "never"
 				},
+				{
+					"command": "cosmosDB.api.getConnectionString",
+					"when": "never"
+				},
```
StephenWeatherford: Hide getDatabase from command palette

src/extension.ts
```diff
@@ -134,6 +134,18 @@ export function activate(context: vscode.ExtensionContext) {
...
+		if (node === undefined) {
+			throw new Error("Coudn't find node with provided Id.");
+		}
+		switch (node.treeItem.contextValue) {
+			case MongoDatabaseTreeItem.contextValue:
+				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
+			default:
+				throw new Error("Not implemented yet. Works only with Mongo.");
+		}
+	});
```
StephenWeatherford: What about api.getDatabase?

src/extension.ts
```diff
@@ -134,6 +134,18 @@ export function activate(context: vscode.ExtensionContext) {
...
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
+		if (node === undefined) {
+			throw new Error("Coudn't find node with provided Id.");
+		}
+		switch (node.treeItem.contextValue) {
+			case MongoDatabaseTreeItem.contextValue:
+				return (<IAzureNode<MongoDatabaseTreeItem>>node).treeItem.connectionString;
```
StephenWeatherford: Needs a mocha test. Let me know if you need help getting started.

package.json
```diff
@@ -37,8 +37,8 @@
 	"activationEvents": [
 		"onLanguage:mongo",
 		"onView:cosmosDBExplorer",
-		"onCommand:cosmosDB.executeMongoCommand",
-		"onCommand:cosmosDB.executeAllMongoCommands",
+		"onCommand:cosmosDB.api.getConnectionString",
```
StephenWeatherford: So I believe we were going to discuss whether it would be better to provide an API off of the extension object rather than using a command ID, right?  Were you planning on merging before that discussion?

Omelianenko: Right, as I understand Eric doesn't want to provide any CosmosDB objects in AppService. I can continue to work on appservice changes even without merging, so I'll merge after discussion.

package.json
Omelianenko: Done that on the [previous PR](https://github.com/Microsoft/vscode-cosmosdb/pull/840)

src/extension.ts
StephenWeatherford: Good observation.  That's more reason to use an API format rather than commands.

src/extension.ts
```diff
@@ -134,6 +134,18 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getDatabase', async () => {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
+		if (node === undefined) {
+			throw new Error("Couldn't find the database node in Cosmos DB with provided Id.");
```
StephenWeatherford: Please include the ID in the error message.

src/extension.ts
```diff
@@ -134,6 +134,18 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getDatabase', async () => {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
+		if (node === undefined) {
```
StephenWeatherford: I prefer

  if (!node) 

That way it handles undefined and null.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -3,19 +3,55 @@
...
+        let [, , , database] = parseConnectionString(connectionString);
+        return database;
+    } catch (error) {
+        // Shouldn't happen, but ignore if does
+    }
+
+    return undefined;
+}
 
+export function getDatabaseConnectionStringByAccountConnectionString(connectionString: string, databaseName: string): string | undefined {
```
EricJizbaMSFT: How about we rename this to `addDatabaseToConnectionString`

package.json
```diff
@@ -590,6 +597,26 @@
 					"command": "cosmosDB.copyConnectionString",
 					"when": "view == cosmosDBExplorer && viewItem == cosmosDBTableAccount"
 				},
+				{
+					"command": "cosmosDB.copyConnectionString",
```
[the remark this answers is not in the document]
Omelianenko: Not the best idea to make this without discussion, just couldn't find any reason for not allowing to copy connecting string of attached databases, so going to make a new pr and issue for this

test/mongoConnectionStrings.test.ts
```diff
@@ -5,53 +5,54 @@
 
 // The module 'assert' provides assertion methods from node
 import * as assert from 'assert';
-import { getDatabaseNameFromConnectionString } from '../src/mongo/mongoConnectionStrings';
+import { parseConnectionString } from '../src/mongo/mongoConnectionStrings';
 
-function testDatabaseFromConnectionString(connectionString: string, expectedDatabase: string | undefined) {
-    let database = getDatabaseNameFromConnectionString(connectionString);
-    assert.equal(database, expectedDatabase);
+function testParseConnectionString(connectionString: string, expectedDatabase: string[]) {
```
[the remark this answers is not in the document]
Omelianenko: The goal was to use existed tests for both functions, so I'm going to create a new tests for addDatabaseToConnectionString(aka getDatabaseConnectionStringByAccountConnectionString). I think that my tests changes may be useful for future development in case if any other part of the connection string will be needed, so what do you think if I make them and refactor testing function in a next way:
`
function testDatabaseItemFromConnectionString(connectionString: string, expectedDatabase: string[]) {
    let database = getDatabaseNameFromConnectionString(connectionString);
    assert.equal(database, expectedDatabase[2]);
}

suite(`mongoCollectionStrings`, () => {
    test(`getDatabaseItemFromConnectionString`, () => {
        testDatabaseItemFromConnectionString(...);
        ...
  `

src/mongo/mongoConnectionStrings.ts
```diff
@@ -3,19 +3,55 @@
...
+    return undefined;
+}
 
+export function addDatabaseToConnectionString(connectionString: string, databaseName: string): string | undefined {
     try {
-        let matches = connectionString.match(/[a-zA-Z]+:\/\/[^/]+(\/([^/?]+))/);
-        if (matches) {
-            let [, , database] = matches;
-            return database;
+        let [, , , parsedDatabaseName] = parseConnectionString(connectionString);
```
EricJizbaMSFT: Why do you need to get parsedDatabaseName?

src/mongo/mongoConnectionStrings.ts
```diff
@@ -3,19 +3,55 @@
...
+        if (!parsedDatabaseName) {
+            return connectionString.replace(new RegExp('(^mongodb(?:[+]srv)?:\/\/[^\/]*)(?:\/((?:[^/?]*).*)?|$)'), `$1\/${databaseName}$2`);
```
EricJizbaMSFT: Why change from `[a-zA-Z]+` to `(^mongodb(?:[+]srv)?`?

Why change from `[^\/]+` to `[^\/]*`?

Why do you have the second capturing group in the replace?

We want to keep this regexp as close to the existing as possible and I don't think those changes are necessary.

EricJizbaMSFT: getDatabaseNameFromConnectionString and addDatabaseToConnectionString should use the same regex if possible (You can make it a `const` for this file).

Note: If you do that I don't think you need the `parseConnectionString` function.

EricJizbaMSFT: nit: I like `/insertRegex/` better than `new RegExp('insertRegex')`

test/mongoConnectionStrings.test.ts
```diff
@@ -5,55 +5,80 @@
...
+        testDatabaseToConnectionString(`mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/our-mongo?ssl=true&replicaSet=globaldb`, 'our-mongo', 'mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/our-mongo?ssl=true&replicaSet=globaldb')
+
+        testDatabaseToConnectionString(`mongodb://dbuser:dbpassword@dbname.mlab.com:14118`, 'mydata', 'mongodb://dbuser:dbpassword@dbname.mlab.com:14118/mydata');
+        testDatabaseToConnectionString(`mongodb://dbuser:dbpassword@dbname.mlab.com:14118/`, 'database', 'mongodb://dbuser:dbpassword@dbname.mlab.com:14118/database');
+        testDatabaseToConnectionString(`mongodb+srv://dbuser:dbpassword@dbname.mlab.com:14118/database`, 'database', 'mongodb+srv://dbuser:dbpassword@dbname.mlab.com:14118/database');
+
+        testDatabaseToConnectionString(`mongodb://db1.example.net:27017,db2.example.net:2500/?replicaSet=test`, 'my-database', 'mongodb://db1.example.net:27017,db2.example.net:2500/my-database?replicaSet=test');
```
EricJizbaMSFT: These tests need to change the database name to something _other_ than 'my-database' (which is already in the connection string)

test/mongoConnectionStrings.test.ts
```diff
@@ -5,55 +5,80 @@
 
 // The module 'assert' provides assertion methods from node
 import * as assert from 'assert';
-import { getDatabaseNameFromConnectionString } from '../src/mongo/mongoConnectionStrings';
+import { getDatabaseNameFromConnectionString, addDatabaseToConnectionString } from '../src/mongo/mongoConnectionStrings';
```
EricJizbaMSFT: > I think that my tests changes may be useful for future development in case if any other part of the connection string will be needed

Not a fan of "planning for the future" for tests. I'd rather keep it simple and only test the existing stuff. We can make it more complicated once it's actually necessary

test/mongoConnectionStrings.test.ts
Omelianenko: Not sure that it's even possible to get parameters like these: database name and database name inside connection string are different. How you can see, for now, I don't replace the name of database in connection string. Is this a way how we want to resolve this?

src/mongo/mongoConnectionStrings.ts
Omelianenko: 1. `[a-zA-Z]+` is not the exact pattern for MongoDB connection string, it should be ```mongodb``` or ```mongodb+srv```.
2. `[^\/]*` parses strings like this: `mongodb:///?replicaSet=test` as `["mongodb://", ""(undefined), ""(undefined), "?replicaSet=test"]`, but `[^\/]+` doesn't parse. Not sure which behavior is better 
3. Options are the part of mongo connection string, so I thought that we also should keep it for database connection strings.

Actually, I have the same pattern in `parseConnectionString`

Omelianenko: I'm going to delete all tests changes for getDatabaseNameFromConnectionString that I've made, so I'll use the same regex and delete that function

src/mongo/mongoConnectionStrings.ts
Omelianenko: In case if connection string already contains database name I don't want to replace it, but if you think that it's better to replace I won't need it anymore. mongodb+srv://dbuser:dbpassword@dbname.mlab.com:14118/database`, 'notdatabase'

src/mongo/mongoConnectionStrings.ts
EricJizbaMSFT: 1. Can you link to documentation for 1? You're making the pattern more strict and I want to make sure we're not breaking users with connections strings we haven't thought of.
1. I don't think `mongodb:///?replicaSet=test` is a valid connection string
1. Try a test without the capturing group. The 'replace' method only replaces the matching part and should keep the rest of the string (the options) intact

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: This needs to show some example transformations in the comments.

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: From a quick rereading of the docs (in the comments), I don't think the mongodb+srv scheme allows a database name. Do you disagree?

src/mongo/mongoConnectionStrings.ts
```diff
@@ -3,19 +3,55 @@
...
+    //   mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/database?ssh=true
+    // Regex splits into five parts:
+    //   Full match
+    //   mongodb[+srv]
+    //   [username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]]
+    //   [database]
+    //   [?options]
+
+    try {
+        let matches = connectionString.match('(^mongodb(?:[+]srv)?):\/\/([^\/]*)(?:\/([^/?]*)(.*))?$');
```
StephenWeatherford: I don't think you should have that ^ inside the parentheses.

StephenWeatherford: Would be much easier to read if you break up the regex into multiple string constants.

src/mongo/mongoConnectionStrings.ts
Omelianenko: Seems like you're right, couldn't find any docs that say about mongo+srv connection string with a database too. Is it possible to attach or maybe even create a database account with connection string of `mongodb+srv` format inside cosmos?

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: Yes, but not sure it should be inside a group.  At least, I've had trouble with this before. I'd put it before the group.

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: @PrashanthCorp You're the Cosmos expert now. Do you know how to test a connection string with mongodb+srv?

EricJizbaMSFT: Can we keep the original `[a-zA-Z]+` and ignore (or file an issue for) the 'mongodb+srv' scenario? I feel like we're going down a rabbit hole that I'm not convinced anyone is using (and I don't want to test). In general I think it would be better for @Omelianenko to keep the regex as similar to the original as possible and focus on the App Service scenario rather than Cosmos DB edge cases.

Omelianenko: It's for sure one of the two possible variants: `mongodb `or `mongodb+srv` ([link](https://docs.mongodb.com/manual/reference/connection-string/)), so it's not hard to make `mongodb(?:[+]srv)` instead of `[a-zA-Z]+`, but it's not a big deal at all, so just going to leave `[a-zA-Z]+`

test/mongoConnectionStrings.test.ts
```diff
@@ -5,55 +5,83 @@
...
         // mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
 
-        testDatabaseFromConnectionString(`mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/?ssl=true&replicaSet=globaldb`, undefined);
-        testDatabaseFromConnectionString(`mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/our-mongo?ssl=true&replicaSet=globaldb`, "our-mongo");
-        testDatabaseFromConnectionString(`mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/?ssl=true&replicaSet=globaldb`, undefined);
+        testDatabaseNameFromConectionString(`mongodb://my-mongo:ayO83FFfUoHE97Jm7WbfnpNCqiF0Yq0za2YmvuLAKYJKf7h7hQaRKWfZfsv8Ux41H66Gls7lVPEKlKm0ueSozg==@your-mongo.documents.azure.com:10255/?ssl=true&replicaSet=globaldb`, undefined);
+        testDatabaseNameFromConectionString(`mongodb://a/b?c`, "b");
```
EricJizbaMSFT: Why'd you change this connection string?

test/mongoConnectionStrings.test.ts
```diff
@@ -5,55 +5,83 @@
...
+function testDatabaseToAccountConnectionString(connectionString: string, databaseName: string, expectedConnectionString: string | undefined) {
+    let databaseConnectionString = addDatabaseToAccountConnectionString(connectionString, databaseName);
+    assert.equal(databaseConnectionString, expectedConnectionString);
+}
+
+function testDatabaseNameFromConectionString(connectionString: string, expectedDatabaseName: string | undefined) {
+    let databaseName = getDatabaseNameFromConnectionString(connectionString);
+    assert.equal(databaseName, expectedDatabaseName);
+    if (databaseName) {
+        testDatabaseToAccountConnectionString(connectionString, databaseName, connectionString);
```
EricJizbaMSFT: I'm not sure I see the value of having` testDatabaseToAccountConnectionString` inside `testDatabaseNameFromConectionString`

It seems cleaner and more reliable to have explicit/separate tests for each one

src/mongo/mongoConnectionStrings.ts
```diff
@@ -3,20 +3,34 @@
...
+//   mongodb://dbuser:dbpassword@dbname.mlab.com:14118
+//   mongodb+srv://db1.example.net:27017,db2.example.net:2500/?replicaSet=test
+//   mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/database?ssh=true
+// Regex splits into three parts:
+//   Full match
+//   mongodb[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]]
+//   [database]
+const parsePrefix = '^(mongodb(?:[+]srv)?:\/\/[^\/]*)';
+const parseDatabaseName = '\/?([^/?]+)?';
+const parseConnectionString = new RegExp(parsePrefix + parseDatabaseName);
```
EricJizbaMSFT: > Would be much easier to read if you break up the regex into multiple string constants

Not a blocker for the PR, but I disagree with @StephenWeatherford on this one. I think splitting it up makes it more prone to mistakes (plus "easier to read" is debatable). When you're typing out a RegExp in the form `/fdsafsdfasd/` you will get errors if you have delimiters in the wrong place, etc. but you don't get that with strings. For example only the bottom one here has a red squiggly:
screen shot 2018-10-12 at 11 33 53 am

EricJizbaMSFT: Rename this to `connectionStringRegExp`

StephenWeatherford: You'll get the error when you create the RegExp.

StephenWeatherford: And it's easier to read because:
1) you can easily see where the groups are
2) it becomes self-documentating

The longer it becomes, the more this is true.

test/mongoConnectionStrings.test.ts
Omelianenko: Had issues with regex, so changed that for simplifying my fixing process and forgot to turn it back

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: 3. You can re-use pieces in similar regexes if necessary

src/mongo/mongoConnectionStrings.ts
```diff
@@ -13,7 +13,7 @@
 //   Full match
 //   mongodb[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]]
 //   [database]
-const parsePrefix = '^(mongodb(?:[+]srv)?:\/\/[^\/]*)';
+const parsePrefix = '([a-zA-Z]+:\/\/[^\/]*)';
```
StephenWeatherford: You removed the ^ and $.  But this is better anyway (^ and $ shouldn't be in the component regexes - separation of concerns).

const parseConnectionString = new RegExp(`^${parsePrefix}${parseDatabaseName}\$`);

StephenWeatherford: … which indicates you should add some tests that have extra characters at the beginning and end of the string

test/mongoConnectionStrings.test.ts
StephenWeatherford: Given the discussion, makes sense.

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: Never mind... :-)

src/mongo/mongoConnectionStrings.ts
StephenWeatherford: Given the shorter length, less important.  Your call.

src/mongo/mongoConnectionStrings.ts
EricJizbaMSFT: I'm not worried about it. In an ideal world @Omelianenko should file an issue to test the "mongodb+srv" scenario before release, but at this point I just care about getting the core scenarios merged with minimal risk of regressions in Cosmos DB.

src/extension.ts
```diff
@@ -134,6 +134,18 @@ export function activate(context: vscode.ExtensionContext) {
 	registerCommand('cosmosDB.api.getDatabase', async () => {
 		return (<IAzureParentNode>await tree.showNodePicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).id;
 	});
+	registerCommand('cosmosDB.api.getConnectionString', async (treeItemId: string) => {
+		const node = (<IAzureNode | undefined>await tree.findNode(treeItemId));
```
EricJizbaMSFT: nit: Casting to `IAzureNode | undefined` should be unnecessary. That should already be the return type of findNode

package.json
```diff
@@ -590,6 +597,10 @@
...
+					"when": "view == cosmosDBExplorer && viewItem == mongoDb"
```
EricJizbaMSFT: Again - this is a UI change unrelated to the main PR and should be a separate for the reasons I mentioned earlier. I also forgot to mention this is important for the commit history (For example sometimes we look at the commit history when writing the changelog before a release)

Also you would need to change this `https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L112` to also accept `MongoDatabaseTreeItem`

Issue #795, opened 2018-08-20
MongoDB limit() function doesn't seem to be supported when querying from scrapbook files


zachsa:
MongoDB `limit(<n>)` is not supported by API? Writing the following in `Scrapbook-1.mongo`:

This works:
`db.<collectionName>.find({})`

But this does not;
`db.<collectionName>.find({}).limit(10)`

I'm quite new to Mongo so I could be missing something obvious. But looking at this documentation: https://docs.mongodb.com/manual/reference/method/cursor.limit/

It seems to me that the query with `limit(10)` should work. should it work?

PrashanthCorp:
Hi Zach, 
Thanks for using the extension and reaching out to us. 
Unfortunately, we don't support the limit query as of now. 
Would allowing a limit field inside the parameters of `find` work for you?

zachsa:
Thanks for getting back to me. I’m new to Mongo and didn’t realize there was a limit field.

 

It would be helpful if you mentioned the features that aren’t supported in the readme – would have saved me quite a lot of time (and I’m sorry if such is present and I just didn’t see it!).

 

I really like using VS Code as a database IDE by the way. Thank you very much for providing this software and please add me to any mailing lists you have.

PrashanthCorp:
Hi @zachsa , 

All public discussion happens on Github. 

It looks like another symptom of your bug is that method chaining is not supported. Looks like referencing #785 might help us keep track of this.

PrashanthCorp:
Closing via #850

PR #850, opened 2018-10-05
 Set method chaining to flag; use it to execute command in shell

### LINKED ISSUE #785: [Scrapbook]: Chaining method does not work?
Typing cmd like `db.users.find(_criteria_).count()` returns suitable user documents, not their count.

PrashanthCorp:
Fixes #785 , #795


test/mongoGetCommand.test.ts
```diff
@@ -450,12 +450,11 @@ suite("scrapbook parsing Tests", () => {
         assert.deepEqual(err.range.start.character, 26);
     });
 
-    test("Trying to use pretty()", () => {
+    test("Chained command: to use pretty()", () => {
```
StephenWeatherford: Need more tests.  Be creative - try to break your code.  And what about all the testcases in the issues that have been filed?

PrashanthCorp: As a first iteration, I wanted feedback on the code initially (whether my approach is starkly different from what's expected). 

I'll add tests too now. :)

StephenWeatherford: Gotcha.  Saw nothing obviously wrong with the approach, although I'd need to play with it to determine how successful it is.

StephenWeatherford: How good is the error handling experience (both in the editor and trying to run commands)?

test/mongoGetCommand.test.ts
```diff
@@ -749,6 +774,30 @@ suite("scrapbook parsing Tests", () => {
...
+
+    test("Chained command- test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/795", () => {
+        testParse('db.timesheets.find({}).limit(10);', {
+            collection: "timesheets",
+            name: "limit",
+            args: [{}, 10]
+        });
+    });
+
+    test("Chained command alternative for rs.slaveOk()- test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/565", () => {
```
StephenWeatherford: How about a case with it spread out on multiple lines?

test/mongoGetCommand.test.ts
StephenWeatherford: Reply?

PrashanthCorp: I tried out the following cases: 
I miss the separating period -> Expected an EOF, rather than the second functionCall
Mid-typing: 
Have dot, not function call -> Expected a functionCall instead of ';'
Haven't entered brackets -> Expected '(', instead of ';'

Across commands: 
If there's no semi-colon, the error occurs at the start of the next command. 
If there's a semicolon, the error position is contained within current command. 

Running: 
It only runs if the command seems legit. There could be errors from the shell.
Eg. `db.people.find().limit()` would execute, but return a shell error . The limit function needs an argument.

StephenWeatherford: So do you think the user experience is acceptable?

PrashanthCorp: The behavior for errors across commands seems compliant with behavior elsewhere. 
Running the command has a fail-safe (the shell). 
Mid-type : The errors make sense to me. 

I think it is acceptable UX wise.

PrashanthCorp:
Ping

PR #879, opened 2018-10-17
Refactored reveal command

Omelianenko:
Moved the reveal command code to the separate file.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,6 +8,7 @@
 import * as copypaste from 'copy-paste';
 import * as vscode from 'vscode';
 import { AzureTreeDataProvider, AzureTreeItem, AzureUserInput, IActionContext, IAzureUserInput, registerCommand, registerEvent, registerUIExtensionVariables, SubscriptionTreeItem } from 'vscode-azureextensionui';
+import { revealTreeItem } from '../src/commands/revealTreeItem';
 import { importDocuments } from './commands/importDocuments';
 import { CosmosEditorManager } from './CosmosEditorManager';
 import { DocDBDocumentNodeEditor } from './docdb/editors/DocDBDocumentNodeEditor';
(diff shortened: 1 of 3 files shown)


Omelianenko:
Should I create "api" folder under "commands" folder and move all the other command code there?

src/extension.ts
```diff
@@ -45,7 +46,9 @@ export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(tree);
 	ext.tree = tree;
 	context.subscriptions.push(vscode.window.registerTreeDataProvider('cosmosDBExplorer', tree));
-	const customView = vscode.window.createTreeView('cosmosDBExplorer', { treeDataProvider: tree });
+
+	const cosmosView = vscode.window.createTreeView('cosmosDBExplorer', { treeDataProvider: tree });
+	ext.cosmosView = cosmosView;
```
EricJizbaMSFT: nit: you can just do:
```
ext.cosmosView = vscode.window.createTreeView('cosmosDBExplorer', { treeDataProvider: tree });
```

src/extensionVariables.ts
```diff
@@ -18,6 +18,7 @@ export namespace ext {
     export let outputChannel: OutputChannel;
     export let reporter: TelemetryReporter | undefined;
     export let tree: AzureTreeDataProvider;
+    export let cosmosView: TreeView<AzureTreeItem<ISubscriptionRoot>>;
```
EricJizbaMSFT: I prefer to just name this `treeView`. This whole repo is Cosmos, so there's no need to have that in the name

src/commands/api/getDatabase.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DocDBDatabaseTreeItem } from '../../../src/docdb/tree/DocDBDatabaseTreeItem';
```
EricJizbaMSFT: nit: `'../../../src/docdb/tree/DocDBDatabaseTreeItem'` is equivalent to `'../../docdb/tree/DocDBDatabaseTreeItem'`

In other words you can remove '../src' from the middle of the path. Same comment in the getConnectionString file

PR #880, opened 2018-10-17
Inform user to start emulator on timeout(SQL) or specific error(Mongo)

### LINKED ISSUE #706: [Windows][Mongo] We need a better error message if emulator not started

### LINKED ISSUE #873: [SQL][Emulator] Expanding the tree item when emulator not installed leads to a long delay before error message
The main issue is that the documentClient is unable to hit the given endpoint, waits a full seconds, and retries another 120 times. This amounts to at least 2 minutes, and that is not good. 
This happens _instead_ of a thrown error (even on await), and hence a simple try catch with async await will not work. 

Ideally, this should be fixed along with the other emulator bug.

PrashanthCorp:
Fixes #706 
Fixes #873

### DIFF
--- a/src/docdb/tree/DocDBAccountTreeItemBase.ts
+++ b/src/docdb/tree/DocDBAccountTreeItemBase.ts
@@ -8,6 +8,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { AzureParentTreeItem, AzureTreeItem, UserCancelledError } from 'vscode-azureextensionui';
 import { deleteCosmosDBAccount } from '../../commands/deleteCosmosDBAccount';
+import { rejectOnTimeout } from '../../utils/timeout';
 import { getDocumentClient } from '../getDocumentClient';
 import { DocDBTreeItemBase } from './DocDBTreeItemBase';
 import { IDocDBTreeRoot } from './IDocDBTreeRoot';
@@ -79,6 +80,11 @@ export abstract class DocDBAccountTreeItemBase extends DocDBTreeItemBase<Databas
         }
 
         throw new UserCancelledError();
+    }
+
+    public async loadMoreChildrenImpl(clearCache: boolean) {
+        let unableToReachEmulatorMessage: string = "Unable to reach emulator. Please ensure it is started and writing to the appropriate port. Then try again.";
+        return await rejectOnTimeout(2000, () => super.loadMoreChildrenImpl(clearCache), unableToReachEmulatorMessage);
     }
 
     private static validateDatabaseName(name: string): string | undefined | null {
--- a/src/docdb/tree/DocDBAccountTreeItemBase.ts
+++ b/src/docdb/tree/DocDBAccountTreeItemBase.ts
@@ -83,8 +83,12 @@ export abstract class DocDBAccountTreeItemBase extends DocDBTreeItemBase<Databas
     }
 
     public async loadMoreChildrenImpl(clearCache: boolean) {
-        let unableToReachEmulatorMessage: string = "Unable to reach emulator. Please ensure it is started and writing to the appropriate port. Then try again.";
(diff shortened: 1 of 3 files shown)


src/docdb/tree/DocDBAccountTreeItemBase.ts
```diff
@@ -81,6 +82,11 @@ export abstract class DocDBAccountTreeItemBase extends DocDBTreeItemBase<Databas
         throw new UserCancelledError();
     }
 
+    public async loadMoreChildrenImpl(clearCache: boolean) {
+        let unableToReachEmulatorMessage: string = "Unable to reach emulator. Please ensure it is started and writing to the appropriate port. Then try again.";
+        return await rejectOnTimeout(2000, () => super.loadMoreChildrenImpl(clearCache), unableToReachEmulatorMessage);
```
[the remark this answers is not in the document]
PrashanthCorp: I should have wrapped this inside an isEmulator flag if possible. 
I'm saying that an account fetching a list of databases shouldn't be more that 2 seconds. I'll admit that I forgot to confine this to emulators (my testing of your graph account didn't show this miscalculation on my part).

src/mongo/tree/MongoAccountTreeItem.ts
```diff
@@ -73,7 +73,14 @@ export class MongoAccountTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
             return databases
                 .filter((database: IDatabaseInfo) => !(database.name && database.name.toLowerCase() === "admin" && database.empty)) // Filter out the 'admin' database if it's empty
                 .map(database => new MongoDatabaseTreeItem(this, database.name, this.connectionString));
-        } finally {
+        } catch (error) {
+            let message = parseError(error).message;
+            if (this._root.isEmulator && message.includes("ECONNREFUSED")) {
+                error.message = "Unable to reach emulator. Please ensure it is started and writing to the appropriate port. Then try again.";
```
StephenWeatherford: Still not really happy with this approach, although I understand the logic. If you're going to use it, at least add the original message to the end, thanks.

Ideally we'd have a common util to do this properly.

src/utils/timeout.ts
```diff
@@ -24,12 +24,12 @@ export async function valueOnTimeout<T>(timeoutMs: number, timeoutValue: T, acti
 /**
  * Returns the result of awaiting a specified action. Rejects if the action throws or if the time-out occurs.
  */
-export async function rejectOnTimeout<T>(timeoutMs: number, action: () => Promise<T> | T) {
+export async function rejectOnTimeout<T>(timeoutMs: number, action: () => Promise<T> | T, callerTimeOutError?: string) {
```
StephenWeatherford: callerTimeOutMessage

It would actually be useful in the future to allow an Error object, but for now please be more specific about what it is, thanks.

src/mongo/tree/MongoAccountTreeItem.ts
PrashanthCorp: Do you prefer this PR to add a common util of the form: 
```
prefixErrorWithMessage(error, message) ..

appendErrorWithMessage(error, message)...
```

StephenWeatherford: Would love to see that, yeah.  How about nestError(errorish: unknown, mainMessage: string)

src/docdb/tree/DocDBAccountTreeItemBase.ts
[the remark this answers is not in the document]
StephenWeatherford: Okay. :-)  Still, you should restrict the timeout to emulator. Networks can be slow or down, and the current code will say the emulator's at fault no matter what.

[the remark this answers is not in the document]
PrashanthCorp: Fixed this in the subsequent push

PR #886, opened 2018-10-19
Moved API commands into object

### LINKED ISSUE #884: Create API object instead of cosmosDB.api.[commands]

### LINKED ISSUE #902: findTreeItem api method that gets object with connectionstring and returns revealable database object

Omelianenko:
Need for App Service Connections feature. Used in [this AppService PR](https://github.com/Microsoft/vscode-azureappservice/pull/665) instead of command execution. 
Fix #884
Fix #902


src/vscode-cosmosdb.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        if (node instanceof MongoDatabaseTreeItem) {
+            return node.connectionString;
+        } else {
+            throw new Error('Not implemented yet. For now works only with Mongo.');
+        }
+    }
+
+    async getDatabase(): Promise<string> {
+        return (await ext.tree.showTreeItemPicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).fullId;
```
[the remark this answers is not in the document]
Omelianenko: Tried to cancel without these changes and didn't get an error(without "All Exceptions", and got a lot of them with checked "All Exceptions" box). Is it necessary?

image
image

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Traverses the CosmosDB tree with a quick pick at each level. Goes until find item with database-level context value. Returns the CosmosDB tree item id.
+     */
+    readonly getDatabase: () => Promise<string>;
+
+    /**
+     * Reveal tree item in the CosmosDB explorer by its id
+     * @param treeItemId The id of the CosmosDB tree item
+     */
+    readonly revealTreeItem: (treeItemId: string) => Promise<void>;
```
EricJizbaMSFT: You can simplify these function declarations. For example this one can just be:
```
revealTreeItem(treeItemId: string): Promise<void>;
```

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Returns connection string of the CosmosDB item by specified id. Throw an error if the item doesn't have a connection string.
+     * @param treeItemId The id of the CosmosDB tree item
+     */
+    readonly getConnectionString: (treeItemId: string) => Promise<string>;
+
+    /**
+     * Traverses the CosmosDB tree with a quick pick at each level. Goes until find item with database-level context value. Returns the CosmosDB tree item id.
+     */
+    readonly getDatabase: () => Promise<string>;
```
EricJizbaMSFT: I think this PR would be a good opportunity to refactor the API. For example - we discussed in standup how we don't want you parsing the tree id. Instead you can return an object like this (or whatever properties you need):
```
{
    id: string;
    accountName: string;
    databaseName: string;
}
```

Also couldn't you just include the connection string on that object and get rid of the `getConnectionString` function?

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface VSCodeCosmosDB {
+    /**
+     * Returns connection string of the CosmosDB item by specified id. Throw an error if the item doesn't have a connection string.
+     * @param treeItemId The id of the CosmosDB tree item
+     */
+    readonly getConnectionString: (treeItemId: string) => Promise<string>;
+
+    /**
+     * Traverses the CosmosDB tree with a quick pick at each level. Goes until find item with database-level context value. Returns the CosmosDB tree item id.
```
StephenWeatherford: getDatabase doesn't imply any interaction with the user.  How about quickPickDatabase or pickDatabase?

src/commands/api/getDatabase.ts
```diff
@@ -7,6 +7,6 @@ import { DocDBDatabaseTreeItem } from '../../docdb/tree/DocDBDatabaseTreeItem';
 import { ext } from '../../extensionVariables';
 import { MongoDatabaseTreeItem } from '../../mongo/tree/MongoDatabaseTreeItem';
 
-export async function getDatabase(): Promise<String> {
+export async function getDatabase(): Promise<string> {
     return (await ext.tree.showTreeItemPicker([MongoDatabaseTreeItem.contextValue, DocDBDatabaseTreeItem.contextValue])).fullId;
```
StephenWeatherford: Also, we should not be returning the raw id.  As mentioned in the meeting, the ID format that's exposed externally from cosmos has to remain unchanged, even if the internal implementation changes.

So you need a conversion function to/from fullId to the external Id, and we should discuss what the external ID looks like.

src/vscode-cosmosdb.api.d.ts
Omelianenko: If I don't parse id than I know 2 options of getting accountName & databaseName:
1. When CosmosDB node is revealing in AppService explorer under Connections node I need to run some sort of getDatabase API function to get those Names
2. Store this info in workspace settings

I feel like the first option is really bad, and I'm not quite sure that the idea of storing the additional info in workspace settings is good. I prefer to return some sort of "stable" id as @StephenWeatherford mentioned [here](https://github.com/Microsoft/vscode-cosmosdb/pull/886#discussion_r227990941) and parse it to get accountName & databaseName.

It's possible to get rid of the `getConnectionString`, but if I do then there is no way to get a connection string after I picked right database after getDatabase() call and got settings added. I don't like this because it complicates creating some sort of re-connect procedure that possibly could be useful:
- In the case of multiple connections of the same type, you could change which one is in the application settings(actually connected) without deleting/re-adding
- If any of appSetting were accidentally deleted you can restore it without deleting/re-adding

Haven't spoken about this with Matt, so absolutely not sure that it's the things that someday we're going to implement.

So, if we'll decide to delete getConnectionString method I can return object like this: 
```
{
    id: string,
    connectionString: string
}
```

StephenWeatherford: > When CosmosDB node(in AppService explorer under Connections node) is revealing I need to run some sort of getDatabase API function to get those Names

Are you talking about code in Cosmos or in App Service?  CosmosDB owns the format of the ID, it's perfectly fine for it to parse/interpret the value.

App Service should make no assumptions about the ID format.

Omelianenko: > App Service should make no assumptions about the ID format.

That's a reason why I need to create conversion function that you mentioned, so AppService makes an assumption about id from conversion function. Do I understand right?

That's how I store a connection in workspace settings:
image
accountName and databaseName I got from those strings under "appService.connections" -> "cosmosDB". I need them for UI:
image


I'm going to implement conversion function, which will return always the same ID value for the node, so this "external" ID will be always the same and you don't need to change it in all extensions that uses it.

StephenWeatherford: Let me know when we can discuss.

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface CosmosDBDatabase {
+    accountName: string
+    connectionString: string
+    treeItemId: string
+    databaseName: string | undefined
```
EricJizbaMSFT: databaseName should always be defined

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface CosmosDBDatabase {
+    accountName: string
+    connectionString: string
+    treeItemId: string
```
EricJizbaMSFT: I don't think we want to return treeItemId. I think `revealTreeItem` should take in the same thing that `getDatabase` does (aka connection string for now) and the Cosmos DB extension should find the node based on that information.

As Stephen has said before - the tree item id is an internal id that we probably don't want to expose.

src/commands/api/pickDatabase.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export async function pickDatabase(): Promise<CosmosDBDatabase> {
+    const pickedDatabase = (await ext.tree.showTreeItemPicker([MongoDatabaseTreeItem.contextValue]));
+
+    const attachedAccountsNode = <AttachedAccountsTreeItem | undefined>(await ext.tree.getChildren()).find((subscription) => {
+        return (subscription.id === 'cosmosDBAttachedAccounts');
+    });
+
+    if (pickedDatabase instanceof MongoDatabaseTreeItem) {
+        const databaseFullId = await attachedAccountsNode.attachDatabase(pickedDatabase);
```
EricJizbaMSFT: pickDatabase should not attach anything. It should only pick from existing databases and should have no side effects on the Cosmos DB extension.

On the flip-side I think you should pass in the connectionString to revealTreeItem and _that_ can attach the node if it doesn't find it under the Azure subscriptions

src/commands/api/getDatabase.ts
```diff
@@ -3,10 +3,33 @@
...
+    });
+    if (attachedAccountsNode) {
+        const attachedAccounts = await attachedAccountsNode.getCachedChildren();
+        for (const account of attachedAccounts) {
+            if (account instanceof AzureParentTreeItem) {
+                const databases = await account.getCachedChildren();
+                for (const database of databases) {
+                    if (database instanceof MongoDatabaseTreeItem && database.connectionString === detectionData.connectionString) {
+                        return {
+                            accountName: account.label,
```
EricJizbaMSFT: You don't actually need to get a node in the tree. You just need to leverage some existing functions (probably `getServerIdFromConnectionString` and `getDatabaseNameFromConnectionString`) to get the accountName and databaseName

src/vscode-cosmosdb.api.d.ts
Omelianenko: As I understand @StephenWeatherford  doesn't want to store treeItemId anywhere because anytime this id can be changed and everything will be broken(Am I wrong?). But now I don't store it before extension running, so even if you changed or deleted some databases/accounts you can always run refresh and everything will be fixed(actually if in running time we catch the error that some connection got broken we can just refresh connection node).
Also, don't like this because the getDatabase command is very heavy as it goes among all attached accounts/databases and I feel like it's very not-optimal that we went through all attached databases every time when we want to reveal something.

src/vscode-cosmosdb.api.d.ts
Omelianenko: Agreed, but have a feeling that this API became very strict

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    databaseName: string | undefined
+}
+
+export interface VSCodeCosmosDB {
+    /**
+     * Finds the database in CosmosDB and returns CosmosDBDatabase object or undefined if can't find
+     * @param detectionData The database connection string
```
StephenWeatherford: How about 'searchCriteria'?

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,30 @@
...
+     */
+    getDatabase(detectionData: { connectionString: string }): Promise<CosmosDBDatabase | undefined>;
+
+    /**
+     *  Traverses the CosmosDB tree with a quick pick at each level. Goes until find item with database-level context value. Returns the CosmosDBDatabase object based on picked db.
+     */
+    pickDatabase(): Promise<CosmosDBDatabase | undefined>;
```
StephenWeatherford: Our other UI methods throw UserCancelled instead of returning undefined.  Is there a reason this one doesn't do the same?

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Reveal tree item in the CosmosDB explorer by its id
+     * @param treeItemId The id of the CosmosDB tree item
+     */
+    revealTreeItem(treeItemId: string): Promise<void>;
```
StephenWeatherford: 'revealDatabase'?

src/vscode-cosmosdb.api.d.ts
EricJizbaMSFT: I asked him to do this. I thought it didn't make sense to "expose" UserCancelledError through an "official" api since other teams might use this eventually. I figured it was better to follow the VS Code pattern which just returns undefined.

src/vscode-cosmosdb.api.d.ts
EricJizbaMSFT: You either need to make databaseName always defined or change the name of the interface - because it doesn't make sense for something named `CosmosDBDatabase` not to have `databseName`

src/vscode-cosmosdb.api.d.ts
StephenWeatherford: Okay. Let's doc that behavior, then.

src/vscode-cosmosdb.api.d.ts
Omelianenko: This method can reveal any treeItemId(not only databases) or is it better to have current scenario names for it?

src/commands/api/pickDatabase.ts
Omelianenko: I think it's better to create the attachDatabase api command, how do you think?

Omelianenko: Now reveal command can process any CosmosDB explorer tree items. If we decide to attach the database in it, we will have less functionality and this command won't be general, so I prefer to have two API commands that do simple things.

src/commands/api/findTreeItem.ts
```diff
@@ -0,0 +1,73 @@
...
+            if ((account instanceof MongoAccountTreeItem) && isParentAccount(connectionString, account)) {
+                const databases = await account.getCachedChildren();
+                for (const database of databases) {
+                    if ((database instanceof MongoDatabaseTreeItem) && connectionString === database.connectionString) {
+                        return {
+                            connectionString: connectionString,
+                            databaseName: database.databaseName,
+                            hostName: hostport.host,
+                            port: hostport.port,
+                            reveal: () => reveal(database.fullId)
```
EricJizbaMSFT: This can just be `reveal: async () => await ext.treeView.reveal(database)`

src/commands/api/pickTreeItem.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const pickedDatabase = <MongoDatabaseTreeItem>(await ext.tree.showTreeItemPicker(contextValuesToFind));
+    const hostport = await getHostPortFromConnectionString(pickedDatabase.connectionString);
+    return {
+        databaseName: pickedDatabase.databaseName,
+        connectionString: pickedDatabase.connectionString,
+        hostName: hostport.host,
+        port: hostport.port,
+        azureData: { accountName: pickedDatabase.parent.name },
+        reveal: () => reveal(pickedDatabase.fullId)
```
EricJizbaMSFT: Same here. This can just be:
```
reveal: async () => await ext.treeView.reveal(pickedDatabase)
```

src/mongo/mongoConnectionStrings.ts
```diff
@@ -1,3 +1,7 @@
+import { ReplSet } from "mongodb";
```
EricJizbaMSFT: nit: Move these below the copyright notice

src/graph/tree/GraphAccountTreeItem.ts
```diff
@@ -17,8 +17,8 @@ export class GraphAccountTreeItem extends DocDBAccountTreeItemBase {
     public static contextValue: string = "cosmosDBGraphAccount";
     public contextValue: string = GraphAccountTreeItem.contextValue;
 
-    constructor(parent: AzureParentTreeItem, id: string, label: string, documentEndpoint: string, private _gremlinEndpoint: IGremlinEndpoint | undefined, masterKey: string, isEmulator: boolean) {
-        super(parent, id, label, documentEndpoint, masterKey, isEmulator);
+    constructor(parent: AzureParentTreeItem, id: string, name: string, label: string, documentEndpoint: string, private _gremlinEndpoint: IGremlinEndpoint | undefined, masterKey: string, isEmulator: boolean) {
```
EricJizbaMSFT: instead of adding a `name: string` parameter, please add a parameter at the end that is `, databaseAccount?: DatabaseAccount` which is only passed in from `CosmosDBAccountProvider` (not AttachedAccountsTreeItem)

src/commands/api/pickTreeItem.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const pickedDatabase = <MongoDatabaseTreeItem>(await ext.tree.showTreeItemPicker(contextValuesToFind));
+    const hostport = await getHostPortFromConnectionString(pickedDatabase.connectionString);
+    return {
+        databaseName: pickedDatabase.databaseName,
+        connectionString: pickedDatabase.connectionString,
+        hostName: hostport.host,
+        port: hostport.port,
+        azureData: { accountName: pickedDatabase.parent.name },
```
EricJizbaMSFT: Based on my other comment, this should be:
```
azureData: pickedDatabase.parent.databaseAccount ? { accountName: pickedDatabase.parent.databaseAccount.name } : undefined,
```

We only want azureData defined if we know this is from Azure

src/extension.ts
```diff
@@ -144,9 +144,11 @@ export function activate(context: vscode.ExtensionContext) {
 				await vscode.commands.executeCommand("cosmosDB.refresh");
 			}
 		});
-	registerCommand('cosmosDB.api.getConnectionString', getConnectionString);
-	registerCommand('cosmosDB.api.getDatabase', getDatabase);
-	registerCommand('cosmosDB.api.revealTreeItem', revealTreeItem);
+
+	return {
```
EricJizbaMSFT: Please merge with master to address the merge conflicts here. The `apiVersion` on the new object should be "1.0.0"

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,71 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+
+export interface CosmosDBExtensionApi {
+    /**
+     * Finds the first matching item in the Cosmos DB tree, or otherwise returns undefined.
+     * NOTE: The item may not actually be loaded/attached in the tree until 'reveal' is called.
+     *
+     * @param query The query object to use for the find
+     * @param options Configures the behavior of the find
```
EricJizbaMSFT: nit: Remove this line about options. I didn't put any options here

src/vscode-cosmosdb.api.d.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export interface CosmosDBTreeItem {
+    /**
+     * Reveals the item in the tree. This may result in loading more Cosmos DB tree items or manually attaching by connection string.
+     */
+    reveal(): Promise<void>;
+}
+
+export interface AccountTreeItem extends CosmosDBTreeItem {
```
EricJizbaMSFT: This should be `DatabaseAccountTreeItem` for the sake of consistency

src/graph/tree/GraphAccountTreeItem.ts
Omelianenko: The DatabaseAccount class doesn't contain the name value there, not sure that it's possible to get it

src/commands/api/findTreeItem.ts
```diff
@@ -0,0 +1,60 @@
...
+import { MongoAccountTreeItem } from '../../mongo/tree/MongoAccountTreeItem';
+import { MongoDatabaseTreeItem } from '../../mongo/tree/MongoDatabaseTreeItem';
+import { AttachedAccountsTreeItem } from '../../tree/AttachedAccountsTreeItem';
+import { DatabaseTreeItem, TreeItemQuery } from '../../vscode-cosmosdb.api';
+import { reveal } from './reveal';
+
+function isParentAccount(connectionString: string, account: MongoAccountTreeItem): boolean {
+    const databaseHostPort = getHostPortFromConnectionString(connectionString);
+    const accountHostPort = getHostPortFromConnectionString(account.connectionString);
+    return (databaseHostPort === accountHostPort);
```
EricJizbaMSFT: I don't think this works. I think you need to do:
```
databaseHostPort.host === accountHostPort.host && databaseHostPort.port === accountHostPort.port
```

Omelianenko:
After Azure database was connected, if you click on Reveal Database button on the notification, it reveals database you picked. But if you click on reveal button on explorer side, it attaches this database(if it's not already attached) and reveals in Attached. It's obvious how to fix this, but I'm not sure that we want it as I'm going to create one more PR that will change findTreeItem function, so it will be possible to find Azure database under subscriptions nodes.

PR #887, opened 2018-10-19
Task 2.0.0

torn4dom4n:
Detail: https://code.visualstudio.com/Docs/editor/tasks

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -11,7 +11,8 @@
 	"tasks": [
 		{
 			"label": "Compile",
-			"type": "npm",
+			"type": "shell",
+			"command": "npm",
 			"presentation": {
 				"reveal": "never"
 			},
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -7,13 +7,15 @@
 // ${cwd}: the current working directory of the spawned process
 // A task runner that calls a custom npm script that compiles the extension.
 {
-	"version": "0.1.0",
+	"version": "2.0.0",
 	// we want to run npm
 	"command": "npm",
 	// the command is a shell script
-	"isShellCommand": true,
+	"type": "shell",
 	// show the output window only if unrecognized errors occur.
-	"showOutput": "silent",
+	"presentation": {
+		"reveal": "never"
+	},
 	// we run the custom script "compile" as defined in package.json
 	"args": [
 		"run",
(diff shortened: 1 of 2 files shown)


.vscode/tasks.json
```diff
@@ -7,22 +7,23 @@
...
-		"silent"
-	],
-	// The tsc compiler is started in watching mode
-	"isBackground": true,
-	// use the standard tsc in watch mode problem matcher to find compile problems in the output.
-	"problemMatcher": "$tsc-watch"
-}
\ No newline at end of file
+	"version": "2.0.0",
+	"task": [
```
EricJizbaMSFT: "task" should be "tasks"

.vscode/tasks.json
```diff
@@ -7,22 +7,23 @@
...
+		{
+			"label": "build",
```
EricJizbaMSFT: This label needs to match the "preLaunchTask" in `.vscode/launch.json`. I would change those to "build".

Otherwise you see this when you start debugging:
screen shot 2018-10-19 at 10 06 54 am

.vscode/tasks.json
```diff
@@ -7,22 +7,23 @@
...
+			"command": "npm",
+			"type": "shell",
+			"presentation": {
+				"reveal": "never"
+			},
+			"args": [
```
EricJizbaMSFT: Since you're updating the task, can you add:
```
"group": {
	"kind": "build",
	"isDefault": true
},
```

If you open up the command palette:
screen shot 2018-10-19 at 10 15 24 am

This designates this task as the build task to be run as a part of this:
screen shot 2018-10-19 at 10 14 43 am

torn4dom4n:
@EricJizbaMSFT can you review it again?

.vscode/tasks.json
torn4dom4n: I changed it to "Compile". Is it OK?

EricJizbaMSFT: I'm fine with compile, but the important part is that it needs to match whatever the value for "preLaunchTask" is in `.vscode\launch.json`

https://github.com/Microsoft/vscode-cosmosdb/blob/master/.vscode/launch.json#L18

torn4dom4n: Oh sorry! It is a mistake. I fixed it. Thanks.

PR #888, opened 2018-10-20
Add supporting of doc db SQL databases.

### LINKED ISSUE #883: Create connection string for SQLdatabases
Need this for supporting SQL connections in App Service

Omelianenko:
Fix #883

### DIFF
--- a/src/commands/api/getConnectionString.ts
+++ b/src/commands/api/getConnectionString.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { DocDBDatabaseTreeItem } from '../../docdb/tree/DocDBDatabaseTreeItem';
 import { ext } from '../../extensionVariables';
 import { MongoDatabaseTreeItem } from '../../mongo/tree/MongoDatabaseTreeItem';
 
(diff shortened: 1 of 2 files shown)


src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -48,6 +48,10 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
         return this._database._self;
     }
 
+    public get connectionString(): string {
+        return `AccountEndpoint=${this.root.documentEndpoint};AccountKey=${this.root.masterKey};Database=${this.id}`;
```
EricJizbaMSFT: Can you use `this.parent.connectionString` and just append the Database?

src/commands/api/getConnectionString.ts
```diff
@@ -12,9 +13,9 @@ export async function getConnectionString(treeItemId: string): Promise<string> {
         throw new Error(`Couldn't find the database node in Cosmos DB with provided Id: ${treeItemId}`);
     }
 
-    if (node instanceof MongoDatabaseTreeItem) {
+    if (node instanceof MongoDatabaseTreeItem || node instanceof DocDBDatabaseTreeItem) {
```
EricJizbaMSFT: I would do `instanceof DocDBDatabaseTreeItemBase` since that's the class you implemented this on

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -49,7 +50,7 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
     }
 
     public get connectionString(): string {
-        return `AccountEndpoint=${this.root.documentEndpoint};AccountKey=${this.root.masterKey};Database=${this.id}`;
+        return this.parent.connectionString.concat(`;Database=${this.id}`);
```
EricJizbaMSFT: `this.id` is the _tree_'s id and _happens_ to be the same as the database id, but that might change in the future. I would use `this._database.id` directly.

PR #894, opened 2018-10-23
Mongo connection string - account for emulator when adding or getting database name

### LINKED ISSUE #881: [Regression]Trying to create database through Mongo emulator fails on current master
Doesn't repro on last shipped version.
Action: cosmosDB.createMongoDatabase
Error type: 18
Error Message: Authentication Failed

Version: 0.8.1-alpha
OS: win32

PrashanthCorp:
FIxes #881 . 
We cannot parse the connection string based on slashes (the emulator password contains slashes). 
Using `@` does not work either : some conn strings might not have a `username:password`. I could not find a rule that fits both emulator and non-emulator accounts, so I handle it in a separate case.

### DIFF
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -18,7 +18,6 @@ export namespace ext {
     export let reporter: ITelemetryReporter;
     export let tree: AzureTreeDataProvider;
     export let treeView: TreeView<AzureTreeItem<ISubscriptionRoot>>;
-    export let emulatorPassword = 'C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==';
 
     export namespace settingsKeys {
         export const mongoShellPath = 'mongo.shell.path';
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -18,6 +18,7 @@ export namespace ext {
     export let reporter: ITelemetryReporter;
     export let tree: AzureTreeDataProvider;
     export let treeView: TreeView<AzureTreeItem<ISubscriptionRoot>>;
+    export let emulatorPassword = 'C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==';
 
     export namespace settingsKeys {
         export const mongoShellPath = 'mongo.shell.path';
--- a/src/mongo/mongoConnectionStrings.ts
+++ b/src/mongo/mongoConnectionStrings.ts
@@ -1,5 +1,3 @@
-import { ext } from "../extensionVariables";
-
 /*---------------------------------------------------------------------------------------------
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
@@ -16,13 +14,15 @@ import { ext } from "../extensionVariables";
 //   mongodb[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]]
 //   [database]
 
+import { emulatorPassword } from "../constants";
+
 const parsePrefix = '([a-zA-Z]+:\/\/[^\/]*)';
 const parseDatabaseName = '\/?([^/?]+)?';
 const connectionStringRegExp = new RegExp(parsePrefix + parseDatabaseName);
 
 export function getDatabaseNameFromConnectionString(connectionString: string): string | undefined {
     try {
-        if (connectionString.includes(ext.emulatorPassword)) {
+        if (connectionString.includes(emulatorPassword)) {
             let portWithDB = connectionString.substring(connectionString.lastIndexOf(':') + 1, connectionString.lastIndexOf('?'));
             if (!portWithDB.includes('/')) {
                 return undefined;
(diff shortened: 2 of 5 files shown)


src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -166,10 +167,10 @@ export class AttachedAccountsTreeItem extends RootTreeItem<ISubscriptionRoot> {
             if (port) {
                 if (defaultExperience.api === API.MongoDB) {
                     // Mongo shell doesn't parse passwords with slashes, so we need to URI encode it
-                    connectionString = `mongodb://localhost:${encodeURIComponent('C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==')}@localhost:${port}/?ssl=true`;
+                    connectionString = `mongodb://localhost:${encodeURIComponent(ext.emulatorPassword)}@localhost:${port}/?ssl=true`;
```
EricJizbaMSFT: How did Stephen get the bug to occur if we're using `encodeURIComponent` here? Did he manually attach to the local emulator? We should handle the manual attach scenario anyways - I'm just curious.

src/extensionVariables.ts
```diff
@@ -18,6 +18,7 @@ export namespace ext {
     export let reporter: ITelemetryReporter;
     export let tree: AzureTreeDataProvider;
     export let treeView: TreeView<AzureTreeItem<ISubscriptionRoot>>;
+    export let emulatorPassword = 'C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==';
```
EricJizbaMSFT: I think this fits better in `constants.ts` than `extensionVariables.ts`

For the record I feel the same way about settingsKeys below, but that's unrelated to your PR so no need to change if you don't want to.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -31,6 +41,9 @@ export function getDatabaseNameFromConnectionString(connectionString: string): s
 
 export function addDatabaseToAccountConnectionString(connectionString: string, databaseName: string): string | undefined {
     try {
+        if (connectionString.includes(ext.emulatorPassword)) {
```
EricJizbaMSFT: You have logic in both `getDatabaseNameFromConnectionString` and `addDatabaseToAccountConnectionString ` to make sure the emulator connection string works. However, I think you should move this logic up the stack so that the connection string on the account tree item is always correct _before_ it gets passed to either of these functions (or anywhere else it's used).

Also I like the replace strategy better than what you did above, but I would suggest this:
```
connectionString.replace(emulatorPassword, encodeUriComponent(emulatorPassword))
```

PrashanthCorp: I'm trying to understand what correct would mean here. Andrii's code would not work because it looks to separate by slashes - that's fine for all cases but the emulator password (`C2y6yDjf5/R+...`). To that end, I don't see how I can sanitize the connection string to  work with the current code.

test/mongoConnectionStrings.test.ts
```diff
@@ -95,6 +103,13 @@ suite(`mongoCollectionStrings`, () => {
...
         testDatabaseToAccountConnectionString(`mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/?`, 'abc.def.-ghi_jkl', 'mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/abc.def.-ghi_jkl?');
         testDatabaseToAccountConnectionString(`mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017`, 'icantlikespaces', 'mongodb://router1.example.com:27017,router2.example2.com:27017,router3.example3.com:27017/icantlikespaces');
+
+        // Emulator
+        testDatabaseToAccountConnectionString(`mongodb://localhost:${ext.emulatorPassword}@localhost:10255?ssl=true`, 'admin', `mongodb://localhost:${ext.emulatorPassword}@localhost:10255/admin?ssl=true`);
+        // Collection within emulator
+        testDatabaseToAccountConnectionString(`mongodb://localhost:${ext.emulatorPassword}@localhost:10255?ssl=true`, 'admin/level1/level2', `mongodb://localhost:${ext.emulatorPassword}@localhost:10255/admin/level1/level2?ssl=true`);
```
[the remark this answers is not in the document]
PrashanthCorp: It cannot. [Here's the doc](https://docs.mongodb.com/manual/reference/limits/#naming-restrictions)

src/mongo/mongoConnectionStrings.ts
EricJizbaMSFT: Correct means you've encoded the password (using encodeUriComponent) so that there are no slashes

src/tree/AttachedAccountsTreeItem.ts
PrashanthCorp: The bug happened outside the attach emulator code. Most likely the emulator was already attached, before encodeURI was used in the code. 
Tested with master, and a newly attached emulator doesn't repro the bug because the connectionString is now correct.

StephenWeatherford: This happened because I rolled back the fix to get another bug to repro. Then the bad URI got saved in settings.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -31,6 +41,9 @@ export function getDatabaseNameFromConnectionString(connectionString: string): s
...
+            return connectionString.replace('?', `/${databaseName}?`);
+        }
         return connectionString.replace(connectionStringRegExp, `$1\/${databaseName}`);
```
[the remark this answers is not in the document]
Omelianenko: It's not possible: 

> mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]]`[/[database][?options]]`

There should be '/' if options are present

PrashanthCorp:
I've incorporated the requested changes. Let me know if there's any other clarifications or changes you need.

PR #900, opened 2018-10-24
Add setting to allow user to change timeout. 

### LINKED ISSUE #809: Can't change timeout
Repro steps:
5 seconds is to little for my mongo database respond. I try to change timeout connection and timeout connection on mongoDB url connection but I think cosmos have a timeout for wait mongo response... Can I change this timeout?

Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "use placeholder"

Version: 0.8.0
OS: darwin

PrashanthCorp:
Fixes #809

### DIFF
--- a/package.json
+++ b/package.json
@@ -765,6 +765,11 @@
 					"description": "Command to execute or full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
 					"default": null
 				},
+				"mongo.shell.timeoutInMilliseconds": {
+					"type": "number",
+					"description": "The duration allowed (in milliseconds) for the mongo shell to execute a command. Default value is 5 seconds",
+					"default": 5000
+				},
 				"cosmosDB.showExplorer": {
 					"type": "boolean",
 					"default": true,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -765,6 +765,11 @@
 					"description": "Command to execute or full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
 					"default": null
 				},
+				"mongo.shell.timeoutInMilliseconds": {
+					"type": "number",
+					"description": "The duration allowed (in milliseconds) for the mongo shell to execute a command. Default value is 5 seconds",
```
StephenWeatherford: Mongo

StephenWeatherford: I like "inMilliseconds" for variable names, but not so sure users will like it for options. I'd go with simple mongo.shell.timeout

PrashanthCorp: Mongo in 'Mongo shell'?

EricJizbaMSFT: If you're going to get rid of "inMilliseconds" I would make the timeout based on seconds. I feel like that's much more common for these types of settings

StephenWeatherford: Current settings are mixed:
image

But I'm good with seconds.

PR #901, opened 2018-10-26
Add conditions to RegexLiteral to avoid matching with multi-line comment '/*' 

### LINKED ISSUE #899: Parser does not recognize multi-line comments
Repro: 

Open scrapbook. Type in: 
```
db.heroes.find()

/*db.c2.insertMany([
    {"name": "Stephen", "age": 38, "male": true},
    {"name": "Stephen", "age": 38, "male": true}
    ]);
*/
```

See a red squiggly near the `/*` citing extraneous input. You also cannot run the uncommented command.

PrashanthCorp:
Fixes #899


grammar/mongo.g4
```diff
@@ -57,7 +57,8 @@ propertyName: StringLiteral | IDENTIFIER;
 
 comment: SingleLineComment | MultiLineComment;
 
-RegexLiteral: '/' (~[/] | '\\/')+ '/' (RegexFlag)*;
+RegexLiteral:
+	'/' (~[/\n*] | '\\/') (~[/\n] | '\\/')* '/' (RegexFlag)*;
```
StephenWeatherford: Don't you need \r, too?

StephenWeatherford: I would add a comment explaining what you're doing.

StephenWeatherford: is // allowed in the Console as a regular expression?

PrashanthCorp: // is not allowed.

test/mongoGetCommand.test.ts
```diff
@@ -958,5 +958,13 @@ suite("scrapbook parsing Tests", () => {
         assert.deepEqual(command.argumentObjects, [{}, {}]);
         assert.deepEqual(command.errors[0].message, "mismatched input 'f' expecting {'{', '[', RegexLiteral, StringLiteral, 'null', BooleanLiteral, NumericLiteral}")
     });
+
+    test("test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/899 - multi-line comment, not regex", () => {
```
StephenWeatherford: If you used testParse it would be a lot easier for you.  Also, it automatically runs with \n changed to \r\n, etc., which I suspect you have a bug with.

PrashanthCorp: I looked into using testParse. However, testParse uses ensureSingleCommand. I didn't want to remove it and change all the test cases depending on it, so I created another test case instead.

grammar/mongo.g4
PrashanthCorp: It looks like \t is allowed, so I'm going to allow that.

test/mongoGetCommand.test.ts
StephenWeatherford: Okay, but then where's the testcase using \r?  We need this since we've had bugs in the past that were Windows only because of this.

StephenWeatherford: Thx

PR #908, opened 2018-10-31
Change returned objects of regexp, ObjectId, Date

### LINKED ISSUE #898: [Regression]The result is null after executing multiple lines and projections
Platform/OS Version: Mac/Windows 10
Architecture: x64
Node.js: 8.9.3
Extension Version: image

Repro Steps:
1. New a scrapbook and connect to one database.
2. Edit the following and run it:
db.heroes.drop();
db.heroes.insert({
    "name": "Batman",
    "saying": "I'm Batman"
});
db.heroes.insert({
"_id": {"$oid" : "5a7e04558ceb7f33005d6134"},
    "name": "Joker",
    "saying": "Ha ha"
});
db.heroes.find({"name": "Joker",
"_id": {"$oid" : "5a7e04558ceb7f33005d6134"},
}, {"name": 0})

Expect:
Get the right result.
image

Actual:
The result is null.
image

PrashanthCorp:
Fixes #898


src/mongo/MongoScrapbook.ts
```diff
@@ -372,23 +374,24 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 			} catch (error) {
 				let err: IParsedError = parseError(error);
 				this.addErrorToCommand(err, ctx);
+				return {};
 			}
 		}
-		return constructedObject;
+		return { $oid: constructedObject.toString() };
```
StephenWeatherford: If you hit the error case, this will cause null ref exception

StephenWeatherford: Do you have a testcase for that scenario?

src/mongo/MongoScrapbook.ts
```diff
@@ -372,23 +374,24 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
 
 	private regexLiteralContextToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, text: string): Object {
 		let separator = text.lastIndexOf('/');
 		let flags = separator !== text.length - 1 ? text.substring(separator + 1) : "";
 		let pattern = text.substring(1, separator);
-		let tokenObject: Object = {};
 		try {
-			tokenObject = new RegExp(pattern, flags);
+			let tokenObject = new RegExp(pattern, flags);
+			return { $regex: tokenObject.source, $options: tokenObject.flags };
```
StephenWeatherford: Why tokenObject.flags instead of just flags?

src/mongo/MongoScrapbook.ts
PrashanthCorp: The error catch block returns an empty object, and [this](https://github.com/Microsoft/vscode-cosmosdb/blob/a0fafcab0093a7ac23f2bcca40d3aadf809b323d/test/mongoGetCommand.test.ts#L691) test case handles an error case. Can you give me more detail on what example you're expecting?

StephenWeatherford: My fault, could've sworn there was no return in the error case.

src/mongo/MongoScrapbook.ts
PrashanthCorp: Here's my thought process. 

In your suggestion an invalid pattern ([example](https://github.com/Microsoft/vscode-cosmosdb/blob/master/test/mongoGetCommand.test.ts#L739)) would still be checked for by `EJSON.parse`. 

Using the Regexp constructor helps because: 
1. The errors thrown by Regexp would be closer to the root cause rather than wrapped by EJSON.
2. Throwing an error here adds to the problems view with the correct position, handling this at EJSON would not. 

This is a bit non-obvious - any suggestion on making this clearer?

src/mongo/MongoScrapbook.ts
```diff
@@ -386,7 +396,10 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 		let pattern = text.substring(1, separator);
 		try {
 			let tokenObject = new RegExp(pattern, flags);
-			return { $regex: tokenObject.source, $options: tokenObject.flags };
+			tokenObject = tokenObject;
+			// validate the pattern and flags.
```
StephenWeatherford: This comment should go before the new RegExp line

src/mongo/MongoScrapbook.ts
```diff
@@ -247,7 +247,17 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 				const argAsObject = this.contextToObject(ctx);
 				const argText = EJSON.stringify(argAsObject);
 				lastCommand.arguments.push(argText);
-				lastCommand.argumentObjects.push(EJSON.parse(argText));
+				let removeDuplicatedBackslash = /\\{4}(?=[0-9bwds.*])/gi;
```
StephenWeatherford: Please remember to comment when you're doing some unexplained like this. Maintainers will have absolutely no clue why you're doing this. It also helps you if you have to explain it in comments.

Suggested comment (please check for correctness):

// In a regex, \b matches a word boundary: {$regex: /\b/}
// But with a string, \b means a backspace, so a word boundary equivalent using a string is {$regex: "\\b"}
// So converting from string to regex means we need to 

…  see, I can't even remember the full details from yesterday's discussion...  That's why we need comments. Can you please finish or modify this comment to explain why you're doing this?  Thx.

test/mongoGetCommand.test.ts
```diff
@@ -705,23 +705,23 @@ suite("scrapbook parsing Tests", () => {
     });
     //Examples inspired from https://docs.mongodb.com/manual/reference/operator/query/regex/
     test("test regular expressions - only pattern, no flags", () => {
-        let text = `db.test1.beep.find({ sku: { $regex: /789$/ } })`;
```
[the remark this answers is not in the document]
PrashanthCorp: Looks like the diff wasn't showing it. It's on lines [738](https://github.com/Microsoft/vscode-cosmosdb/pull/908/files#diff-965f8aeb4d06f91fd82d4d491e31843bR738) through 772.

src/mongo/MongoScrapbook.ts
PrashanthCorp: Done in the next commit. Please let me know if it needs more explanation.

PR #909, opened 2018-11-01
Use unique value for treeItem.id

### LINKED ISSUE #588: Extension will not show any Documents if there are any with the same ID but different partition key values
This is because the ID doesn't include the partition key value, and VS Code throws.

1) In a collection with a partition key, create two new documents with the same ID, but set one to have a value for the partition key (currently has to be done in the portal)
2) in the extension, browse to Documents
!) Documents node will not open

### LINKED ISSUE #892: Collection with documents with the same id
I accidently created a (SQL) document with an id that already existed. Now I'm stuck in a state where I can't delete that document or view any documents.

screen shot 2018-10-22 at 9 43 39 am

First question: Should I be allowed to create a document with the same id? Shouldn't our extension prevent that? 
Second question: Should we handle this situation anyways? Not sure how likely it is - but seems pretty bad that it completely blocks me from viewing _any_ documents or fixing the situation

Here are the two documents (as viewed in the portal). Maybe we should be using "_rid" for the tree's id since that's actually different?
```
{
    "id": "fdsa",
    "dfafsd": "fsda",
    "_rid": "FSQaAIvV3wgDAAAAAAAAAA==",
    "_self": "dbs/FSQaAA==/colls/FSQaAIvV3wg=/docs/FSQaAIvV3wgDAAAAAAAAAA==/",
    "_etag": "\"3500d66a-0000-0700-0000-5bccf3260000\"",
    "_attachments": "attachments/",
    "_ts": 1540158246
}
```

```
{
    "id": "fdsa",
    "dfafsd": "fdsa",
    "_rid": "FSQaAIvV3wgEAAAAAAAAAA==",
    "_self": "dbs/FSQaAA==/colls/FSQaAIvV3wg=/docs/FSQaAIvV3wgEAAAAAAAAAA==/",
    "_etag": "\"41004c6c-0000-0700-0000-5bcdfdc30000\"",
    "_attachments": "attachments/",
    "_ts": 1540226499
}
```

PrashanthCorp:
Fixes #588.
Fixes #892. 

Two documents with same "id" field. 
image

Creating two documents with the same "id" and partition key value still gives the error: 
image


src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -33,7 +33,7 @@ export class DocDBDocumentTreeItem extends AzureTreeItem<IDocDBTreeRoot> {
     }
 
     public get id(): string {
-        return this.document.id;
+        return this.document._rid || `${this.document.id}:${this.getPartitionKeyValue()}`;
```
StephenWeatherford: Add a comment that you expect _rid to be non-empty and the rest is fallback.

Issue #935, opened 2018-11-10
Upgrade findTreeItem api command


Omelianenko:
Now, "findTreeItem" API command trying to find databases only under Attached Database Accounts node. When you reveal Azure database connection, even if it exists under Azure subscription, this command attaches it under Attached and reveals there

PR #942, opened 2018-11-13
Implement searching through subscriptions for findTreeItem

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-cosmosdb/issues/935

Andrii previously only implemented searching through attached accounts (I asked him to do that first because it was simpler). This adds searching through subscriptions as well. I was worried about searching subscriptions taking a while so I added some basic caching and timeouts. (NOTE: In the case of a timeout it doesn't even fail - it just attaches the account instead).


EricJizbaMSFT:
@StephenWeatherford @PrashanthCorp @nturinski FYI this is my main PR that Matt needs by EOD Wednesday for the video he's making. Still work to do before releasing, but this should be good enough for a demo. Please take a look!

src/mongo/mongoConnectionStrings.ts
```diff
@@ -64,8 +64,27 @@ export async function getHostPortFromConnectionString(connectionString: string):
...
-        host: host,
-        port: port
-    };
+    return new ParsedMongoConnectionString(connectionString, host, port, getDatabaseNameFromConnectionString(connectionString));
+}
+
+export class ParsedMongoConnectionString {
+    public readonly host: string;
+    public readonly port: string;
+    public readonly databaseName: string;
```
StephenWeatherford: databaseName: string|undefined

Also please add comment that no db means account level

StephenWeatherford: Aren't some of the others optional, too?

src/commands/api/apiCache.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (sessionCache.has(parsedCS.fullId)) {
+        return sessionCache.get(parsedCS.fullId);
+    } else {
+        return undefined;
+    }
+}
+
+export function removeTreeItemFromCache(expected: ParsedMongoConnectionString): void {
+    if (!expected.databaseName) {
+        // If parsedCS represents an account, iterate and remove all databases that match that account
```
StephenWeatherford: Seems like it might be possible to end up with both an account representation and also database representations, i.e. do you also need to check if there's an account cache to delete, too?

src/commands/api/findTreeItem.ts
```diff
@@ -3,57 +3,87 @@
...
+    // 3. Search subscriptions
+    if (!result) {
+        const rootNodes = await ext.tree.getChildren();
+        for (const rootNode of rootNodes) {
+            if (Date.now() > maxTime) {
+                break;
+            }
+
+            if (rootNode instanceof CosmosDBAccountProvider) {
+                const dbAccounts = await rootNode.getCachedChildren();
```
StephenWeatherford: getCachedChildren means we won't ask Azure for more nodes?

EricJizbaMSFT: Correct. It will ask for the first batch of nodes if they haven't been retrieved yet, but that's it.

src/mongo/mongoConnectionStrings.ts
EricJizbaMSFT: Urgg stictNullChecks would've caught this for me :P 

And I think the rest are required.

src/commands/api/apiCache.ts
EricJizbaMSFT: I think the current code would still work. I'm iterating through the whole cache regardless of if it's a database or account. Anyways I made a few minor changes to hopefully make this more clear.

Admittedly I haven't done a full round of testing for account-level connection strings. Andrii didn't get that "feature" into master yet and I plan to add support in another PR. See error thrown here: https://github.com/Microsoft/vscode-cosmosdb/pull/942/files#diff-bdabf54afc33845cbe50e0fb1f636b7eR25

PR #944, opened 2018-11-14
Add support for account-level connection strings

EricJizbaMSFT:
~~Currently based on this PR: https://github.com/Microsoft/vscode-cosmosdb/pull/942~~


src/commands/api/findTreeItem.ts
```diff
@@ -73,15 +69,19 @@ async function searchDbAccounts(dbAccounts: AzureTreeItem[], expected: ParsedMon
...
+                if (expected.databaseName) {
+                    const dbs = await dbAccount.getCachedChildren();
+                    for (const db of dbs) {
+                        if (db instanceof MongoDatabaseTreeItem && expected.databaseName === db.databaseName) {
+                            return new DatabaseTreeItemInternal(expected, dbAccount, db);
+                        }
                     }
+
+                    // We found the right account - just not the db. In this case we can still 'reveal' the account
+                    return new DatabaseTreeItemInternal(expected, dbAccount);
```
nturinski: Is this the right type?  Not `DatabaseAccountTreeItemInternal`?

EricJizbaMSFT: Yeah. `expected` still has the databaseName and we might be able to find the db node later here: https://github.com/Microsoft/vscode-cosmosdb/pull/944/files#diff-b27d8d588b7338fb2b60237c0e9d82a5R74

PR #951, opened 2018-11-20
Support 'DatabaseAccount' as resourceType in pickTreeItem api

EricJizbaMSFT:
We're not using this directly as a part of the App Service Connections feature, but I believe this is the last "Not supported" part of the api. Plus I might start leveraging this in the Functions extension

### DIFF
--- a/src/commands/api/pickTreeItem.ts
+++ b/src/commands/api/pickTreeItem.ts
@@ -4,17 +4,26 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { parseDocDBConnectionString } from '../../docdb/docDBConnectionStrings';
+import { DocDBAccountTreeItem } from '../../docdb/tree/DocDBAccountTreeItem';
+import { DocDBAccountTreeItemBase } from '../../docdb/tree/DocDBAccountTreeItemBase';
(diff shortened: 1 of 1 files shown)


src/commands/api/pickTreeItem.ts
```diff
@@ -29,19 +38,59 @@ function getDatabaseContextValue(apiType: CosmosDBApiType) {
...
-    let contextValuesToFind = options.apiType ? options.apiType.map(getDatabaseContextValue) : allSupportedDatabaseContextValues;
+export async function pickTreeItem(options: PickTreeItemOptions): Promise<DatabaseTreeItem | DatabaseAccountTreeItem | undefined> {
+    let contextValuesToFind;
+    switch (options.resourceType) {
+        case 'Database':
+            contextValuesToFind = options.apiType ? options.apiType.map(getDatabaseContextValue) : databaseContextValues;
+            break;
+        case 'DatabaseAccount':
+            contextValuesToFind = options.apiType ? options.apiType.map(getAccountContextValue) : accountContextValues;
+            contextValuesToFind = contextValuesToFind.concat(contextValuesToFind.map((val: string) => val += AttachedAccountSuffix));
```
StephenWeatherford: Is that supposed to be "+"?

EricJizbaMSFT: Works either way - but yes :)

PR #954, opened 2018-11-20
Add debounce to openDocument, openGraph, openStoredProc

### LINKED ISSUE #681: Error opening (via double click in azure: cosmos db) mongodb db record.
Repro steps:
<Enter steps to reproduce issue>

Action: cosmosDB.openDocument
Error type: Error
Error Message: TextEditor(vs.editor.ICodeEditor:1,$model1) has been disposed

Version: 0.7.2
OS: win32

PrashanthCorp:
Fixes #681
I initially chose 500ms, but it seemed too long a time to debounce. I chose 200ms as it looked improbable a user could try opening 2 documents within that time frame.

### DIFF
--- a/src/docdb/registerDocDBCommands.ts
+++ b/src/docdb/registerDocDBCommands.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { doubleClickDebounceDelay } from "src/constants";
 import { commands } from "vscode";
 import { registerCommand } from "vscode-azureextensionui";
 import { CosmosEditorManager } from "../CosmosEditorManager";
@@ -68,7 +69,7 @@ export function registerDocDBCommands(editorManager: CosmosEditorManager): void 
         }
         await editorManager.showDocument(new DocDBStoredProcedureNodeEditor(node), node.label + '-cosmos-stored-procedure.js');
         // tslint:disable-next-line:align
-    }, 200);
+    }, doubleClickDebounceDelay);
     registerCommand('cosmosDB.deleteDocDBDocument', async (node?: DocDBDocumentTreeItem) => {
         if (!node) {
             node = <DocDBDocumentTreeItem>await ext.tree.showTreeItemPicker(DocDBDocumentTreeItem.contextValue);
--- a/src/docdb/registerDocDBCommands.ts
+++ b/src/docdb/registerDocDBCommands.ts
@@ -3,9 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { doubleClickDebounceDelay } from "src/constants";
 import { commands } from "vscode";
 import { registerCommand } from "vscode-azureextensionui";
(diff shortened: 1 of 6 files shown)


src/docdb/registerDocDBCommands.ts
```diff
@@ -67,7 +67,8 @@ export function registerDocDBCommands(editorManager: CosmosEditorManager): void
             node = <DocDBStoredProcedureTreeItem>await ext.tree.showTreeItemPicker([DocDBStoredProcedureTreeItem.contextValue]);
         }
         await editorManager.showDocument(new DocDBStoredProcedureNodeEditor(node), node.label + '-cosmos-stored-procedure.js');
-    });
+        // tslint:disable-next-line:align
+    }, 200);
```
StephenWeatherford: How about creating a const?

PrashanthCorp: Sounds good. 

Does the 200 value seem agreeable? I don't quite know what value VSCode is comfortable with (200 worked in my test, but I couldn't find any documentation regarding minimum values)

StephenWeatherford: I'd use what Nathan used.

src/docdb/registerDocDBCommands.ts
```diff
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { doubleClickDebounceDelay } from "src/constants";
```
EricJizbaMSFT: These have to be relative paths like `../../constants.ts` (I'm guessing that's why the build is failing)

VS Code just recently started autocompleting to a path like this. It's super annoying and I want to try to fix it as a part of some engineering work eventually

StephenWeatherford: @EricJizbaMSFT This shouldn't be entered as an issue against vscode?

PrashanthCorp: It looks like it's already entered: 
https://github.com/Microsoft/vscode/issues/59506

The fix is changing the setting `typescript.preferences.importModuleSpecifier` from auto to relative.

EricJizbaMSFT: No I want to get these types of imports working:
```
import { doubleClickDebounceDelay } from "src/constants";
```

I think they look much cleaner and are easier to refactor/change than this:
```
import { doubleClickDebounceDelay } from "../../constants.ts"
```

This is all related to the `baseUrl` setting I added to our tsconfig.json back in August. I did it for an unrelated reason (see [this PR](https://github.com/Microsoft/vscode-cosmosdb/pull/801) for why I originally added it), but it's main point is to support the imports like above. I think VS Code just recently added support for the `baseUrl` property as a part of it's autocomplete which is why we're seeing this now.

EricJizbaMSFT: But I kind of want to wait to do any 'import' or 'module' stuff until after you do webpack :)

EricJizbaMSFT: @PrashanthCorp nice find! Definitely worth changing that in the repo's `.vscode/settings.json` (but I still want to change our imports eventually as described above)

PR #959, opened 2018-11-21
Object to context : add error to problems instead of throwing.

### LINKED ISSUE #958: Parser [patch] - handle thrown errors more gracefully
We currently throw an error on any unexpected input.

PrashanthCorp:
Fixes #958

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -239,23 +239,27 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 	}
 
 	visitArgument(ctx: mongoParser.ArgumentContext): MongoCommand[] {
-		let argumentsContext = ctx.parent;
-		if (argumentsContext) {
-			let functionCallContext = argumentsContext.parent;
-			if (functionCallContext && functionCallContext.parent instanceof mongoParser.CommandContext) {
-				const lastCommand = this.commands[this.commands.length - 1];
-				const argAsObject = this.contextToObject(ctx);
-				const argText = EJSON.stringify(argAsObject);
-				lastCommand.arguments.push(argText);
-				let escapeHandled = this.deduplicateEscapesForRegex(argText);
-				let ejsonParsed = {};
-				try {
-					ejsonParsed = EJSON.parse(escapeHandled);
-				} catch (err) { //EJSON parse failed due to a wrong flag, etc.
-					this.addErrorToCommand(parseError(err), ctx);
+		try {
+			let argumentsContext = ctx.parent;
+			if (argumentsContext) {
+				let functionCallContext = argumentsContext.parent;
+				if (functionCallContext && functionCallContext.parent instanceof mongoParser.CommandContext) {
+					const lastCommand = this.commands[this.commands.length - 1];
+					const argAsObject = this.contextToObject(ctx);
+					const argText = EJSON.stringify(argAsObject);
+					lastCommand.arguments.push(argText);
+					let escapeHandled = this.deduplicateEscapesForRegex(argText);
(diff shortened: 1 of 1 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -281,7 +281,9 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 		} else if (child instanceof ErrorNode) {
 			return {};
 		} else {
-			throw new Error("Unrecognized node type encountered");
+			let err = new Error(`Unrecognized node type encountered. We could not parse ${child.text}`);
```
EricJizbaMSFT: I don't think this is the right way to go about this. If you're call `addErrorToCommand` each time individually then you're asking for another bug like this to happen. Someone else could easily forget to do this in the future.

Plus the current code means we don't get any telemetry.

StephenWeatherford: This was meant only for a quick temporary patch.

src/mongo/MongoScrapbook.ts
```diff
@@ -281,7 +281,9 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+			this.addErrorToCommand(parseError(err), ctx);
```
StephenWeatherford: Why are you creating an error then parsing it and only using the message?

src/mongo/MongoScrapbook.ts
EricJizbaMSFT: @StephenWeatherford I assumed that "patch" was the planned release we have for Cosmos DB next Wednesday (as a part of the app service "Connections" feature). So I had already gone ahead with the other approach: https://github.com/Microsoft/vscode-cosmosdb/pull/960

StephenWeatherford: I'd rather take enough time to make sure this is fixed properly after a quick patch.

PrashanthCorp: The screenshot shows errors thrown from functionCallContextToObject. Almost all the errors in the telemetry arrive from there. I've covered that approach below.

src/mongo/MongoScrapbook.ts
PrashanthCorp: I didn't want to change a whole lot of code. addErrorToCommand takes in an IParsedError. I'll change the typing for addErrorToCommand to only require the a message field.

src/mongo/MongoScrapbook.ts
EricJizbaMSFT: I still want an all-encompassing try/catch before I approve this. Something like this: https://github.com/Microsoft/vscode-cosmosdb/commit/869dd1dae9d51a9a381312adef2f3e7cc5e1b3d2?w=1

StephenWeatherford: That would be good.

src/mongo/MongoScrapbook.ts
```diff
@@ -405,7 +411,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 		}
 	}
 
-	private addErrorToCommand(error: IParsedError, ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext): void {
+	private addErrorToCommand(error: { message: string }, ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext): void {
```
StephenWeatherford: Ah, I see.  Silly to create an error just to pull out them message, but don't change it for the patch.

src/mongo/MongoScrapbook.ts
PrashanthCorp: Pushed. :)

PR #960, opened 2018-11-21
Never show Mongo parse errors as dialog

EricJizbaMSFT:
And track these in telemetry.

Didn't realize @PrashanthCorp had been working on this as well and I figured I needed to get this done if I wanted the release to go out next week for the Connections feature.

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -10,7 +10,7 @@ import { ParseTree } from 'antlr4ts/tree/ParseTree';
 import { TerminalNode } from 'antlr4ts/tree/TerminalNode';
 import { ObjectID } from 'bson';
 import * as vscode from 'vscode';
-import { IActionContext, IParsedError, parseError, callWithTelemetryAndErrorHandlingSync } from 'vscode-azureextensionui';
+import { callWithTelemetryAndErrorHandlingSync, IActionContext, IParsedError, parseError } from 'vscode-azureextensionui';
 import { CosmosEditorManager } from '../CosmosEditorManager';
 import { ext } from '../extensionVariables';
 import { filterType, findType } from '../utils/array';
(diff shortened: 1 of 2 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -243,18 +243,24 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+						this.suppressTelemetry = true;
+
+						const lastCommand = me.commands[me.commands.length - 1];
+						const argAsObject = me.contextToObject(ctx);
+						const argText = EJSON.stringify(argAsObject);
+						lastCommand.arguments.push(argText);
+						let escapeHandled = me.deduplicateEscapesForRegex(argText);
+						lastCommand.argumentObjects.push(EJSON.parse(escapeHandled));
+					});
+				} catch (err) {
```
PrashanthCorp: This seems to use the position from the larger context. 
Have you tested to check if the positions mentioned are correct?

EricJizbaMSFT: It uses the same `ctx` passed in to `contextToObject` a few lines above. Isn't that right?

screen shot 2018-11-21 at 5 06 37 pm

src/mongo/MongoScrapbook.ts
```diff
@@ -243,18 +243,24 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
-				const argAsObject = this.contextToObject(ctx);
-				const argText = EJSON.stringify(argAsObject);
-				lastCommand.arguments.push(argText);
-				let escapeHandled = this.deduplicateEscapesForRegex(argText);
-				let ejsonParsed = {};
 				try {
-					ejsonParsed = EJSON.parse(escapeHandled);
-				} catch (err) { //EJSON parse failed due to a wrong flag, etc.
+					//tslint:disable:no-var-self
+					const me = this;
```
PrashanthCorp: If you're going to store the current object context in another variable, it would make more sense for the name to be discernible - say visitor.

StephenWeatherford: "me = this" is a well-established JavaScript pattern

StephenWeatherford:
I'd rather have the patch put in place today, then fix properly but not have it be a blocker for next week's release.

EricJizbaMSFT:
Closing due to too many changes

PR #967, opened 2018-11-28
askForDocuments - throw error if no workspace found

### LINKED ISSUE #961: There shows a confused error message when running 'Import Document into a Collection' with no Workspace opened
**Platform / OS Version:**  Windows/ Mac / Linux
**Node.js** : V8.12.0
**Build Version:** [20181121.5](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=773&view=logs)

**Repro Steps:**
1. Make sure there is no warkspace opened.
2. Open Command Palette -> Type 'Import Document into a Collection' -> Run this command.
3. Check the result.

**Expect Experiment:**
There shows a clear error message.

**Actual Experiment:**
There shows a confused error message as below.
error

PrashanthCorp:
Fixes #961. 
The vague error arises from showOpenDialog. Hence I'm performing a check and printing a better message instead.

### DIFF
--- a/src/commands/importDocuments.ts
+++ b/src/commands/importDocuments.ts
@@ -57,18 +57,18 @@ export async function importDocuments(uris: vscode.Uri[] | undefined, collection
 }
 
 async function askForDocuments(): Promise<vscode.Uri[]> {
-    let rootPath = vscode.workspace.rootPath;
-    if (!rootPath) {
-        throw new Error("No workspace found. Please open a workspace");
-    }
-    return await ext.ui.showOpenDialog({
+    let openDialogOptions: vscode.OpenDialogOptions = {
         canSelectMany: true,
         openLabel: "Import",
         filters: {
             "JSON": ["json"]
-        },
-        defaultUri: vscode.Uri.file(rootPath)
-    });
+        }
+    };
+    let rootPath = vscode.workspace.rootPath;
+    if (rootPath) {
+        openDialogOptions.defaultUri = vscode.Uri.file(rootPath);
+    }
+    return await ext.ui.showOpenDialog(openDialogOptions);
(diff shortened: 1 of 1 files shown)


src/commands/importDocuments.ts
```diff
@@ -57,13 +57,17 @@ export async function importDocuments(uris: vscode.Uri[] | undefined, collection
 }
 
 async function askForDocuments(): Promise<vscode.Uri[]> {
+    let rootPath = vscode.workspace.rootPath;
+    if (!rootPath) {
+        throw new Error("No workspace found. Please open a workspace");
```
EricJizbaMSFT: I don't think this should block the user from importing documents. I'm pretty sure `defaultUri` is an optional parameter below or you could consider doing `os.homedir()` if it's not optional

PrashanthCorp: Sounds good.

src/commands/importDocuments.ts
```diff
@@ -57,14 +57,18 @@ export async function importDocuments(uris: vscode.Uri[] | undefined, collection
...
 
 async function askForDocuments(): Promise<vscode.Uri[]> {
-    return await ext.ui.showOpenDialog({
+    let openDialogOptions: vscode.OpenDialogOptions = {
         canSelectMany: true,
         openLabel: "Import",
         filters: {
             "JSON": ["json"]
-        },
-        defaultUri: vscode.Uri.file(vscode.workspace.rootPath)
```
StephenWeatherford: 'Course, you could've just done this:

  defaultUri: vscode.workspace.rootPath ? vscode.Uri.file(vscode.workspace.rootPath) : undefined

PR #969, opened 2018-11-28
Add back package-lock

EricJizbaMSFT:
And bump to next alpha


EricJizbaMSFT:
This file is auto-generated by npm. We either use it or don't use it.

> Also, having ^x.y.z does not address the problem that existed (a malware update on a dependent package). This too occurs a few times.

I think you need to look closer at how a package-lock works. For example if you see this:
```
"@types/d3": {
			"version": "5.0.0",
			"requires": {
				"@types/d3-array": "*",
```

That doesn't mean any version of d3-array is allowed. You have to go to the actual entry for d3-array, which has this:
```
"@types/d3-array": {
			"version": "1.2.4",
```

The version is locked to 1.2.4.

StephenWeatherford:
You mentioned there was an issue with npm link.  I usually use npm pack instead, so that's not an issue for me, though.

PR #972, opened 2018-11-30
copy-paste -> clipboardy

PrashanthCorp:
Pending test on linux


package.json
```diff
@@ -895,13 +894,13 @@
 		"azure-arm-cosmosdb": "^1.1.2",
 		"azure-arm-resource": "^3.0.0-preview",
 		"azure-graph": "^2.2.0",
-		"copy-paste": "^1.3.0",
+		"clipboardy": "^1.2.3",
```
EricJizbaMSFT: Maybe add `@types/clipboardy` as well

package.json
```diff
@@ -895,13 +894,13 @@
...
 		"d3": "^3.0.0",
 		"documentdb": "^1.14.2",
+		"event-stream": "3.3.4",
```
EricJizbaMSFT: Please remove event-stream. It's no longer needed since we have package-lock

PR #973, opened 2018-12-03
 Update gulp to 4.0.0. Handle breaking changes in gulp tasks

PrashanthCorp:


gulpfile.js
```diff
@@ -41,7 +32,19 @@ gulp.task('install-azure-account', () => {
             }))
             .pipe(gulp.dest(extensionPath));
     } else {
-        console.log('Azure Account extension already installed.');
+        // We don't use console.log because we need to signal completion of this async task to gulp
+        return gulp.src('package.json')
```
StephenWeatherford: Haven't used gulp much, but the piping seems unnecessary.  Can't you do something like return Promise.resolve('Azure account already installed.')?

gulpfile.js
```diff
@@ -41,7 +32,19 @@ gulp.task('install-azure-account', () => {
...
+
+const testTask = () => {
+    const env = process.env;
+    env.DEBUGTELEMETRY = 1;
+    env.MOCHA_reporter = 'mocha-junit-reporter';
+    env.MOCHA_FILE = path.join(__dirname, 'test-results.xml');
+    return cp.spawn('node', ['./node_modules/vscode/bin/test'], { stdio: 'inherit', env });
+};
+
+gulp.task('test', gulp.series('install-azure-account', testTask));
```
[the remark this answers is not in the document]
PrashanthCorp: It felt better to modularize the definition of the test and the chaining. I don't feel strongly however.

gulpfile.js
PrashanthCorp: I was trying to maintain the same return as the other case, for homogeneity. 

Gulp reports the status correctly, but the message seems missing when using `Promise.resolve("Azure account already installed.")`
image

gulpfile.js
```diff
@@ -41,7 +32,19 @@ gulp.task('install-azure-account', () => {
...
     }
 });
 
+
+const testTask = () => {
+    const env = process.env;
+    env.DEBUGTELEMETRY = 1;
+    env.MOCHA_reporter = 'mocha-junit-reporter';
+    env.MOCHA_FILE = path.join(__dirname, 'test-results.xml');
+    return cp.spawn('node', ['./node_modules/vscode/bin/test'], { stdio: 'inherit', env });
```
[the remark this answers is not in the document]
EricJizbaMSFT: This seems to be new for gulp 4.0. You can return the result of a `cp.spawn` and it just handles it correctly.

https://www.joezimjs.com/javascript/complete-guide-upgrading-gulp-4/ (search for spawn)

Keep in mind gulp 4.0 is still in preview. This could always change, but I think I'm okay relying on it since it gets rid of security warnings.

gulpfile.js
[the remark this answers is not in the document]
EricJizbaMSFT: Slight pref for what Stephen said

gulpfile.js
EricJizbaMSFT: Suggest:
```
console.log('Azure Account extension already installed.');
return Promise.resolve();
```

gulpfile.js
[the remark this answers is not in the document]
StephenWeatherford: Yeah, if it's listening, that's great.

gulpfile.js
StephenWeatherford: Yep. There's no reason to complicate (and do something weird, like pull in a file unrelated to the logic) in order to return the same type of value.

gulpfile.js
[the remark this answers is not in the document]
EricJizbaMSFT: Actually official docs say not to use `gulp.task` any more.

> In the past, task() was used to register your functions as tasks. While that API is still available, exporting should be the primary registration mechanism, except in edge cases where exports won't work.

https://github.com/gulpjs/gulp/blob/master/docs/getting-started/3-creating-tasks.md#exporting

Still no strong preference - but might be reason enough to switch back to what you had before.

package.json
```diff
@@ -898,10 +898,11 @@
 		"copy-paste": "^1.3.0",
 		"d3": "^3.0.0",
 		"documentdb": "^1.14.2",
+		"event-stream": "3.3.4",
```
StephenWeatherford: remove

package.json
```diff
@@ -898,10 +898,11 @@
...
 		"fs-extra": "^4.0.2",
 		"gremlin": "^2.6.0",
+		"gulp-print": "^5.0.0",
```
StephenWeatherford: remove

gulpfile.js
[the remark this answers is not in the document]
PrashanthCorp: Didn't know of this. Thanks for pointing it out. I don't want to make too many changes to the gulpfile just yet (by adding exports and stuff). So I'll leave it as is.

gulpfile.js
```diff
@@ -7,20 +7,11 @@ const gulp = require('gulp');
 const decompress = require('gulp-decompress');
 const download = require('gulp-download');
 const path = require('path');
+const print = require('gulp-print').default;
```
EricJizbaMSFT: nit: remove

Issue #607, opened 2018-05-04
OPINIONS NEEDED: Too many 'result-1.json' files are generated after running mongo commands


EricJizbaMSFT:
Most of the time I never want to save the 'result-1.json' files after running a mongo command. I'm talking about stuff like this:
![screen shot 2018-05-04 at 1 46 18 pm](https://user-images.githubusercontent.com/11282622/39652169-91904888-4fa1-11e8-9482-6e8ade2428a3.png)

And it's made worse because I always get this prompt when I go to close the files:
![screen shot 2018-05-04 at 1 42 47 pm](https://user-images.githubusercontent.com/11282622/39652032-0bfd1dea-4fa1-11e8-950a-f38219e2c9cd.png)

The only thing I might want to save is the output of a 'find' or 'findOne', but that's already handled separately. Maybe this should go to the output channel? I'm not sure.

StephenWeatherford:
I agree we need to figure out an improved experience here, especially with the new Execute All.

StephenWeatherford:
Investigate telemetry: How badly is this needed?

Consider using virtual file loading.

sa-mm:
@dprice I've run into this as well. From looking at the code, it appears that `find` and `findOne` are the only commands that will display in the same window. My guess is that this is because those two commands use an "editor" that's meant to contain some azure integration. The other commands do not, so they just sent to a new window every time. (As an aside, it appears that if you chain a command with find, it won't reuse the editor either—`find().limit(1)`. I'm not sure if that's intended.)

Anyways, I've made some changes that will cause all commands that aren't `find` and `findOne` to reuse a different window. This might be what @StephenWeatherford when he said, "Consider using virtual file loading." I'll put a PR and see if they are interested.

StephenWeatherford:
@dprice @sa-mm and all others - what's the ideal behavior you'd like to see if you could get it?  
Also, what are your normal usage patterns? Do you frequently run the same query multiple times?  Different ones?  Do you save and re-run scrapbooks?  Etc.  Thanks!

cc @fiveisprime @EricJizbaMSFT @PrashanthCorp

sa-mm:
@StephenWeatherford My ideal behavior would be for every query (collection method, cursor method, etc.) to open in the same window, which would get reused if the query was re-run. I'd also like to be able to delete the results window without having to confirm that I actually want to delete it. I'd also like to have the results pretty printed like they are with `find` and `findOne`.

[...]

PR #976, opened 2018-12-10
Reuse window for scrapbook results.

sa-mm:
In #607 @EricJizbaMSFT commented that he rarely wanted to save the results of some of the queries he made in the scrapbook, so he disliked the window that was opened and disliked that he had to select "Don't Save" in a dialog box in order to close the window. He suggested putting the results in the Output tab. @StephenWeatherford suggested using "virtual file loading." @dprice later commented that running the same aggregation via "Execute MongoDB Command" multiple times would cause multiple windows to open. I later commented that even running something like `find().limit(1)` multiple times would cause multiple result windows to open.

This PR causes the results of all commands that are not `find` and `findOne` to appear in a window called `scrapbook-result.json`. The window is reused for future commands and the user is able to close the window without being prompted to save. It accomplishes this by adding a `showFile` function to `vscodeUtils`.

I haven't done any work with VS Code extensions in the past, so I'm not sure if this heading in the right direction. Let me know if you're interested in or would like changes.

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -122,7 +122,8 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 				}
 				await editorManager.showDocument(new MongoFindOneResultEditor(database, command.collection, result), 'cosmos-result.json', { showInNextColumn: true });
 			} else {
-				await vscodeUtil.showNewFile(result, extensionPath, 'result', '.json', activeEditor.viewColumn + 1);
+				await vscodeUtil.showFile(result, extensionPath, 'scrapbook-result', '.json', activeEditor.viewColumn + 1);
+
 				await refreshTreeAfterCommand(database, command);
 			}
 		}
--- a/src/utils/vscodeUtils.ts
+++ b/src/utils/vscodeUtils.ts
@@ -37,6 +37,16 @@ export async function showNewFile(data: string, extensionPath: string, fileName:
     await writeToEditor(editor, data);
 }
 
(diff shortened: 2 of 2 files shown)


sa-mm:
@EricJizbaMSFT Thanks for taking a look and for the link regarding virtual file loading.

Do you know how people would feel about adding an option in the extension's preferences for the output channel? I find the current behavior of seeing the results in the editor a bit more readable, but I see how others would disagree.

> my biggest concern is the "Execute All" scenario in a scrapbook (You should see that as an option at the top of the scrapbook). It's pretty darn annoying to have all the extra files, but at least we're not losing information in that case.

Yup, it definitely would be nice to not lose information. While trying to troubleshoot this, I was not able to get "Execute All" to work on `master`. Here's a minimal example that I would expect to work:

```
db.orders.find({"_id": "abc"})
db.orders.find({"_id": "def"})
```

Could you provide a set of queries where it works? That might help in narrowing things down on my end.

EricJizbaMSFT:
What wasn't working for you? Were you getting an error? Those queries worked for me, although `find` and `findOne` are a little unique and have the same problem with Execute All that I described (only the last result will be shown). I think you can ignore that scenario for now (not sure if we already have an issue open) and focus on the non-find operations. For example if you run:
```
db.orders.count();
db.orders.count();
```

You should see two files opened on master:
screen shot 2019-01-07 at 10 36 53 am

sa-mm:
@EricJizbaMSFT Ah, I see it now.

> What wasn't working for you? Were you getting an error?

I thought using find/findOne would be a good test case, so when I couldn't get "Execute All" to work correctly for them, I didn't think it was related to my PR. With the examples of `db.orders.count()`, I can see the behavior you want to preserve.

EricJizbaMSFT:
Closing for now due to lack of activity

PR #996, opened 2019-01-15
webpack

StephenWeatherford:
NOTES:
1) Might be more speed-up to be obtained here by reducing contents of externalNodeModules (update: cold start: 3.5s, warm start: 300ms, installation time somewhat longer than Docker, but all good enough)
2) Might change the way we're dealing with resources in vscode-azureextensionui, but this does work and I thought it might be similar to what we'd do for third party code.  That remains to be seen.
3) Intend to move some of the webpack.config.js code to our shared dev package

### DIFF
--- a/entrypoint.js
+++ b/entrypoint.js
@@ -1,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+"use strict";
+
+// This is the extension entrypoint module, which imports extension.js, the actual extension code.
+//
+// This is in a separate file so we can properly measure extension.js load time.
+
+let perfStats = {
+    loadStartTime: Date.now(),
+    loadEndTime: undefined
+};
+
+Object.defineProperty(exports, "__esModule", { value: true });
+
+const extension = require("./dist/extension");
+
+async function activate(ctx) {
+    return await extension.activateInternal(ctx, perfStats);
+}
+
+exports.activate = activate;
+
+perfStats.loadEndTime = Date.now();
--- a/entrypoint.js
+++ b/entrypoint.js
@@ -22,6 +22,12 @@ async function activate(ctx) {
     return await extension.activateInternal(ctx, perfStats);
 }
 
+async function deactivate(ctx) {
+    return await extension.deactivateInternal(ctx, perfStats);
+}
+
+// Export as entrypoints for vscode
 exports.activate = activate;
+exports.deactivate = deactivate;
 
 perfStats.loadEndTime = Date.now();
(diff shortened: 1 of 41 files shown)


entrypoint.js
```diff
@@ -0,0 +1,27 @@
...
+
+Object.defineProperty(exports, "__esModule", { value: true });
+
+const extension = require("./dist/extension");
+
+async function activate(ctx) {
+    return await extension.activateInternal(ctx, perfStats);
+}
+
+exports.activate = activate;
```
PrashanthCorp: I don't see this being imported anywhere. Are these used by AzureTools?

StephenWeatherford: Yes.  Added comment.

PrashanthCorp: Ah, sounds good.

extension.ts
```diff
@@ -0,0 +1,34 @@
...
+export { parseDocDBConnectionString } from './src/docdb/docDBConnectionStrings';
+export { emulatorPassword } from './src/constants';
+export { getDocumentTreeItemLabel } from './src/utils/vscodeUtils';
+export { validOnTimeoutOrException } from './src/utils/inputValidation';
+export { addDatabaseToAccountConnectionString, getDatabaseNameFromConnectionString } from './src/mongo/mongoConnectionStrings';
+export { MongoCommand } from './src/mongo/MongoCommand';
+export { getAllCommandsFromText, getCommandFromTextAtLocation } from './src/mongo/MongoScrapbook';
+export { rejectOnTimeout, valueOnTimeout } from './src/utils/timeout';
+
+// These use instanceof and therefore we need to make sure we're using the same version of the bson module in the tests as in the bundle,
```
EricJizbaMSFT: Does this only apply to test code? If not, it makes me kinda nervous. What exactly is the problem with instanceof? I guess I'd rather find a fix for instanceof than single out every single use of it we run in to.

Is this at all related to the thing you did here? https://github.com/Microsoft/vscode-docker/pull/745 Either way don't forget to do that here :)

package-lock.json
```diff
@@ -126,9 +126,9 @@
 			}
```
EricJizbaMSFT: Can you run another update on this file? I still see some changes when I run `npm install` on my machine. (typescript changed from 2.9.2 to 3.2.2)

src/extensionVariables.ts
```diff
@@ -17,6 +18,7 @@ export namespace ext {
     export let context: ExtensionContext;
     export let outputChannel: OutputChannel;
     export let reporter: ITelemetryReporter;
+    export let resourcesPath = path.join(__dirname, 'resources');
```
EricJizbaMSFT: You have `resourcesPath` in both constants and extensionVariables. Do we need both? Seems kinda confusing.

extension.ts
StephenWeatherford: Yes, tests only (updated comment).  The issue is that the test code is completely separate from the extension bundle, so if we do a blind import from the tests, you get two different instances of the module and hence two difference classes with different identity.  That's also why we're importing everything from extension.ts instead of importing individual files directly from the tests like we used to, because you'd end up with two different instances of the code (and hence extensionVariables isn't a singleton, etc.).  

extension.js is separate from the test code and uses an internal webpack version of import, so it doesn't know about anything imported by code outside of it (something I wish we could change - I tried).

StephenWeatherford: > Is this at all related to the thing you did here? Microsoft/vscode-docker#745 Either way don't forget to do that here :)

No.  The other fix will be in the shared version.

src/extensionVariables.ts
StephenWeatherford: Removed from extvars

PR #1015, opened 2019-01-25
Fix scrapbook highlighting

### LINKED ISSUE #1011: No syntax hightlighting shows on scrapbook
**Version**: 1.31.0-insider (user setup)
**Node.js**: 10.2.0
**OS**: All
**Regression From**: Work well on previous release 0.9.1
**Extension build:** [20190124.12](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2452)

**Repro Steps:**
1. Create a new mongo scrapbook.
2. Connect to one mongo database then type the following into it.
image

**Expect**:
There is a syntax hightlighting for the entered code.

**Actual**:
There is no syntax highlighting.
image

StephenWeatherford:
Fixes #1011 

Need grammar folder to end up in package

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -20,5 +20,6 @@ gulp*
 webpack.config*
 stats.json
 *.tgz
-grammar
+grammar/*.md
+grammar/*.g4
 build
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -7,7 +7,6 @@ src/**
 **/*.map
 .gitignore
 .travis.yml
-package-lock.json
 tsconfig.json
 tslint.json
 .github/**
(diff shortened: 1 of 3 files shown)


.vscodeignore
```diff
@@ -20,5 +20,6 @@ gulp*
 webpack.config*
```
EricJizbaMSFT: While you're working on this file - what are your thoughts on removing `package-lock.json` as well? There's been a few times where I go back inside an old vsix to see what version of a dependency is used, but I won't be able to do that going forward since they all get webpacked together. Seems including the single json file won't be too big to ship in the vsix, right? Or is there a better way to get this info?

StephenWeatherford: Sounds like a good idea regardless, thanks.

PR #1061, opened 2019-03-07
previewHtml -> webview

### LINKED ISSUE #931: Upgrade to VS Code's webview API
Hi, I'm on the VS Code team. I noticed that your extension uses the `vscode.previewHtml` command which is deprecated and which we are actively working to remove: https://github.com/Microsoft/vscode/issues/62630

We've developed a [webview API][docs] that provides a much better developer experience and offers a number of important security and compatibility benefits over `previewHtml`. We cannot fix `previewHtml` without breaking backwards compatibility, and have instead opted to remove it. 

To ensure that your extension continues to work properly in VS Code, please try upgrading to use the new Webview API. You can find documentation on the API usage [here][docs]. Let me know if you have any questions or concerns about this migration

[docs]: https://code.visualstudio.com/docs/extensions/webview

PrashanthCorp:
- Create a panel on a request for graph explorer
- Due [restrictions on accessing local content](https://code.visualstudio.com/api/extension-guides/webview#loading-local-content), I've converted file locations into absolute paths under the vscode-resource sheme. 
- Fixes #931

### DIFF
--- a/resources/graphClient/graphClient.html
+++ b/resources/graphClient/graphClient.html
@@ -13,18 +13,18 @@
 <!--The xyzUri strings in the scripts' and icon's src field will be replaced by the approrpriate strings at runtime when this html file is read -->
 
 <head>
-    <link rel="stylesheet" type="text/css" href="BASEURI/resources/graphClient/graphClient.css">
+    <link rel="stylesheet" type="text/css" href="$BASEURI/resources/graphClient/graphClient.css">
     </link>
 
-    <script src='BASEURI/dist/node_modules/d3/d3.js' charset="UTF-8"></script>
+    <script src='$BASEURI/dist/node_modules/d3/d3.js' charset="UTF-8"></script>
 
     <script>window.exports = {};</script>
-    <script src="BASEURI/dist/node_modules/socket.io-client/dist/socket.io.js"></script>
+    <script src="$BASEURI/dist/node_modules/socket.io-client/dist/socket.io.js"></script>
     <script>
         io = exports.io;
     </script>
 
-    <script src="BASEURI/dist/graphClient.js"></script>
+    <script src="$BASEURI/dist/graphClient.js"></script>
 </head>
 
 <!-- Possible states (set on #states - see "type State" in graphClient.ts)
@@ -41,7 +41,7 @@
     <div id="states" class="state-initial">
         </script>
         <header>
-            <img id="cosmos" src="BASEURI/resources/cosmos.png">
+            <img id="cosmos" src="$BASEURI/resources/cosmos.png">
             <h1 id="title">
                 &nbsp;
                 <!-- placeholder -->
@@ -99,7 +99,7 @@
 
         <script>
             // "clientport" will be replaced with the actual port when the file is read
-            var graphClient = new GraphClient(clientPort);
+            var graphClient = new GraphClient($CLIENTPORT);
 
             window.onload = () => {
                 graphClient.getPageState();
--- a/resources/graphClient/graphClient.html
+++ b/resources/graphClient/graphClient.html
@@ -13,18 +13,18 @@
 <!--The xyzUri strings in the scripts' and icon's src field will be replaced by the approrpriate strings at runtime when this html file is read -->
 
 <head>
-    <link rel="stylesheet" type="text/css" href="cssUri">
+    <link rel="stylesheet" type="text/css" href="BASEURI/resources/graphClient/graphClient.css">
     </link>
 
-    <script src='d3Uri' charset="UTF-8"></script>
+    <script src='BASEURI/dist/node_modules/d3/d3.js' charset="UTF-8"></script>
 
     <script>window.exports = {};</script>
-    <script src="socketUri"></script>
+    <script src="BASEURI/dist/node_modules/socket.io-client/dist/socket.io.js"></script>
     <script>
         io = exports.io;
     </script>
 
-    <script src="graphClientUri"></script>
+    <script src="BASEURI/dist/graphClient.js"></script>
 </head>
 
 <!-- Possible states (set on #states - see "type State" in graphClient.ts)
@@ -41,7 +41,7 @@
     <div id="states" class="state-initial">
         </script>
         <header>
-            <img id="cosmos" src="cosmosIconUri">
+            <img id="cosmos" src="BASEURI/resources/cosmos.png">
             <h1 id="title">
                 &nbsp;
                 <!-- placeholder -->
@@ -98,7 +98,7 @@
         </div>
 
         <script>
-            // the variable will be replaced with the actual port when the file is read
+            // "clientport" will be replaced with the actual port when the file is read
             var graphClient = new GraphClient(clientPort);
 
             window.onload = () => {
--- a/resources/graphClient/graphClient.html
+++ b/resources/graphClient/graphClient.html
@@ -10,19 +10,21 @@
     This is the HTML for exploring graphs, and is hosted inside a WebView in a VS HTML preview window
 -->
 
+<!--The xyzUri strings in the scripts' and icon's src field will be replaced by the approrpriate strings at runtime when this html file is read -->
+
 <head>
-    <link rel="stylesheet" type="text/css" href="graphClient.css">
+    <link rel="stylesheet" type="text/css" href="cssUri">
     </link>
 
-    <script src='../../dist/node_modules/d3/d3.js' charset="UTF-8"></script>
+    <script src='d3Uri' charset="UTF-8"></script>
 
     <script>window.exports = {};</script>
-    <script src="../../dist/node_modules/socket.io-client/dist/socket.io.js"></script>
+    <script src="socketUri"></script>
     <script>
         io = exports.io;
     </script>
 
-    <script src="../../dist/graphClient.js"></script>
+    <script src="graphClientUri"></script>
 </head>
 
 <!-- Possible states (set on #states - see "type State" in graphClient.ts)
@@ -39,7 +41,7 @@
     <div id="states" class="state-initial">
         </script>
         <header>
-            <img id="cosmos" src="../../resources/cosmos.png">
+            <img id="cosmos" src="cosmosIconUri">
             <h1 id="title">
                 &nbsp;
                 <!-- placeholder -->
@@ -96,9 +98,8 @@
         </div>
 
         <script>
-            // Retrieve port from query string
-            var [, port] = window.location.search.match(/port=(.*)/);
-            var graphClient = new GraphClient(parseInt(port));
+            // the variable will be replaced with the actual port when the file is read
+            var graphClient = new GraphClient(clientPort);
 
             window.onload = () => {
                 graphClient.getPageState();
--- a/src/graph/GraphViewsManager.ts
+++ b/src/graph/GraphViewsManager.ts
@@ -26,10 +26,10 @@ interface IGraphClientHtmlPaths {
 export class GraphViewsManager implements IServerProvider { //Graphviews Panel
   private _lastServerId = 0;
 
-  // One server (and one HTML view) per graph, as represented by unique configurations
-  private _servers = new Map<number, GraphViewServer>(); // map of id -> map
-  private _panels = new Map<number, vscode.WebviewPanel>(); // map of id -> map
-  private _panelViewType: string = "GraphExplorer";
+  // One server (and one webview panel) per graph, as represented by unique configurations
+  private readonly _servers = new Map<number, GraphViewServer>(); // map of id -> server
+  private readonly _panels = new Map<number, vscode.WebviewPanel>(); // map of id -> webview panel
+  private readonly _panelViewType: string = "CosmosDB.GraphExplorer";
 
   constructor(private _context: vscode.ExtensionContext) {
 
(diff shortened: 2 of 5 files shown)


src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+  cosmosIconUri: string;
+  cssUri: string;
+}
+
+export class GraphViewsManager implements IServerProvider { //Graphviews Panel
   private _lastServerId = 0;
 
   // One server (and one HTML view) per graph, as represented by unique configurations
   private _servers = new Map<number, GraphViewServer>(); // map of id -> map
+  private _panels = new Map<number, vscode.WebviewPanel>(); // map of id -> map
```
StephenWeatherford: map of id -> WebViewPanel

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+  graphClientUri: string;
+  cosmosIconUri: string;
+  cssUri: string;
+}
+
+export class GraphViewsManager implements IServerProvider { //Graphviews Panel
   private _lastServerId = 0;
 
   // One server (and one HTML view) per graph, as represented by unique configurations
   private _servers = new Map<number, GraphViewServer>(); // map of id -> map
```
StephenWeatherford: Should have been:
// map of id -> graph view server

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+  private _panels = new Map<number, vscode.WebviewPanel>(); // map of id -> map
+  private _panelViewType: string = "GraphExplorer";
```
StephenWeatherford: Let's make it more specific - e.g. add our extension prefix.

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
-
-      // Add server ID to the URL so that GraphViewDocumentContentProvider knows which port to use in the HTML
-      let serverUri = previewBaseUri + id.toString();
-      await vscode.commands.executeCommand('vscode.previewHtml', vscode.Uri.parse(serverUri), vscode.ViewColumn.One, tabTitle);
-    } catch (error) {
-      vscode.window.showErrorMessage(error.message || error);
+    let id = await this.getOrCreateServer(config);
+
+    let existingPanel: vscode.WebviewPanel;
+    this._panels.forEach((p, key) => {
```
StephenWeatherford: You can use this._panels.find()

src/graph/GraphViewsManager.ts
StephenWeatherford: should be a const

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+    const column = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.viewColumn : undefined;
+    const options: vscode.WebviewOptions & vscode.WebviewPanelOptions = {
+      enableScripts: true,
+      enableCommandUris: true,
+      enableFindWidget: true,
+      retainContextWhenHidden: true,
+      localResourceRoots: [vscode.Uri.file(this._context.extensionPath)]
+    };
+    const panel = vscode.window.createWebviewPanel(this._panelViewType, tabTitle, { viewColumn: column, preserveFocus: true }, options);
+    let documentProvider = new GraphViewDocumentContentProvider(this, this._context);
```
StephenWeatherford: Let's rename GraphViewDocumentContentProvider

src/graph/GraphViewsManager.ts
```diff
@@ -69,41 +104,180 @@ export class GraphViewsManager implements IServerProvider {
...
-      <style>
-        body {
-          padding: 0;
-          margin: 0;
-        }
-      </style>
-      <body>
-        <iframe src="file://${clientHtmlPath}?port=${server.port}" style="width: 100%; height: 100%; position: absolute; padding: 0; margin: 0; border: none"></iframe>
-      </body>
+
```
StephenWeatherford: I assume you'll move this back into the separate HTML file...

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+      if (key === id) {
+        existingPanel = p;
+      }
+    });
+    if (existingPanel) {
+      try { //existing panel might have been disposed
+        existingPanel.reveal();
+        return;
+      } catch (_e) {
+        //empty block
```
StephenWeatherford: You should only catch disposed exception

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,80 @@
...
+    const options: vscode.WebviewOptions & vscode.WebviewPanelOptions = {
+      enableScripts: true,
+      enableCommandUris: true,
+      enableFindWidget: true,
+      retainContextWhenHidden: true,
+      localResourceRoots: [vscode.Uri.file(this._context.extensionPath)]
+    };
+    const panel = vscode.window.createWebviewPanel(this._panelViewType, tabTitle, { viewColumn: column, preserveFocus: true }, options);
+    let documentProvider = new GraphViewDocumentContentProvider(this, this._context);
+    panel.webview.html = documentProvider.provideHtmlContent(vscode.Uri.parse(id.toString()));
```
StephenWeatherford: What does vscode.Uri.parse(id.toString()) return?

src/graph/GraphViewsManager.ts
```diff
@@ -69,41 +104,180 @@ export class GraphViewsManager implements IServerProvider {
...
 
-class GraphViewDocumentContentProvider implements vscode.TextDocumentContentProvider {
+class GraphViewDocumentContentProvider {
   public onDidChange?: vscode.Event<vscode.Uri>;
 
-  public constructor(private _serverProvider: IServerProvider) { }
+  public constructor(private _serverProvider: IServerProvider, private _context: vscode.ExtensionContext) { }
 
-  public provideTextDocumentContent(uri: vscode.Uri, _token: vscode.CancellationToken): vscode.ProviderResult<string> {
+  public provideHtmlContent(uri: vscode.Uri): string {
```
StephenWeatherford: I'd pass in the id rather than a weird URI

StephenWeatherford: Before, I had to communicate via a URI, that's no longer the case.

src/graph/GraphViewsManager.ts
```diff
@@ -69,41 +104,180 @@ export class GraphViewsManager implements IServerProvider {
...
     // Figure out which client to attach this to
     // tslint:disable-next-line:no-single-line-block-comment
-    let serverId = parseInt(uri.path.slice(1) /* remove '/' from beginning */, 10);
+    let serverId = parseInt(uri.path, 10);
     console.assert(serverId > 0);
     let server = this._serverProvider.findServerById(serverId);
     if (server) {
-      let clientHtmlPath = path.join(resourcesPath, 'graphClient', 'graphClient.html');
-      console.assert(fs.existsSync(clientHtmlPath), `Couldn't find ${clientHtmlPath}`);
+      let options: IGraphClientHtmlPaths = {
```
StephenWeatherford: nit: it be easier just to pass in the base path for these as a single variable?  I don't think web browser care whether the '/' and '\' are mismatched.  Could be wrong, though.

src/graph/GraphViewsManager.ts
```diff
@@ -69,41 +104,180 @@ export class GraphViewsManager implements IServerProvider {
...
+        d3Uri: this._getVscodeResourceUri(['dist', 'node_modules', 'd3', 'd3.js']),
+        socketUri: this._getVscodeResourceUri(['dist', 'node_modules', 'socket.io-client', 'dist', 'socket.io.js']),
+        graphClientUri: this._getVscodeResourceUri(['dist', 'graphClient.js']),
+        cosmosIconUri: this._getVscodeResourceUri(['resources', 'cosmos.png']),
+        cssUri: this._getVscodeResourceUri(['resources', 'graphClient', 'graphClient.css'])
+      };
+      return this._graphClientHtmlAsString(server.port, options);
+    }
+
+    return "This resource is no longer available.";
```
StephenWeatherford: throw??

src/graph/GraphViewsManager.ts
```diff
@@ -69,41 +104,180 @@ export class GraphViewsManager implements IServerProvider {
...
+  }
+
+  private _getVscodeResourceUri(direcotryList: string[]): string {
```
StephenWeatherford: direcotryList misspelled

src/graph/GraphViewsManager.ts
PrashanthCorp: I don't think Maps have a `find` function, and that's why getOrCreateServer too does a `forEach`.

src/graph/GraphViewsManager.ts
PrashanthCorp: I agree. I found it harder to edit myself. How would the `options.uri` strings work once the HTML is read from disk though?

src/graph/GraphViewsManager.ts
StephenWeatherford: Map.has()/get().  Wouldn't be very useful as a map without that.

getOrCreateServer needs to find via areConfigsEqual, not key

StephenWeatherford: Isn't this just findPanelById?

resources/graphClient/graphClient.html
```diff
@@ -96,9 +98,8 @@ <h2>Debug log</h2>
         </div>
 
         <script>
-            // Retrieve port from query string
-            var [, port] = window.location.search.match(/port=(.*)/);
-            var graphClient = new GraphClient(parseInt(port));
+            // the variable will be replaced with the actual port when the file is read
```
StephenWeatherford: which variable?  Don't understand the comment

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,82 @@
...
 import * as path from "path";
 import * as vscode from 'vscode';
-import { resourcesPath } from "../constants";
+import { resourcesPath } from '../constants';
 import { areConfigsEqual, GraphConfiguration } from './GraphConfiguration';
 import { GraphViewServer } from './GraphViewServer';
 
-const scheme = "vscode-cosmosdb-graphresults";
-const previewBaseUri = scheme + '://results/';
+// const scheme = "vscode-cosmosdb-graphresults";
```
StephenWeatherford: remove

resources/graphClient/graphClient.html
```diff
@@ -10,19 +10,21 @@
     This is the HTML for exploring graphs, and is hosted inside a WebView in a VS HTML preview window
 -->
 
+<!--The xyzUri strings in the scripts' and icon's src field will be replaced by the approrpriate strings at runtime when this html file is read -->
+
 <head>
-    <link rel="stylesheet" type="text/css" href="graphClient.css">
+    <link rel="stylesheet" type="text/css" href="cssUri">
```
StephenWeatherford: I'd prefer something that stands out more as being something to get replaced, like $CSSURI$.

It also seems like it would be easier to just have a single $BASEURL$ replacement and use src="$BASEURL/dist/node_modules/d3/d3.js" etc.

PrashanthCorp: I didn't take that route since it looked like the paths would work differently on Windows : 
image


But it looks like it doesn't matter whether '\\' or '/' is used. Your approach looks easier to maintain.

src/graph/GraphViewsManager.ts
StephenWeatherford: That's what I was saying - I don't think Chrome cares.

resources/graphClient/graphClient.html
PrashanthCorp: Looks like it works.

StephenWeatherford:
Is this ready for final review?

PrashanthCorp:
Yup

resources/graphClient/graphClient.html
```diff
@@ -10,19 +10,21 @@
     This is the HTML for exploring graphs, and is hosted inside a WebView in a VS HTML preview window
 -->
 
+<!--The xyzUri strings in the scripts' and icon's src field will be replaced by the approrpriate strings at runtime when this html file is read -->
+
 <head>
-    <link rel="stylesheet" type="text/css" href="graphClient.css">
+    <link rel="stylesheet" type="text/css" href="BASEURI/resources/graphClient/graphClient.css">
```
StephenWeatherford: nit: $BASEURI would be more easily recognized as a replacement string

src/graph/GraphViewsManager.ts
```diff
@@ -3,45 +3,62 @@
...
   ): Promise<void> {
-    try {
-      let id = await this.getOrCreateServer(config);
-
-      // Add server ID to the URL so that GraphViewDocumentContentProvider knows which port to use in the HTML
-      let serverUri = previewBaseUri + id.toString();
-      await vscode.commands.executeCommand('vscode.previewHtml', vscode.Uri.parse(serverUri), vscode.ViewColumn.One, tabTitle);
-    } catch (error) {
-      vscode.window.showErrorMessage(error.message || error);
+    let id = await this.getOrCreateServer(config);
```
StephenWeatherford: What happens if this throws?  Is something catching it?

resources/graphClient/graphClient.html
```diff
@@ -96,9 +98,8 @@ <h2>Debug log</h2>
         </div>
 
         <script>
-            // Retrieve port from query string
-            var [, port] = window.location.search.match(/port=(.*)/);
-            var graphClient = new GraphClient(parseInt(port));
+            // "clientport" will be replaced with the actual port when the file is read
+            var graphClient = new GraphClient(clientPort);
```
StephenWeatherford: Likewise, suggest $CLIENTPORT

resources/graphClient/graphClient.html
```diff
@@ -10,19 +10,21 @@
     This is the HTML for exploring graphs, and is hosted inside a WebView in a VS HTML preview window
 -->
```
StephenWeatherford: You might want to create a VSIX and test it with the VSIX to make sure webpack is working with your changes.

PrashanthCorp: Done. :-)

PrashanthCorp:
I ran the onenote tests on this branch, and looks like there aren't any regressions. There's one issue I've PM'ed @StephenWeatherford about. Merging this.

Issue #998, opened 2019-01-15
Collection name must be a string.


camerone:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: cosmosDB.executeMongoCommand
Error type: Error
Error Message: collection name must be a String

Version: 0.9.0
OS: darwin

StephenWeatherford:
@camerone We need more information on what you were doing in order to analyze the issue.  Please fill in the repro steps.  Thanks.

StephenWeatherford:
Here's a way to repro:

db.find()

PrashanthCorp:
Will be fixed by #1089

nturinski:
This is now fixed with v0.12.0

PR #1089, opened 2019-04-15
Check if command is chained before making API call

### LINKED ISSUE #981: Error when using hyphenated collection names
When I try to perform a find operation in a collection with a hyphenated name I get this error.

## Repro steps:

1. Open a scratchbook
2. `db.getCollection('my-collection').find({})`
3. Execute

## Result:

Action: cosmosDB.executeMongoCommand
Error type: Error
Error Message: collection name must be a String

Version: 0.9.1
OS: darwin

PrashanthCorp:
Fixes #981, #998 
Fix part of #1096.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "vscode-cosmosdb",
-	"version": "0.10.1-alpha",
+	"version": "0.10.1",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
@@ -3611,7 +3611,8 @@
 				"ansi-regex": {
 					"version": "2.1.1",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"aproba": {
 					"version": "1.2.0",
@@ -3632,12 +3633,14 @@
 				"balanced-match": {
 					"version": "1.0.0",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"brace-expansion": {
 					"version": "1.1.11",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"balanced-match": "^1.0.0",
 						"concat-map": "0.0.1"
@@ -3652,17 +3655,20 @@
 				"code-point-at": {
 					"version": "1.1.0",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"concat-map": {
 					"version": "0.0.1",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"console-control-strings": {
 					"version": "1.1.0",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"core-util-is": {
 					"version": "1.0.2",
@@ -3779,7 +3785,8 @@
 				"inherits": {
 					"version": "2.0.3",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"ini": {
 					"version": "1.3.5",
@@ -3791,6 +3798,7 @@
 					"version": "1.0.0",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"number-is-nan": "^1.0.0"
 					}
@@ -3805,6 +3813,7 @@
 					"version": "3.0.4",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"brace-expansion": "^1.1.7"
 					}
@@ -3812,12 +3821,14 @@
 				"minimist": {
 					"version": "0.0.8",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"minipass": {
 					"version": "2.2.4",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"safe-buffer": "^5.1.1",
 						"yallist": "^3.0.0"
@@ -3836,6 +3847,7 @@
 					"version": "0.5.1",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"minimist": "0.0.8"
 					}
@@ -3916,7 +3928,8 @@
 				"number-is-nan": {
 					"version": "1.0.1",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"object-assign": {
 					"version": "4.1.1",
@@ -3928,6 +3941,7 @@
 					"version": "1.4.0",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"wrappy": "1"
 					}
@@ -4013,7 +4027,8 @@
 				"safe-buffer": {
 					"version": "5.1.1",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"safer-buffer": {
 					"version": "2.1.2",
@@ -4049,6 +4064,7 @@
 					"version": "1.0.2",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"code-point-at": "^1.0.0",
 						"is-fullwidth-code-point": "^1.0.0",
@@ -4068,6 +4084,7 @@
 					"version": "3.0.1",
 					"bundled": true,
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"ansi-regex": "^2.0.0"
 					}
@@ -4111,12 +4128,14 @@
 				"wrappy": {
 					"version": "1.0.2",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"yallist": {
 					"version": "3.0.2",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
 				}
 			}
 		},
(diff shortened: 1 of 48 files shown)


package-lock.json
```diff
@@ -9481,25 +9555,18 @@
 			}
 		},
 		"vscode": {
-			"version": "1.1.26",
-			"resolved": "https://registry.npmjs.org/vscode/-/vscode-1.1.26.tgz",
-			"integrity": "sha512-z1Nf5J38gjUFbuDCbJHPN6OJ//5EG+e/yHlh6ERxj/U9B2Qc3aiHaFr38/fee/GGnxvRw/XegLMOG+UJwKi/Qg==",
+			"version": "1.1.33",
```
StephenWeatherford: Where are these changes coming from?

src/mongo/MongoScrapbook.ts
```diff
@@ -112,16 +112,13 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
...
-		if (command.name === 'find') {
+		if (command.name === 'find' && !!command.collection && !command.chained) {
 			await editorManager.showDocument(new MongoFindResultEditor(database, command), 'cosmos-result.json', { showInNextColumn: true });
 		} else {
 			const result = await database.executeCommand(command, context);
-			if (command.name === 'findOne') {
-				if (result === "null") {
-					throw new Error(`Could not find any documents`);
-				}
+			if (command.name === 'findOne' && !!command.collection) {
```
StephenWeatherford: Don't you need that same check for findOne?

test/mongoGetCommand.test.ts
```diff
@@ -934,6 +932,23 @@ suite("scrapbook parsing Tests", () => {
         });
     });
 
+    test("Chained command where final function call is recognized. Ensure execution is sent to shell - test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/981", () => {
+        testParse(`db.getCollection('my-collection').find({});`, {
```
StephenWeatherford: What about this checks that it gets sent to the shell?

PrashanthCorp: The command is chained - MongoFindResultEditor cannot handle this.

PrashanthCorp: > What about this checks that it gets sent to the shell?

The collection name is undefined. A command like `db.abc.find()` would have the collection name be abc.

package-lock.json
PrashanthCorp: I'm assuming they come from an npm install after I merge from master

src/mongo/editors/MongoFindOneResultEditor.ts
```diff
@@ -18,6 +18,9 @@ export class MongoFindOneResultEditor implements ICosmosEditor<IMongoDocument> {
     constructor(databaseNode: MongoDatabaseTreeItem, collectionName: string, data: string) {
         this._databaseNode = databaseNode;
         this._collectionName = collectionName;
+        if (data === "null") {
```
EricJizbaMSFT: I don't understand this check. Is it really checking for the _string_ null? When can this happen?

test/mongoGetCommand.test.ts
```diff
@@ -934,6 +932,23 @@ suite("scrapbook parsing Tests", () => {
...
+    test("Chained command where final function call is recognized. Ensure execution is sent to shell (the command has not collection name) - test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/981", () => {
+        testParse(`db.getCollection('my-collection').find({});`, {
+            collection: undefined,
+            name: "find",
+            args: ["my-collection", {}]
+        });
+    });
+
+    test("Chained command where final function call is recognized as findOne. Ensure execution is sent to shell (the command has not collection name) - test user issues: https://github.com/Microsoft/vscode-cosmosdb/issues/981", () => {
+        testParse(`db.getCollection('my-collection').find({});`, {
```
EricJizbaMSFT: Is this supposed to be `findOne`? It looks like a dupe of the above test.

package-lock.json
EricJizbaMSFT: Either way they shouldn't be a part of a PR where you didn't change "package.json". You either need to merge this branch with master or clean up somehow

test/mongoGetCommand.test.ts
EricJizbaMSFT: I agree with Stephen that the name of this test isn't accurate. I would just name it something like `find command for collection with hyphens in name, from #981`

If you want to create a test that represents the name you currently have, it should somehow actually verify against the shell. But I don't think that's worth doing until after https://github.com/microsoft/vscode-cosmosdb/pull/1105

src/mongo/editors/MongoFindOneResultEditor.ts
PrashanthCorp: Yes. The variable name `data` is misleading - `data` is a _string_  of the query result (returned from `executeCommand`), and hence line 25 has a `EJSON.parse`. 

It is the string "null" when there are no documents matching the filter. :)

PrashanthCorp: I didn't see the need to rename the variable since the type annotations explain the expected handling.

src/mongo/MongoScrapbook.ts
```diff
@@ -348,7 +345,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 		}
 		let argumentContextArray: mongoParser.ArgumentContext[] = filterType(argumentsToken.children, mongoParser.ArgumentContext);
 
-		let functionMap = { "ObjectId": this.objectIdToObject, "ISODate": this.dateToObject, "Date": this.dateToObject };
```
EricJizbaMSFT: Why did this change? This doesn't seem related to https://github.com/microsoft/vscode-cosmosdb/issues/981. If you're fixing a different bug I would prefer a separate PR or at least mention that bug in the PR description.

src/mongo/editors/MongoFindOneResultEditor.ts
EricJizbaMSFT: Suggest adding a comment. A string of "null" is very uncommon, plus this repo doesn't have strictNullChecks on so the types aren't super accurate. Someone could easily mess this up if they didn't know the full story. Just add something like:
```
// Intentionally checking against string of "null", which can be returned by mongo shell
```

src/mongo/MongoScrapbook.ts
```diff
@@ -221,7 +218,7 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 			name: '',
 			arguments: [],
 			argumentObjects: [],
-			chained: funcCallCount > 1 ? true : false
+			sendToShell: funcCallCount > 1 ? true : false
```
EricJizbaMSFT: I don't think it's accurate to rename `chained` to `sendToShell`. We use the shell as a catch-all for several different unsupported cases, one of which happens to be `chained`. I would prefer to leave the name as-is and wait for https://github.com/microsoft/vscode-cosmosdb/pull/1105 to do any refactoring related to this

src/mongo/MongoScrapbook.ts
PrashanthCorp: The changes there would depend on this PR (doing it independently complicates the code).  I chose to address the issue here. I mentioned it in the commit message.  Will add to PR description.

src/mongo/MongoScrapbook.ts
PrashanthCorp: This is related to the fix for #1096. When both bugs are fixed (chained commands and the Date), having two separate flags for the same action of sending to the shell is redundant when on flag would do.

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -97,7 +97,7 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
 	}
 
 	async executeCommand(command: MongoCommand, context: IActionContext): Promise<string> {
-		if (command.collection) {
+		if (command.collection && !command.sendToShell) {
```
EricJizbaMSFT: If you want to name it `sendToShell` - I think we need to refactor this method. Something like this:
```
if (!command.sendToShell) {
	if (command.collection) {
		let db = await this.getDb();
		const collection = db.collection(command.collection);
		if (collection) {
			const collectionTreeItem = new MongoCollectionTreeItem(this, collection, command.arguments);
			const result = await collectionTreeItem.executeCommand(command.name, command.arguments);
			if (result) {
				return result;
			}
		}
	} else if (command.name === 'createCollection') {
		return withProgress(this.createCollection(stripQuotes(command.arguments.join(','))).then(() => JSON.stringify({ 'Created': 'Ok' })), 'Creating collection');
	}
}

return withProgress(this.executeCommandInShell(command, context), 'Executing command');
```

My problem with the current name is that there are cases where sendToShell could be `true`, but it's not actually sent to the shell. With my above code, that's guaranteed never to happen.

src/mongo/MongoScrapbook.ts
EricJizbaMSFT: see other comment near executeCommand

src/mongo/MongoScrapbook.ts
```diff
@@ -112,16 +112,13 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 			throw new Error(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
 		}
 
-		if (command.name === 'find') {
+		if (command.name === 'find' && !!command.collection && !command.sendToShell) {
```
EricJizbaMSFT: Can we move this code directly into `database.executeCommand`

```
if (command.name === 'find' && !!command.collection && !command.sendToShell) {
	await editorManager.showDocument(context, new MongoFindResultEditor(database, command), 'cosmos-result.json', { showInNextColumn: true });
} else {
	const result = await database.executeCommand(command, context);
	if (command.name === 'findOne' && !!command.collection) {
		await editorManager.showDocument(context, new MongoFindOneResultEditor(database, command.collection, result), 'cosmos-result.json', { showInNextColumn: true });
	} else { // failover also includes chained commands
		await vscodeUtil.showNewFile(result, 'result', '.json', activeEditor.viewColumn + 1);
		await refreshTreeAfterCommand(database, command, context);
	}
}
```

I just think there's a lot of duplicate checks here and I'm concerned we're missing a case of `command.sendToShell`. If you move it all into `database.executeCommand` and integrate it correctly, I think you'll only ever have to check `command.sendToShell` once

EricJizbaMSFT:
Closing due to lack of activity. Feel free to re-open once you're actively working on this

PR #1094, opened 2019-04-24
Change usage : opn -> vscode.open

PrashanthCorp:
Not wrapped in a func call since it's only used once.


ThirdPartyNotice.html
```diff
@@ -5627,15 +5627,6 @@
 
 The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
 
-THE SOFTWARE IS PROVIDED &quot;AS IS&quot;, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
EricJizbaMSFT: @PrashanthCorp did you manually edit this file? Remember this is an auto-generated file, so you'll have to re-generate it before releasing. There's instructions in the team OneNote

PR #1102, opened 2019-05-01
[Post-Build] Make end of result sentinel less ambiguous

StephenWeatherford:
Fixes #114

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,7 +14,7 @@
 			"sourceMaps": true,
 			// outFiles is used for locating generated JavaScript files, see https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_source-maps
 			"outFiles": [
-				"${workspaceFolder}/**/*.js"
+				"${workspaceFolder}/out/**/*.js"
 			],
 			"preLaunchTask": "npm: webpack",
 			"env": {
@@ -54,7 +54,7 @@
 			"sourceMaps": true,
 			// outFiles is used for locating generated JavaScript files, see https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_source-maps
 			"outFiles": [
-				"${workspaceFolder}/dist/**/*.js"
+				"${workspaceFolder}/out/**/*.js"
 			],
 			"preLaunchTask": "npm: webpack",
 			"env": {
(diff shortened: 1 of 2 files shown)


src/mongo/shell.ts
```diff
@@ -9,8 +9,11 @@ import { EventEmitter, window } from 'vscode';
 import { ext } from '../extensionVariables';
 import { IDisposable, toDisposable } from '../utils/vscodeUtils';
 
-export class Shell {
+// This is used at the end of each command we send to the console. When we get this string back,
+// we know we've reached the end of that command's result.
+const endOfDataBaseSentinel: string = '$EOD$';
```
PrashanthCorp: Just pointing out that DataBase still reads as Database. I suggest root for less confusion maybe? :P

.vscode/launch.json
```diff
@@ -54,7 +54,7 @@
 			"sourceMaps": true,
 			// outFiles is used for locating generated JavaScript files, see https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_source-maps
 			"outFiles": [
-				"${workspaceFolder}/dist/**/*.js"
+				"${workspaceFolder}/out/**/*.js"
```
PrashanthCorp: Is this change tangential to the PR?

StephenWeatherford: Yes, don't worry about it, I'll fix it when merging the webpack PR.

src/mongo/shell.ts
```diff
@@ -54,9 +57,9 @@ export class Shell {
 		let buffers: string[] = [];
 		on(this.mongoShell.stdout, 'data', b => {
 			let data: string = b.toString();
-			const delimitter = `${this.executionId}${os.EOL}`;
-			if (data.endsWith(delimitter)) {
-				const result = buffers.join('') + data.substring(0, data.length - delimitter.length);
+			const endOfDataSentinel = `${endOfDataBaseSentinel}${this.executionId}${os.EOL}`;
+			if (data.endsWith(endOfDataSentinel)) {
```
EricJizbaMSFT: I'm fine with this change, but we honestly need to completely refactor this file at some point (this code was inherited from the old mongo repo). If we copy our [cpUtils code](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/utils/cpUtils.ts#L33) from another repo I think we can completely avoid the use of a sentinel. I might take a look after the upcoming Remote bug-fix release.

StephenWeatherford: Um, the refactor already has a PR.

StephenWeatherford: I just wanted this separate because it was specifically related to a bug and I did it before the rewrite.

EricJizbaMSFT: Sorry for missing that lol - filtered out a few "Post build" emails last week in the lead up to build

Issue #1092, opened 2019-04-22
update() timeout BUT find() ok


futurenz:
simple query
db.inventory.updateOne( {rating:{$exists:true}},{$unset:{rating:''}});
causes this error but any find() query works. MongoDB Shell through Command Prompt or Power Shell work without issue
I see several other very similar issues have been logged - I am in New Zealand and the cluster is listed as in N.Virginia - is this more a distance thing?

1.
2.

Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "use test"

Version: 0.10.1
OS: win32
VS Code Version: 1.33.1

PrashanthCorp:
Hi @futurenz , 
The error implies it was unable to set up the shell to run the command. Can you try a command like `db.inventory.findOneAndUpdate(<fields>)` and let us know if the issue exists? 
Also, please let us know if these commands run properly : findOne, count, insertOne

Again, thanks for using our extension. 

--Prashanth

futurenz:
Hi Prashanth,
findOneAndUpdate fails
findOne is good
count is good
insertOne is good

Thanks for the prompt response.
regards

------ Original Message ------

StephenWeatherford:
@futurenz I've made a fix that kept us from showing the real error in certain cases when the Mongo shell was used for a query (which would be the case for findOneAndUpdate).  Hopefully you will now get the correct error message (after next release) and be able to figure out the problem.  Please let us know if it doesn't.  Thanks!

StephenWeatherford:
Fixed by #1162

Issue #1071, opened 2019-03-15
Timeout when executing an aggregate on mongoDB 


lannyfrank:
Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "use database_name"

Version: 0.10.0
OS: win32

This error occurres when using this kind of connectionString
mongodb+srv://<userName>:<password>@host_service?retryWrites=true
(no database has been specified after the service)

while using this kind of connectionString works fine
mongodb+srv://<userName>:<password>@host_service/**database_name**?retryWrites=true

using this last kind of connectionString, you will see only the specified database, while using the first one, you will see every database allowed to the user...

PrashanthCorp:
This error does not seem specific to the aggregate command. 

cc: @StephenWeatherford

StephenWeatherford:
Would you mind quickly trying an experimental version that may report better errors? It's easy to uninstall afterwards via the extensions view in vs code.  See instructions here: https://github.com/Microsoft/vscode-cosmosdb/issues/852#issuecomment-479734886.

Thanks!
Stephen

StephenWeatherford:
Fixed by #1162

Hopefully with this fix, you will now see the correct error message and be able to figure out the problem.  Thanks!

Issue #988, opened 2019-01-04
Fail to run 'db.version();' with an error (Windows)


v-dantch:
**OS:** Win10
**Node.js:** V 10.14.2
**Build Version:** [20190103.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=1805&view=results)

**Repro Steps:**
1. Install the [mongo shell]( https://www.mongodb.com/download-center#community) and set the 'mongo.shell.path' in VS Code.
2. New a scrapbook and run 'db.version();'.
3. Check the result.

**Expect:**
Return a version.

**Actual:**
Pop up an error.
![image](https://user-images.githubusercontent.com/34729092/50677691-31207580-1036-11e9-997d-03204b00dcfa.png)

StephenWeatherford:
Can you try this again?  I can't repro.  Are you sure the mongo.shell.path is correct?  You could also try removing this setting and trying again, and selecting the Browse option to find the .exe directly.  Thanks.

v-dantch:
Hi Stephen, today we verified this issue on the [20190124.8](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2453) and it only reproduces on Windows.

StephenWeatherford:
Still can't repro, but this is an issue some customers are reporting, so I'd love to be able to repro this. Some questions:
1) What kind of database are you connected to? Emulator? Local? Cosmos?
2) Does it work if you try it against a different kind of database?

Also, could you please try installing the extension from https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2476 and see what results you get?  Thx.

StephenWeatherford:
See also #852.

v-dantch:
Hi Stephen, please get the details from below:
-- This issue reproduces for all databases of Cosmos, Local, Emulator.

And, I tried this issue on the provided build and latest build:
--Provided build [20180125.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2476): Pop up an error: login failed. Though I have signed in successfully.
--The latest build [20190127.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2509&view=results): Also repro 'Timed out...'

v-dantch:
Please get more from the screenshot:
![image](https://user-images.githubusercontent.com/34729092/51820306-2601e380-228a-11e9-868c-85e7ecae5a2a.png)

[...]

Issue #852, opened 2018-10-08
trying to run a mongo query fails with Timeout


smv-orrin:
Repro steps:
<Enter steps to reproduce issue>

Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "db.ingestion_queue.aggregate([{$match:{$and:[{"date_intended":{$gte:ISODate("2018-10-04")}},{"date_intended":{$lt:ISODate("2018-10-09")}}]}},{$project:{"_id":0,"date_intended":{$dateToString:{"date":"$date_intended","format":"%Y-%m-%d"}},"hr":{$hour:"$date_time"}}},{$group:{"_id":{"dt":"$date_intended","hr":"$hr"},"hourCt":{$sum:1}}},{$project:{"_id":0,"dt":"$_id.dt","hr":"$_id.hr","hourCt":"$hourCt"}},{$sort:{"dt":1,"hr":1}}])"

Version: 0.8.0
OS: win32

StephenWeatherford:
@smv-orrin Is it possible it really is hitting a timeout? (We have a small class of bugs we're working on where we're getting an incorrect time-out...  From quick inspection this bug doesn't seem to be in that category.)

smv-orrin:
This is a real bug.  Before I submitted the ticket, I confirmed the issue by running the same query in:
A) the mongo cli in a terminal window, and
B) VS Code
the mongo cli successfully returned results in a couple of seconds, VSCode failed with a TImeout

smv-orrin:
Sorry, forgot to mention, just confirmed it again by testing the same query in both the mongo cli (terminal) and VSCode, with the same results - cli succeeded, vscode failed.

StephenWeatherford:
@smv-orrin Thanks.  How long does the query take to execute in mongo console?

smv-orrin:
couple of seconds
I tried attaching a video so you could see that the cli was returning quickly, but I can't attach a video it seems.  I also don't see my comment history on this issue, I remember taking a closer look at this and noticing that it seemed to be tied to my use of the $hour function.  When I comment out the $hour function (and replace with a straight-up integer), the query runs in VSCode without issue.

StephenWeatherford:
Are you using the emulator by any chance?

StephenWeatherford:
#838 ?

smv-orrin:
[...]

PR #1105, opened 2019-05-02
Improve error-handling of Mongo shell commands

### LINKED ISSUE #838: Rewrite shell.ts to avoid 'timeout' when the actual error is something else

### LINKED ISSUE #1104: No way to pass arguments to Mongo shell
I think used to work by adding the arguments to the mongo.shell.path setting, but that no longer works.  I think adding a new setting mongo.shell.arguments is the best approach. It will also help with testing.

StephenWeatherford:
This especially should provide accurate error messages instead of "timed out" when the "use db" command fails.

Fixes #838
Also fixes #1104

Possibly fixes error reporting for:
https://github.com/Microsoft/vscode-cosmosdb/issues/1071
#988
#852
#820 
#794
#1092 

The main part is a near-rewrite of shell.ts.

Testcases in OneNote: 
image
[](url)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"displayName": "Azure Cosmos DB",
 	"description": "Create, browse, and update globally distributed, multi-model databases in Azure.",
 	"engines": {
-		"vscode": "^1.26.0"
+		"vscode": "^1.31.0"
 	},
 	"galleryBanner": {
 		"color": "#3c3c3c",
@@ -97,6 +97,10 @@
 		"languages": [
 			{
 				"id": "mongo",
+				"aliases": [
+					"Mongo Scrapbook",
+					"mongo"
+				],
 				"extensions": [
 					".mongo"
 				],
@@ -696,6 +700,10 @@
 					"when": "never"
 				},
 				{
+					"command": "cosmosDB.update",
+					"when": "never"
+				},
+				{
 					"command": "cosmosDB.executeAllMongoCommands",
 					"when": "editorLangId == 'mongo'"
 				},
@@ -732,8 +740,13 @@
 						"string",
 						"null"
 					],
-					"description": "Command to execute or full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
+					"description": "Full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
 					"default": null
+				},
+				"mongo.shell.args": {
+					"type": "string",
+					"description": "Arguments to pass when starting the Mongo shell.",
+					"default": ""
 				},
 				"mongo.shell.timeout": {
 					"type": "number",
@@ -871,7 +884,6 @@
 		"all": "npm i && npm run lint && npm test"
 	},
 	"devDependencies": {
-		"@types/clipboardy": "^1.1.0",
 		"@types/d3": "5.0.0",
 		"@types/documentdb": "^1.10.2",
 		"@types/fs-extra": "^4.0.3",
@@ -905,7 +917,6 @@
 		"antlr4ts": "^0.4.1-alpha.0",
 		"azure-arm-cosmosdb": "^1.1.2",
 		"azure-arm-resource": "^3.0.0-preview",
-		"clipboardy": "^1.2.3",
 		"d3": "^3.0.0",
 		"documentdb": "^1.14.2",
 		"event-stream": "3.3.4",
@@ -918,7 +929,7 @@
 		"socket.io": "^1.7.3",
 		"socket.io-client": "^1.7.3",
 		"underscore": "^1.8.3",
-		"vscode-azureextensionui": "^0.23.2",
+		"vscode-azureextensionui": "^0.25.0",
 		"vscode-json-languageservice": "^3.0.8",
 		"vscode-languageclient": "^4.4.0",
 		"vscode-languageserver": "^4.4.0",
--- a/src/mongo/shell.ts
+++ b/src/mongo/shell.ts
@@ -5,120 +5,201 @@
 import * as cp from 'child_process';
 import * as os from 'os';
 import * as vscode from "vscode";
-import { EventEmitter, window } from 'vscode';
+import { EventEmitter } from 'vscode';
+import { parseError } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
-import { IDisposable, toDisposable } from '../utils/vscodeUtils';
+import { splitArguments } from '../utils/splitArguments';
+
+type CommandResult = {
+	result?: string;
+	errorMessage?: string;
+};
+
+// This is used at the end of each command we send to the console. When we get this string back,
+// we know we've reached the end of that command's result.
+const endOfDataSentinelBase: string = '$EOD$';
+
+const mongoShowMoreMessage: string = 'Type "it" for more';
+const prematureExitMessage: string = `The Mongo console exited prematurely. Check the output window for possible additional data, and check your mongo.shell.path and mongo.shell.args settings.`;
 
 export class Shell {
+	private _executionId: number = 0;
+	private _stdoutData: string = "";
+	private _stderrData: string = "";
+	private _exited: boolean;
 
-	private executionId: number = 0;
-	private disposables: IDisposable[] = [];
+	private _onResult: EventEmitter<CommandResult> = new EventEmitter<CommandResult>();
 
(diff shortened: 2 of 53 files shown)


src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+						if (result.result !== undefined && !this._stderrData) {
+							if (this._mongoShell.killed || this._exited) {
+								throw new Error(prematureExitMessage);
+							}
+
+							let lines = result.result.split(os.EOL).filter(line => !!line);
+							if (lines[lines.length - 1] === mongoShowMoreMessage) {
+								// CONSIDER: Ideally we would ask or allow the user to ask for more data
+								lines = lines.splice(0, lines.length - 1);
+								lines.push("(More)");
```
EricJizbaMSFT: This seems like a pretty big behavior change. The shell used to keep typing "it" for the user, correct? Perhaps we should discuss this or file an issue to re-create that behavior before releasing as-is?

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+			// enough for other notifications to be a given a to fire first.
+			setTimeout(() => this.fireError(error), 500);
 		}
 
-		return await new Promise<string>((c, e) => {
-			let executed = false;
-			// timeout setting specified in seconds. Convert to ms for setTimeout
-			const timeout: number = 1000 * vscode.workspace.getConfiguration().get<number>(ext.settingsKeys.mongoShellTimeout);
-			const handler = setTimeout(
+		return await new Promise<string>((resolve, reject) => {
```
EricJizbaMSFT: From personal experience, it's best to put as little code inside `await new Promise` as possible. I've had cases where the code inside of that was throwing an error, which meant `resolve` or `reject` were never called and the feature basically hung with no error.

I think this would be closer to the minimum amount of code:
```typescript
const result = await new Promise<CommandResult>((resolve, reject) => {
	const timeoutHandler = setTimeout(
		() => {
			reject(`Timed out executing MongoDB command "${script}". To use a longer timeout, modify the VS Code 'mongo.shell.timeout' setting.`);
		},
		timeout);

	// Handle result or error from the console (via fireOnResultOrError)
	const disposable = this._onResult.event((r: CommandResult) => {
		clearTimeout(timeoutHandler);

		// Only the first result or error will be processed, all others ignored
		disposable.dispose();
		resolve(r);
	});
});

await delay(100); // Give STDOUT/STDERR a chance to empty

...
```

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+								lines = lines.splice(0, lines.length - 1);
+								lines.push("(More)");
+							}
+
+							let text = lines.join(os.EOL);
+							resolve(text);
+						} else {
+							// An error occurred
+							let errorDetails = result.errorMessage || "";
+							if (errorDetails.includes('ENOENT')) {
```
EricJizbaMSFT: Can we somehow leverage `parseError().errorType` here?

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
-				const result = buffers.join('') + data.substring(0, data.length - delimitter.length);
-				buffers = [];
-				this.onResult.fire({
-					exitCode: void 0,
-					result,
-					stderr: void 0
-				});
+		// Monitor STDOUT
+		this._mongoShell.stdout.on('data', (chunk: Buffer) => {
+			let data: string = chunk.toString();
```
EricJizbaMSFT: Perhaps it would be helpful to print `stdout` and `stderr` to an outputChannel (probably shouldn't be the main "Cosmos DB" one) as it happens? Rather than waiting until the end to show any progress?

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+								} else {
+									errorDetails = "Unknown error";
+								}
+							}
+
+							ext.outputChannel.appendLine(`Shell execution details: ${shellDetails}`);
+
+							// Add some STDOUT/STDERR context if available
+							const maxOutputLengthInMessage = 500;
+							if (this._stderrData) {
```
EricJizbaMSFT: Do we want this much error detail in telemetry?

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+								errorDetails += `${os.EOL}${os.EOL}${this._stderrData}`.slice(0, maxOutputLengthInMessage);
+							}
+							if (this._stdoutData) {
```
EricJizbaMSFT: I would actually print everything to an outputChannel as it happens (mentioned above) and just point them to the outputChannel in this error

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -205,13 +207,14 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
...
-				shell => {
-					return shell.useDatabase(this.databaseName).then(() => shell);
-				},
-				error => vscode.window.showErrorMessage(error));
+	private async createShell(shellPath: string, shellArgs: string | undefined): Promise<Shell> {
+		let shell = await Shell.create(shellPath, shellArgs, this.connectionString, this.root.isEmulator);
+		let useDatabaseResult = await shell.useDatabase(this.databaseName);
+		if (!useDatabaseResult.match(/^switched to db /i)) {
+			// This might be unexpected output (but maybe it's not in English), show it to the user just in case
+			ext.outputChannel.appendLine(useDatabaseResult);
```
EricJizbaMSFT: As mentioned elsewhere, I think it makes sense to have an outputChannel dedicated to _all_ output from executing mongo scripts

src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -128,20 +128,22 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
 		return new MongoCollectionTreeItem(this, newCollection);
 	}
 
-	executeCommandInShell(command: MongoCommand, context: IActionContext): Thenable<string> {
+	private executeCommandInShell(command: MongoCommand, context: IActionContext): Thenable<string> {
 		context.properties["executeInShell"] = "true";
-		return this.getShell().then(shell => shell.exec(command.text));
+		return this.getShell().then(shell => shell.executeScript(command.text));
```
EricJizbaMSFT: One of the biggest confusions I had about this code was that `createShell` implied it was keeping a shell around. In reality, it's only ever executing one script and then throwing it away. Can we remove `createShell` and move that code into this function?

package.json
```diff
@@ -732,9 +732,14 @@
 						"string",
 						"null"
 					],
-					"description": "Command to execute or full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
+					"description": "Full path to folder and executable to start the Mongo shell, needed by some Mongo scrapbook commands. If empty, will search in the system path for 'mongo'.",
 					"default": null
 				},
+				"mongo.shell.args": {
+					"type": "string",
```
EricJizbaMSFT: Can we just make this a string array? So that we don't have to split the args ourselves?

src/mongo/shell.ts
```diff
@@ -5,120 +5,201 @@
...
+			}
+		});
+
+		// Monitor STDERR
+		this._mongoShell.stderr.on('data', (buffer: Buffer) => {
+			this._stderrData += buffer.toString();
+		});
+		this._mongoShell.stderr.on('end', () => {
+			if (this._stderrData) {
+				this.fireError(this._stderrData);
```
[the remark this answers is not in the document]
StephenWeatherford: I got this in the case of user login failing, which CTI was getting and possibly other customers.

StephenWeatherford:
@EricJizbaMSFT @PrashanthCorp @nturinski FYI, I was trying to get this finished up yesterday and also greatly simply the algorithm by passing the script to be processed through the command line instead of stdin (which would allow me to use Eric's cpUtils unchanged.  Unfortunately, it causes some differences in output that I don't think I can overcome.  The problem with using cpUtils is that as currently written the mongo shell execution requires an interactive handshaking with stdin/stdout.  Probably giving it a rest until next week, but want to get it out of the way.  Also wondered if it would make sense to use the vscode terminal for this (but we'd have to be able to read the output).

EricJizbaMSFT:
That would've been great if we could just re-use `cpUtils`, but honestly not that surprised that it didn't work. Oh well ¯\\\_(ツ)\_\/¯

StephenWeatherford:
> That would've been great if we could just re-use `cpUtils`, but honestly not that surprised that it didn't work. Oh well ¯\_(ツ)_/¯

It was *so* close.

StephenWeatherford:
Superseded by #1148

Issue #837, opened 2018-09-26
Pop up an error after running command 'Cosmos DB: Update to Cloud'


v-dantch:
Repro Steps:
1. F1 -> Type and select 'Cosmos DB: Update to Cloud'.
2. Check the result.

Expect:
Prompt to select one changed document, then save and upload it to Cloud.
![expect](https://user-images.githubusercontent.com/34729092/46068883-bb208e80-c169-11e8-867a-362b36963459.PNG)

Actual:
Pop up an error.
![error](https://user-images.githubusercontent.com/34729092/46067680-f5d4f780-c166-11e8-9ac9-961501e203ad.png)

StephenWeatherford:
This command should be removed from the command palette.

v-dantch:
Verified on the latest build [20190526.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=6335), this command was removed.

PR #1122, opened 2019-05-24
remove Update to Cloud from command palette

Sneezry:
fix https://github.com/microsoft/vscode-cosmosdb/issues/837

### DIFF
--- a/package.json
+++ b/package.json
@@ -310,6 +310,15 @@
 				"title": "Copy Connection String"
 			},
 			{
+				"category": "Cosmos DB",
+				"command": "cosmosDB.update",
+				"title": "Update to Cloud",
+				"icon": {
+					"light": "/resources/icons/theme-agnostic/UpdateOutline.svg",
+					"dark": "/resources/icons/theme-agnostic/UpdateOutline.svg"
+				}
+			},
+			{
 				"category": "SQL (DocumentDB)",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -309,15 +309,6 @@
 				"command": "cosmosDB.copyConnectionString",
 				"title": "Copy Connection String"
 			},
-			{
-				"category": "Cosmos DB",
```
EricJizbaMSFT: I don't think this will work - we need to define commands here if we're going to reference them from a menu time. Instead, you want to add this command to the "commandPalette" section with `"when": "never"`

Take a look at this: https://github.com/microsoft/vscode-cosmosdb/blob/master/package.json#L689

Sneezry: VS Code should look for command in `menu` with command id which is registered by `vscode.commands.registerCommand`, and there's no need to define that in `commandPalette`. However, I haven't tested that on my side, have you tried to just remove the command from `commandPalette`? Did you meet any problem with that?

EricJizbaMSFT: Did you trying F5-ing the extension with your change? This is what I see:
 Screen Shot 2019-05-24 at 10 14 44 AM

Sneezry: I just tried to debug, and got the same error. I don't think that make sense :-S

I have pushed another commit to set never show `cosmosDB.update` in command palette.

Issue #820, opened 2018-09-14
Commands don't work against cosmosDb account


hanigan6:
Repro steps:
Attach to Db Account
Connect to db
Create new shell command
`db.collection.createIndex({"name": 1, "shardKey": 1}, {unique: true})`

Action: cosmosDB.executeMongoCommand
Error type: string
Error Message: Timed out executing MongoDB command "use klic"

Connected to a VPN with access through the portal
Version: 0.8.0
OS: win32

Other commands like
`db.collection.find({})`
work just fine

only happens on a cosmosDb MongoApi account

hanigan6:
Basic commands work such as 
`db.collection.find({});` this returns all documents like you would expect

StephenWeatherford:
Would you mind quickly trying an experimental version that may report better errors? It's easy to uninstall afterwards via the extensions view in vs code.  See instructions here: https://github.com/Microsoft/vscode-cosmosdb/issues/852#issuecomment-479734886.

Thanks!
Stephen

StephenWeatherford:
With the next release, I've fixed some scenarios where we show timeout instead of the correct error message.  Hopefully that will allow you to better understand the actual issue.  Thanks!

Fixed by #1162

Issue #794, opened 2018-08-16
trying to run a simple aggregate query Times out every time


smv-orrin:
Repro steps:
<Enter steps to reproduce issue>

Action: cosmosDB.executeAllMongoCommands
Error type: string
Error Message: Timed out executing MongoDB command "db.ingestion_queue.aggregate([{$match:{date_time:{$gte:ISODate("2018-08-09")}}},{$group:{_id:{provider:"$provider_name",file_type:"$file_type"},total:{$sum:"$file_size_bytes"}}}],{maxTimeMS:0});"

Version: 0.8.0
OS: darwin

smv-orrin:
No problem, happy to help.  As far as reproducing it, there isn't much to report, just connecting to a mongo instance in VS Code and trying to run the query makes it happen.  I can run the same query without issue when executing via the mongo shell.  Let me know if you need anything else to dig further.

smv-orrin:
![screen shot 2018-08-16 at 1 54 53 pm](https://user-images.githubusercontent.com/8709753/44228920-1ff4cc00-a15c-11e8-8837-60d6323ab3d8.png)
![screen shot 2018-08-16 at 1 55 37 pm](https://user-images.githubusercontent.com/8709753/44228933-26834380-a15c-11e8-9f41-260b361dd943.png)

smv-orrin:
Other info worth noting.  I'm connecting to the primary node in a replicaSet thru a local port forward, so the local port (9204) is forwarded to the 27017 port of the primary node.  I'm able to execute the same query via Robo 3T.

PrashanthCorp:
Ah, I missed the ISODate. :) 
#769 is a log of that. 

We are looking into fixing that. 

As a workaround, we support using [`$date` from EJSON](https://docs.mongodb.com/manual/reference/mongodb-extended-json/#date) . 
Hope that helps.

StephenWeatherford:
I'm guessing the "timed out" is an incorrect error message, we should fix that.  Unfortunately requires a rewrite of one of our classes.

StephenWeatherford:
Would you mind quickly trying an experimental version that may report better errors? It's easy to uninstall afterwards via the extensions view in vs code.  See instructions here: https://github.com/Microsoft/vscode-cosmosdb/issues/852#issuecomment-479734886.

Thanks!
Stephen

StephenWeatherford:
Closing, as the error no longer occurs.

PR #1148, opened 2019-07-20
[WIP] Improve error-handling of mongo shell

### LINKED ISSUE #838: Rewrite shell.ts to avoid 'timeout' when the actual error is something else

### LINKED ISSUE #1104: No way to pass arguments to Mongo shell

StephenWeatherford:
1) Rewrite interaction with Mongo shell - this especially should provide accurate error messages instead of "timed out" when the "use db" command fails.
2) Added mongo.shell.args configuration
3) Some minor cleanup/improvements

Close to finished, sending up the PR early get feedback on the approach.  Ideally I'd like to write some tests, too.  Rewrote shell.ts completely from scratch after Eric's comments from previous PR.

Fixes #838
Also fixes #1104

Possibly fixes error reporting for:
https://github.com/Microsoft/vscode-cosmosdb/issues/1071
#988
#852
#820 
#794
#1092 

Testcases in OneNote: 
image
[](url)


StephenWeatherford:
> Generally a bit wary of all the `debounce`, `setTimeout`, etc. code in InteractiveChildProcess (what does `await delay(0)` even do??) but looks good overall so far. Will wait to do a more thorough review until you say it's ready.

Yeah, not happy about it, either, am open to suggestions.

setTimeout(fn, 0) is essentially "execute on next event loop" in Node.js (http://voidcanvas.com/setimmediate-vs-nexttick-vs-settimeout/ - overly complicated). Shouldn't matter what delay I use, I just have to give the stream time to process.

PR #1178, opened 2019-08-31
Fix some remaining mongo shell issues

### LINKED ISSUE #1115: Increase mongo.shell.timeout default value
At the moment default value is 5 which is very low in case of more or less complex requests to a Mongo Cloud Atlas cluster (like an aggregation query). I would suggest 30 as a default value.

### LINKED ISSUE #1149: An unmatched error occurs after executing 'db.test.foo().pretty();' in scrapbook
**OS:** Windows
**Build Version:** [20190722.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=7901)

**Repro Steps:**
1. Connect a scrapbook to the mongo shell.
2. Set the mongo.shell.args setting to: **"mongo.shell.args": "-u abc -p abc"**
3. Execute this command in the scrapbook: **db.test.foo().pretty();**

**Expect:**
Pops up this error dialog.
image

**Actual:**
Show the below result.
image

**More info:**
1. Expected output.
image
2. Recognize setting **"mongo.shell.args": "-u abc -p abc"** as unknown configuration setting.
image

3. This issue also reproduces when setting mongo.shell.args setting to **"mongo.shell.args": "--version"**.

### LINKED ISSUE #1176: Keep executing command more than 3 mins
**OS**: Win10 & Mac
**Build Version**: [20190829.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=9268)
**Regression:** Previous Release (0.10.2)

**Repro Steps:**
1. Create a database for a Mongo account.
2. New a scrapbook -> Type below comments:
image
3. Execute all --> Check the result.

**Expect:**
Succeed to execute all comments.

**Actual:**
Keep executing command more than 3 mins.
image

**More Info:**
This issue does not reproduce on the previous release 0.10.2.

StephenWeatherford:
Fixes #1176 (not correctly capturing results from some queries)
Fixes #1149 (probably already fixed, but also testcase had bad mongo.shell.args setting)
Fixes #1115 (Increase default timeout)
Add mongo.shell.args to README
Don't choke on bad mongo.shell.{path,args} settings (these utilities can be used to fix all our settings, but I'm not doing it here)
Immediately after browsing to mongo.exe, the query would fail because we're using correctly path from determineShellPathOrCmd)

### DIFF
--- a/src/mongo/tree/MongoDatabaseTreeItem.ts
+++ b/src/mongo/tree/MongoDatabaseTreeItem.ts
@@ -12,6 +12,7 @@ import { appendExtensionUserAgent, AzureParentTreeItem, DialogResponses, IAction
 import { getThemeAgnosticIconPath } from '../../constants';
 import { ext } from '../../extensionVariables';
 import * as cpUtils from '../../utils/cp';
+import { getWorkspaceArrayConfiguration, getWorkspaceConfiguration } from '../../utils/getWorkspaceConfiguration';
 import { connectToMongoClient } from '../connectToMongoClient';
 import { MongoCommand } from '../MongoCommand';
 import { addDatabaseToAccountConnectionString } from '../mongoConnectionStrings';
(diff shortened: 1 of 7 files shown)


src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -191,16 +192,15 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
...
+							await vscode.workspace.getConfiguration().update(ext.settingsKeys.mongoShellPath, fsPath, vscode.ConfigurationTarget.Global);
+							return fsPath;
 						} else {
 							throw new UserCancelledError();
 						}
 					}
 				} else if (response === browse) {
-					this._cachedShellPathOrCmd = undefined;
 					vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://docs.mongodb.com/manual/installation/'));
+					return undefined;
```
PrashanthCorp: Shouldn't this imply the return type to be Promise<string|undefined>?

PR #1190, opened 2019-09-06
Attach accounts: Disallow spaces in connection strings 

### LINKED ISSUE #994: Allow to attach Table/Graph/SQL accounts with trailing whitespace
**VS Code:** 1.31.0-insider (user setup)
**Node.js**: 10.2.0
**OS:** Windows_NT x64 10.0.17134
**CosmosDB Extension Version:** [20190108.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=1895)

**Reproduce Steps:**

1. Right click one SQL/Table/Graph account ->'Copy Connection String'.
2. Click 'Attach Database Account' -> Paste the copied connection string with trailing white-space like below:
Connection String = "AccountEndpoint=https://xuangraphacct4.documents.azure.com:443/;AccountKey=Gv8GQ9NZQyu0gWii0gPubGWpNTWEDkWtCdOjvk1iJjO17FctBi6TelPlPZ2m8thk8cjI3CgKmJAPSDnypjIoAw==    "
3. Press 'Enter'

**Expect Experiment:**
There is two cases for the expected result:
1. Show a red border after pasting the connection string with trailing white-space like below.
image
2. No red border shows but the account can be attached successfully.

**Actual Experiment:**
No red border shows and the account can be attached successfully.

**More info:**
Mongo accounts are not allowed to attach when adding trailing white-space for the connection string(There is no red border and nothing happens after attaching it with this format)

PrashanthCorp:
Fixes #994

### DIFF
--- a/src/docdb/docDBConnectionStrings.ts
+++ b/src/docdb/docDBConnectionStrings.ts
@@ -10,7 +10,7 @@ export function parseDocDBConnectionString(connectionString: string): ParsedDocD
     const endpoint = getPropertyFromConnectionString(connectionString, 'AccountEndpoint');
     const masterKey = getPropertyFromConnectionString(connectionString, 'AccountKey');
     const databaseName = getPropertyFromConnectionString(connectionString, 'Database');
-    if (!endpoint || !masterKey || connectionString.match(" ")) {
+    if (!endpoint || !masterKey || connectionString.includes(" ")) {
         throw new Error('Invalid Document DB connection string.');
     }
     return new ParsedDocDBConnectionString(connectionString, endpoint, masterKey, databaseName);
--- a/src/docdb/docDBConnectionStrings.ts
+++ b/src/docdb/docDBConnectionStrings.ts
@@ -10,7 +10,7 @@ export function parseDocDBConnectionString(connectionString: string): ParsedDocD
     const endpoint = getPropertyFromConnectionString(connectionString, 'AccountEndpoint');
     const masterKey = getPropertyFromConnectionString(connectionString, 'AccountKey');
     const databaseName = getPropertyFromConnectionString(connectionString, 'Database');
-    if (!endpoint || !masterKey) {
+    if (!endpoint || !masterKey || connectionString.match(" ")) {
         throw new Error('Invalid Document DB connection string.');
     }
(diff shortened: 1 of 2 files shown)


src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -325,7 +326,7 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
     }
 
     static validateMongoConnectionString(value: string): string | undefined {
-        if (value && value.match(/^mongodb(\+srv)?:\/\//)) {
+        if (value && value.match(/^mongodb(\+srv)?:\/\//) && !value.match(" ")) {
```
StephenWeatherford: nit: Generally you should use contains() or indexOf>=0 for a simple comparison like this since it's faster.  Won't make a lot of difference in this case, though.

src/docdb/docDBConnectionStrings.ts
```diff
@@ -10,7 +10,7 @@ export function parseDocDBConnectionString(connectionString: string): ParsedDocD
     const endpoint = getPropertyFromConnectionString(connectionString, 'AccountEndpoint');
     const masterKey = getPropertyFromConnectionString(connectionString, 'AccountKey');
     const databaseName = getPropertyFromConnectionString(connectionString, 'Database');
-    if (!endpoint || !masterKey) {
+    if (!endpoint || !masterKey || connectionString.match(" ")) {
```
StephenWeatherford: Can you point to documentation that indicates spaces inside a connection string are not allowed?  For instance, what if a password has spaces?

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -148,6 +148,7 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
             });
 
             if (connectionString) {
+                connectionString = connectionString.trim();
```
StephenWeatherford: nit: if validation doesn't allow spaces, won't trim() do nothing?

SkeLLLa:
MongoDB passwords could contain spaces. So it's better not to disallow them, but escape (or encode).

EricJizbaMSFT:
Closing due to lack of activity

PR #1202, opened 2019-09-24
Update OutputChannel and LocationStep

nturinski:
There are a lot of spots where there are forEach output loops.  I didn't use appendLog there but tried to make sure I did right before things were listed.  Here's an example of how it looks:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -783,6 +783,11 @@
 					"default": true,
 					"description": "Show warning dialog when uploading a document to the cloud."
 				},
+				"cosmosDB.enableOutputTimestamps": {
+					"type": "boolean",
+					"default": true,
+					"description": "Prepends each line displayed in the output channel with a timestamp."
+				},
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -934,7 +939,7 @@
 		"socket.io": "^1.7.3",
 		"socket.io-client": "^1.7.3",
 		"underscore": "^1.8.3",
-		"vscode-azureextensionui": "0.26.5",
```
EricJizbaMSFT: When you do these PRs in the future - can you add comments in the description saying what breaking changes there were and if they affect this extension - especially in cases where you're bumping past multiple breaking changes? For example:
> 0.27 didn't affect this extension and 0.28 was about the output channel

or

> 0.27 was about LocationListStep and 0.28 was about the output channel

nturinski: You got it boss.

PR #1203, opened 2019-09-25
Create Cosmos DB account

v-wuzhai:
Create Cosmos DB account


test/global.test.ts
```diff
@@ -5,12 +5,16 @@
...
 
 // tslint:disable-next-line:strict-boolean-expressions export-name
 export let longRunningTestsEnabled: boolean = !/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
+export const testUserInput: TestUserInput = new TestUserInput(vscode);
 
 // Runs before all tests
 suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
     this.timeout(120 * 1000);
     await vscode.commands.executeCommand('cosmosDB.refresh'); // activate the extension before tests begin
+    ext.ui = testUserInput;
```
nturinski: Add `ext.outputChannel = new TestOutputChannel()` here.  We want to see the output being displayed when the tests run.

test/nightly/global.resource.test.ts
```diff
@@ -0,0 +1,53 @@
...
+export let testAccount: TestAzureAccount;
+export let client: CosmosDBManagementClient;
+export const resourceGroupsToDelete: string[] = [];
+
+suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
+    if (!longRunningTestsEnabled) {
+        this.skip();
+    }
+
+    this.timeout(120 * 1000);
```
nturinski: Could you change both of the timeouts to a format like

`this.timeout(10 * 60 * 1000);`

This is very unreadable.  With the above format, it's easy to tell that the timeout is 10 minutes.

Issue #1201, opened 2019-09-19
Connection strings to with valid characters returns an invalid character error


PrashanthCorp:
The only character in the password that wasn't an alphanumeric character was `=`.

Issue #1197, opened 2019-09-16
Update to the latest Mongo DB driver


fiveisprime:
Customer request to handle some advanced features.

PR #1205, opened 2019-09-28
Update Mongodb to latest driver

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1197 
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1201

Some complications arose from updating Mongodb's type definitions which led me to have to update node's type defs which is why I had to make some changes to the Graph code.

I did some very basic tests for Graph (made sure I could still create an account, db, graph, and that the HTML view still works)

But connecting to dbs work and also works through the App Service Extension via the "Connect to Cosmos DB" node.

Edit: Builds are still failing on CI for some reason, but I'll figure that out on Monday 😓


src/mongo/mongoConnectionStrings.ts
```diff
@@ -48,18 +49,18 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
         // the replicaset host name, which is different than what is in the connection string
         // "s" is not part of ReplSet static definition but can't find any official documentation on it. Yet it is definitely there at runtime. Grandfathering in.
         // tslint:disable-next-line:no-any
         let rs: any = serverConfig;
-        host = rs.s.replset.s.seedlist[0].host;
-        port = rs.s.replset.s.seedlist[0].port;
+        host = rs.s.options.servers[0].host;
+        port = rs.s.options.servers[0].port;
     } else {
         host = serverConfig['host'];
```
EricJizbaMSFT: So the top case changed - I wonder if the else case also changed? That would be bad if we only work in the Azure case and break other mongo dbs

nturinski: Do you just test this with a local mongodb instance running?

EricJizbaMSFT: I haven't tested Cosmos in a while haha... but that sounds like a good option

nturinski: image

image

We good.

PR #1212, opened 2019-10-09
Create MongoDB Database test case

v-wuzhai:


nturinski:
Could you investigate why the nightly tests are failing after we merged the `Create Cosmos DB Account` tests?  I think we have to add installing mongo.exe to our gulp script.

image

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -7,34 +7,74 @@ import * as assert from 'assert';
...
+        const collectionName: string = randomUtils.getRandomHexString(12);
+        const testInputs: string[] = [testAccount.getSubscriptionContext().subscriptionDisplayName, `${accountName} (MongoDB)`, databaseName, collectionName];
+        await testUserInput.runWithInputs(testInputs, async () => {
+            await vscode.commands.executeCommand('cosmosDB.createMongoDatabase');
+        });
+        const response: string = await getMongoDBDatabase(await getUrl(databaseName));
+        assert.equal(JSON.parse(response).properties.id, databaseName);
+    });
+
+    async function getUrl(Path: string): Promise<string> {
```
nturinski: This seems kind of complicated and prone to errors.  I think an easier way to verify that the mongo DB got created would be to list all the databases and find the one that was just created.

image

The code can be found here: https://github.com/microsoft/vscode-cosmosdb/blob/master/src/mongo/tree/MongoAccountTreeItem.ts#L71-L72

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -38,43 +37,22 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
...
-            }
+        await vscode.env.clipboard.writeText('');
+        await testUserInput.runWithInputs([`${accountName} (MongoDB)`], async () => {
+            await vscode.commands.executeCommand('cosmosDB.copyConnectionString');
         });
+        const connectionString: string = await vscode.env.clipboard.readText();
+        const mongoClient: MongoClient | undefined = await connectToMongoClient(connectionString, appendExtensionUserAgent())
+        let listDatabases: { databases: IDatabaseInfo[] } = await mongoClient.db(accountName).admin().listDatabases();
+        const MongoDatabase: IDatabaseInfo[] = listDatabases.databases.filter((database: IDatabaseInfo) => (database.name == databaseName));
+        assert.ok(MongoDatabase[0].name == databaseName, `Mongo Database should be ${databaseName} rather than ${MongoDatabase[0].name}.`);
```
nturinski: If the mongoDatabase is not found (the filter fails to return anything) then this will fail as you cannot read `undefined.name`.  I think you could do `listDatabases.databases.find()` or .`some()` and see if it returns a value.

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -38,43 +37,22 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
...
-                resolve();
-            }
+        await vscode.env.clipboard.writeText('');
+        await testUserInput.runWithInputs([`${accountName} (MongoDB)`], async () => {
+            await vscode.commands.executeCommand('cosmosDB.copyConnectionString');
         });
+        const connectionString: string = await vscode.env.clipboard.readText();
+        const mongoClient: MongoClient | undefined = await connectToMongoClient(connectionString, appendExtensionUserAgent())
+        let listDatabases: { databases: IDatabaseInfo[] } = await mongoClient.db(accountName).admin().listDatabases();
+        const MongoDatabase: IDatabaseInfo[] = listDatabases.databases.filter((database: IDatabaseInfo) => (database.name == databaseName));
```
nturinski: Please change to mongoDatabase.

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -38,43 +37,22 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
...
-            } else {
-                resolve();
-            }
+        await vscode.env.clipboard.writeText('');
+        await testUserInput.runWithInputs([`${accountName} (MongoDB)`], async () => {
+            await vscode.commands.executeCommand('cosmosDB.copyConnectionString');
         });
+        const connectionString: string = await vscode.env.clipboard.readText();
+        const mongoClient: MongoClient | undefined = await connectToMongoClient(connectionString, appendExtensionUserAgent())
+        let listDatabases: { databases: IDatabaseInfo[] } = await mongoClient.db(accountName).admin().listDatabases();
```
nturinski: This could be a `const`.

PR #1213, opened 2019-10-11
Update for patch release

nturinski:
So I didn't do a release branch for this one because the only other change besides the MongoDB driver updates were for the timestamps in the output channel.  I figure when we do testing, we'll just keep an eye out for funky timestamping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
+## 0.11.1 - 2019-10-16
+### Added
+- Update mongodb package to 3.3.2
+- Timestamps are prepended to all output messages.  Turn this off by setting `cosmosDB.enableOutputTimestamps` to `false`
+
 ## 0.11.0 - 2019-09-03
 ### Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
+## 0.11.1 - 2019-10-16
+### Added
+- Update mongodb package to 3.3.2
```
EricJizbaMSFT: As discussed offline - would like to know what functional changes this has for the user. Right now it just sounds like engineering work, which we don't typically put in the changelog

nturinski: @fiveisprime Could we discuss what exactly updating the mongodb package fixed for the customer either offline or in a comment in this thread?

nturinski:
- Retryable Writes
- DNS Seedlist Support
- Change Streams
- Sessions

According to https://github.com/mongodb/node-mongodb-native/blob/master/CHANGES_3.0.0.md#sessions

nturinski:
Closing in favor of 0.12.0 release

Issue #1218, opened 2019-10-14
Pop up an error after selecting "Open installation page"


v-dantch:
**OS**: Win10
**Build Version**: [20191009.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=11035)
**Regression**: Not a Regression

**Repro Steps:**
1. Remove the "mongo.shell.path" vscode setting and make sure you don't have mongo.exe set as a service in your path.
2. Try "db.version()" in a scrapbook.
3. Pop up below error.
![image](https://user-images.githubusercontent.com/34729092/66742822-22130180-eeab-11e9-904f-8a8fa65e5ada.png)

4. Select "Open installation page".

**Expect:**
Open the browser and navigate to a matched website https://docs.mongodb.com/manual/installation/ without errors.

**Actual:**
Pop up an error.
![image](https://user-images.githubusercontent.com/34729092/66742858-3a831c00-eeab-11e9-9373-aeb0eaf8fce9.png)

PR #1219, opened 2019-10-14
Throw userError rather than return undefined

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1218

I thought that this might be a regression from updating the Mongodb package, but that doesn't seem to be the case.


nturinski:
Filed an issue to turn on strictNull checks https://github.com/microsoft/vscode-cosmosdb/issues/1220

PR #1228, opened 2019-11-07
remove invalid characters from local filename

### LINKED ISSUE #1227: Azure Cosmos DB extension error no such file or directory when browsing Azure document
Repro steps:

1. Connected to my MongoDB
2. Browse collection but can't open a particular document

Note: although obfuscated in the error message below, the filename and the `_id` of the document contains `:` characters. Fine in MongoDB but not for Windows filenames.

Action: cosmosDB.openDocument
Error type: ENOENT
Error Message: ENOENT: no such file or directory, open 'C:\Users\<user>\AppData\Local\Temp\vscode-cosmosdb-editor\xxxx-cosmos-document.json'

Version: 0.11.0
OS: win32
Product: Visual Studio Code
Product Version: 1.39.2
Language: en

pschaeflein:
Fixes #1227

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,948 +1,947 @@
 {
-	"name": "vscode-cosmosdb",
-	"version": "0.11.0",
-	"aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
-	"publisher": "ms-azuretools",
-	"displayName": "Azure Cosmos DB",
-	"description": "Create, browse, and update globally distributed, multi-model databases in Azure.",
-	"engines": {
-		"vscode": "^1.31.0"
-	},
-	"galleryBanner": {
-		"color": "#3c3c3c",
-		"theme": "dark"
-	},
-	"icon": "resources/cosmos.png",
-	"categories": [
-		"Azure"
-	],
-	"keywords": [
-		"Cosmos DB",
-		"DocumentDB",
-		"Graph",
-		"Gremlin",
-		"MongoDB"
-	],
-	"preview": true,
-	"homepage": "https://github.com/microsoft/vscode-cosmosdb/blob/master/README.md",
-	"bugs": {
-		"url": "https://github.com/microsoft/vscode-cosmosdb/issues"
-	},
-	"license": "SEE LICENSE IN LICENSE.md",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/microsoft/vscode-cosmosdb"
-	},
-	"main": "./main",
-	"activationEvents": [
-		"onLanguage:mongo",
-		"onView:cosmosDBExplorer",
-		"onCommand:cosmosDB.attachDatabaseAccount",
-		"onCommand:cosmosDB.attachEmulator",
-		"onCommand:cosmosDB.createAccount",
-		"onCommand:cosmosDB.deleteAccount",
-		"onCommand:cosmosDB.detachDatabaseAccount",
-		"onCommand:cosmosDB.createMongoDatabase",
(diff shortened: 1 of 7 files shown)


src/CosmosEditorManager.ts
```diff
@@ -21,153 +22,153 @@ import { MongoDocumentTreeItem } from './mongo/tree/MongoDocumentTreeItem';
 import * as vscodeUtils from './utils/vscodeUtils';
 
 export interface ICosmosEditor<T = {}> {
-    label: string;
```
EricJizbaMSFT: Can you revert all of these whitespace changes? Perhaps you have a VS Code setting that is auto-formatting?

Also please fix the merge conflicts in this file (might be related to your whitespace changes)

src/CosmosEditorManager.ts
```diff
@@ -21,153 +22,153 @@ import { MongoDocumentTreeItem } from './mongo/tree/MongoDocumentTreeItem';
...
-            preserveFocus = true;
-            const activeEditor = vscode.window.activeTextEditor;
-            if (activeEditor && activeEditor.viewColumn >= vscode.ViewColumn.One) {
-                column = activeEditor.viewColumn < ViewColumn.Three ? activeEditor.viewColumn + 1 : ViewColumn.One;
-            }
-        }
-
-        const localDocPath = path.join(os.tmpdir(), 'vscode-cosmosdb-editor', fileName);
-        await fse.ensureFile(localDocPath);
+    const localDocPath = path.join(os.tmpdir(), 'vscode-cosmosdb-editor', filenamify(fileName, { maxLength: 255, replacement: "--" }));
```
EricJizbaMSFT: We try not to add third party dependencies if possible and I'm guessing we could fix the bug ourselves in a few lines of code without using `filenamify`

pschaeflein: I use Windows and am completely in the dark about what characters are invalid for a filename on all the platforms supported by the extension. Is there a function already in the extension for handling that? Is there an authoritative list available?

src/CosmosEditorManager.ts
pschaeflein: Is there an [editorconfig](https://EditorConfig.org) file that describes the whitespace preferences for the project? Perhaps add one to the repo?

src/CosmosEditorManager.ts
EricJizbaMSFT: Seems like mac/linux are much less restrictive than windows, so I wouldn't really worry about them. https://en.wikipedia.org/wiki/Filename

The file name here is purely for readability and doesn't affect the functionality, so we could also just remove any non-alphanumeric character to cover all bases

package.json
```diff
@@ -1,947 +1,947 @@
 {
```
EricJizbaMSFT: Please revert changes to this file - looks like only whitespace changes. FWIW - I've filed https://github.com/microsoft/vscode-cosmosdb/issues/1235 to clean up whitespace in this repo and add VS Code settings to automatically keep it consistent

src/CosmosEditorManager.ts
```diff
@@ -59,7 +59,8 @@ export class CosmosEditorManager {
             }
         }
 
-        const localDocPath = path.join(os.tmpdir(), 'vscode-cosmosdb-editor', fileName);
+        const localFilename = fileName.replace(/[<>:"\/\\|?*]/g, "--");
```
EricJizbaMSFT: nit: I'd prefer to replace with a single hyphen just to keep the length of the file name the same

Issue #1193, opened 2019-09-09
How to access collection with "-" in name


AlexBlokh:
How do I access collection with "-" in name?

<img width="458" alt="Screen Shot 2019-09-09 at 16 56 57" src="https://user-images.githubusercontent.com/4045375/64537076-d6f15600-d322-11e9-9057-eb9003fa8a92.png">

Repro steps:

1. db.{collection-name}.find() or db.getCollection('collection-name').find()

Action: cosmosDB.executeMongoCommand
Error type: Error
Error Message: collection name must be a String

Version: 0.11.0
OS: darwin
Product: Visual Studio Code
Product Version: 1.38.0
Language: en

<summary>Call Stack</summary>
```
module.exports.t.checkCollectionName extension.bundle.js:5:53946
new C extension.bundle.js:109:8503
N.module.exports.N.collection extension.bundle.js:129:33155
MongoFindResultEditor.<anonymous> extension.bundle.js:622:1082492
s extension.bundle.js:622:1081907
process._tickCallback next_tick.js:68:7
```

PrashanthCorp:
Use `db.getcollection(<hyphened-name>)`

PrashanthCorp:
#981  is related.

vilic:
@PrashanthCorp this makes the extension unusable for us, really hope it gets fixed soon.

nturinski:
Hi @vilic and @AlexBlokh 

While this issue isn't resolved, with the release of v0.12.0, there is now a workaround for this.  If you need to access a collection with a hyphen ("-") in the name, please use `getCollection(col-name)`.

Hope that this helps while we solve the actual problem. Have a great weekend!

PR #1232, opened 2019-11-11
Don't prevent Lexer errors from executing mongo scrapbook queries

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1193, https://github.com/microsoft/vscode-cosmosdb/issues/981

Today the lexer is throwing an error for valid mongo commands.  Updating the mongo.g4 file (where the lexer is pulling the grammar) is not a trivial task that I think will take me a lot of investigating to do.

However, if we don't block users from executing commands that the lexer considers invalid, users can still workaround it.

I don't think this is quite a proper fix, perse, but at least it will unblock the users who have run into the above errors.

It looks like this if a mongo command _actually_ fails:
image

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -109,7 +109,8 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 		if (command.errors && command.errors.length > 0) {
 			//Currently, we take the first error pushed. Tests correlate that the parser visits errors in left-to-right, top-to-bottom.
 			const err = command.errors[0];
-			throw new Error(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
+			// don't await intentionally
+			ext.ui.showWarningMessage(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
 		}
 
 		if (command.name === 'find') {


src/mongo/MongoScrapbook.ts
```diff
@@ -109,7 +109,8 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 		if (command.errors && command.errors.length > 0) {
 			//Currently, we take the first error pushed. Tests correlate that the parser visits errors in left-to-right, top-to-bottom.
 			const err = command.errors[0];
-			throw new Error(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
+			// don't await intentionally
+			ext.ui.showWarningMessage(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
```
EricJizbaMSFT: You said it's not trivial to update the mongo.g4 file, but this change doesn't feel trivial either. It feels like a pretty significant behavior change that affects many code paths and we would have to spend a decent amount of time ensuring it doesn't have negative effects. IMO it seems more valuable to spend the time on the mongo.g4 file and properly fix the main issue

A few reasons I don't like this:
1. We no longer get these errors in telemetry
1. If the error here is a false positive, it feels confusing to show an error when the command will actually succeed
1. If the error is accurate, users will get duplicate errors (one as a message, one in the file)
1. If we have syntax errors - that means we can't parse something. Sounds risky to execute a command we can't even parse (what if it sends half of one command and half of another)

StephenWeatherford: Agreed, you should fix the mongo.g4 file, this change would have major effects down the road.  I've got some experience with it, let me know if you'd like help.

nturinski: Please and thank you!

nturinski:
Closing due to undesirable effects.

PR #1233, opened 2019-11-12
Delete MongoDB account

v-wuzhai:


test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -63,6 +63,16 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
...
     });
 
+    test('Delete account', async () => {
+        const createAccount: CosmosDBManagementModels.DatabaseAccount = await client.databaseAccounts.get(resourceGroupName, accountName);
+        assert.ok(createAccount);
+        const testInputs: string[] = [`${accountName} (MongoDB)`, DialogResponses.deleteResponse.title];
+        await testUserInput.runWithInputs(testInputs, async () => {
+            await vscode.commands.executeCommand('cosmosDB.deleteAccount');
+        });
+        await assertThrowsAsync(async () => await client.databaseAccounts.get(resourceGroupName, accountName), /Error/);
```
nturinski: While this does throw an error when the API call fails, there are many reasons that this could call could throw an error (like if there is an outage or something).  

'd rather you get the list of accounts by calling `client.databaseAccounts.list()`  and then try to find the deleted account.  If that returns false, then we can assume that the account was deleted.

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -73,3 +83,13 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
     }
 });
 
+async function assertThrowsAsync(fn: { (): Promise<CosmosDBManagementModels.DatabaseAccount>; (): void; }, regExp: RegExp): Promise<void> {
```
nturinski: And then you won't need this anymore.

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -63,6 +63,18 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
         assert.ok(collection);
     });
 
+    test('Delete account', async () => {
+        const createAccount: CosmosDBManagementModels.DatabaseAccount = await client.databaseAccounts.get(resourceGroupName, accountName);
```
nturinski: Could you rename this to just `mongoAccount`? Since this isn't actually creating an account.

Issue #1222, opened 2019-10-17
MongoDB: Fully qualified namespace is too long error when creating collection with validator


gtbono:
When I try to create a MongoDB collection using the following command on VS Code, I get an error:

```
db.createCollection("teste", {
    validator: {
        jsonSchema: {
            bsonType: "object",
            required: ["teste1", "teste2"],
            properties: {
                teste1: {
                    bsonType: "string",
                    description: "Deve ser uma string"
                },
                teste2: {
                    bsonType: "string",
                    description: "Deve ser uma string"
                },
            }
        }
    }
})
```
But when I ran this exact command on the Mongo shell the command executes without problem, I believe this is somehow being mistaken with the 120 byte limit for the collection name.

nturinski:
Hi @gtbono

Are you creating a local database in the Mongo shell or are you using Azure Cosmos?  I was just curious.

Thanks for reporting! I have a fix for this here: #1236

PR #1236, opened 2019-11-12
Create Collections 📚 supports options in Mongo scrapbooks 📔

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1222

Today, createCollection takes all arguments and stringifies them.  That leads to this valid Mongo query:

``` 
db.createCollection("teste", {
    validator: {
        jsonSchema: {
            bsonType: "object",
            required: ["teste1", "teste2"],
            properties: {
                teste1: {
                    bsonType: "string",
                    description: "Deve ser uma string"
                },
                teste2: {
                    bsonType: "string",
                    description: "Deve ser uma string"
                },
            }
        }
    }
})
```

Being inserted like this:
image

With my changes, it is inserted normally:
image

I don't think Azure supports validators yet: 
image

But it does work for a local database:
`db.getCollectionInfos({name: 'teste3'})`

image

### DIFF
--- a/src/mongo/tree/MongoDatabaseTreeItem.ts
+++ b/src/mongo/tree/MongoDatabaseTreeItem.ts
@@ -112,7 +112,7 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
 
 		if (command.name === 'createCollection') {
 			// arguments are always string (ie collection name) whereas arguementObjects will be an Object (where as in arguments, it's a JSON string)
-			return withProgress(this.createCollection(command.arguments[0], command.argumentObjects[1]).then(() => JSON.stringify({ 'Created': 'Ok' })), 'Creating collection');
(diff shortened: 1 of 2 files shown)


src/mongo/tree/MongoDatabaseTreeItem.ts
```diff
@@ -111,15 +111,16 @@ export class MongoDatabaseTreeItem extends AzureParentTreeItem<IMongoTreeRoot> {
 		}
 
 		if (command.name === 'createCollection') {
-			return withProgress(this.createCollection(stripQuotes(command.arguments.join(','))).then(() => JSON.stringify({ 'Created': 'Ok' })), 'Creating collection');
+			// arguments are always string (ie collection name) whereas arguementObjects will be an Object (where as in arguments, it's a JSON string)
```
EricJizbaMSFT: Can you double check this comment? Feels like it has some duplicate info and also you spelled argu_**e**_ment wrong in one place

PR #1239, opened 2019-11-14
Delete MongoDB

v-wuzhai:


test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -83,4 +91,10 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
         const connectionString: string = await vscode.env.clipboard.readText();
         return await connectToMongoClient(connectionString, appendExtensionUserAgent());
     }
+
+    async function doesMongoDatabaseExist(): Promise<IDatabaseInfo | undefined> {
```
nturinski: Make databaseName1 a parameter that can be passed in that way we can reuse it for any other Database checks we do.  Eventually we can probably make it so the API is also a parameter so we can use it for SQL, Table, and Gremlin as well.

test/nightly/azureResourceMongoDB.test.ts
```diff
@@ -75,12 +83,18 @@ suite('MongoDB action', async function (this: ISuiteCallbackContext): Promise<vo
         assert.ifError(accountExists);
     });
 
-    async function getMongoClient(): Promise<MongoClient> {
+    async function getMongoClient(resourceName: string): Promise<MongoClient> {
```
nturinski: I think it'd be better to use a regexp to find the account name.  If we change the label name (the MongoDB part), it would break the tests.

Issue #1237, opened 2019-11-13
Update the API list to match the portal


fiveisprime:
Update SQL and MongoDB to match what's shown in the portal.

Portal:

![image](https://user-images.githubusercontent.com/1186948/68813484-cca54e00-062a-11ea-822b-f49bd40c663b.png)

VS Code

![image](https://user-images.githubusercontent.com/1186948/68813501-d4fd8900-062a-11ea-9223-71f87ece6a52.png)

PR #1244, opened 2019-11-18
🆙📅 Update API names 📛

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1237 

image

### DIFF
--- a/src/experiences.ts
+++ b/src/experiences.ts
@@ -10,12 +10,12 @@ export enum API {
     MongoDB = 'MongoDB',
     Graph = 'Graph',
     Table = 'Table',
-    DocumentDB = 'DocumentDB'
+    Core = 'Core'
 }
 
 export enum DBAccountKind {
     MongoDB = 'MongoDB',
-    GlobalDocumentDB = 'GlobalDocumentDB'
+    GlobalCore = 'GlobalCore'
 }
 
 export type CapabilityName = 'EnableGremlin' | 'EnableTable';
@@ -23,7 +23,7 @@ export type CapabilityName = 'EnableGremlin' | 'EnableTable';
 export function getExperienceFromApi(api: API): Experience {
     let info = experiencesMap.get(api);
     if (!info) {
-        info = { api: api, shortName: api, longName: api, kind: DBAccountKind.GlobalDocumentDB, tag: api };
+        info = { api: api, shortName: api, longName: api, kind: DBAccountKind.GlobalCore, tag: api };
     }
     return info;
 }
@@ -50,7 +50,7 @@ export function tryGetExperience(account: DatabaseAccount): Experience | undefin
     } else if (account.capabilities.find(cap => cap.name === 'EnableTable')) {
         return TableExperience;
     } else if (account.capabilities.length === 0) {
-        return SQLExperience;
+        return CoreExperience;
     }
 
     return undefined;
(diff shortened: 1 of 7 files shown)


src/experiences.ts
```diff
@@ -85,10 +85,10 @@ export function getExperienceQuickPick(api: API): IAzureQuickPickItem<Experience
 
 // Mongo is distinguished by having kind="MongoDB". All others have kind="GlobalDocumentDB"
 // Table and Gremlin are distinguished from SQL by their capabilities
-const SQLExperience: Experience = { api: API.DocumentDB, longName: "SQL", description: "(DocumentDB)", shortName: "SQL", kind: DBAccountKind.GlobalDocumentDB, tag: "Core (SQL)" };
-const MongoExperience: Experience = { api: API.MongoDB, longName: "MongoDB", shortName: "MongoDB", kind: DBAccountKind.MongoDB, tag: "Azure Cosmos DB for MongoDB API" };
-const TableExperience: Experience = { api: API.Table, longName: "Azure Table", shortName: "Table", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableTable', tag: "Azure Table" };
-const GremlinExperience: Experience = { api: API.Graph, longName: "Gremlin", description: "(Graph)", shortName: "Gremlin", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableGremlin', tag: "Gremlin (graph)" };
+const SQLExperience: Experience = { api: API.DocumentDB, longName: "Core", description: "(SQL)", shortName: "Core (SQL)", kind: DBAccountKind.GlobalDocumentDB, tag: "Core (SQL)" };
```
EricJizbaMSFT: I'm confused - why is shortName longer than longName? I honestly don't remember where each of these is used...

nturinski: I think it's only used here:
`const creatingMessage: string = `Creating Cosmos DB account "${wizardContext.accountName}" with API "${wizardContext.defaultExperience.shortName}"...`;`

And in the label for `AttachedAccountsTreeItem`. 
image

nturinski:
Is it worth renaming the events too? I know it'd probably mess with dashboards?

src/experiences.ts
EricJizbaMSFT: Right so the shortName for core probably looks weird as an attached account due to nested parenthesis. I would think it should just be "Core"

EricJizbaMSFT:
> Is it worth renaming the events too? I know it'd probably mess with dashboards?

User-facing strings are higher priority. No strong preference on changing the event names (but if you change them - I expect you to also update the dashboard and re-categorize all gdpr stuff)

nturinski:
@fiveisprime Said he wants it to remain SQL 🤷‍♂

src/experiences.ts
```diff
@@ -10,20 +10,20 @@ export enum API {
...
     Graph = 'Graph',
     Table = 'Table',
-    DocumentDB = 'DocumentDB'
+    Core = 'Core'
 }
 
 export enum DBAccountKind {
     MongoDB = 'MongoDB',
-    GlobalDocumentDB = 'GlobalDocumentDB'
+    GlobalCore = 'GlobalCore'
```
EricJizbaMSFT: Kind is used when creating a cosmos db account here: https://github.com/microsoft/vscode-cosmosdb/blob/master/src/tree/CosmosDBAccountWizard/CosmosDBAccountCreateStep.ts#L24

Most of your changes are just for labels, but this one likely affects functionality. Are we sure Azure supports "GlobalCore" as the kind when creating an account?

PR #1245, opened 2019-11-19
Check chained ⛓ property when executing command

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/981
Partially fixes https://github.com/microsoft/vscode-cosmosdb/issues/1193

There was an issue when trying to run `db.getCollection('my-collection').find({})`.  The MongoCommand parser was detecting the chained command, but there was no check on that property.  If a command is being chained, then we should just defer to the shell since we don't have chaining logic.

I want to refactor how we `executeCommand` at one point, but I couldn't think of a better way to fix this than to simply check the `chained` property when the `command.name === 'find'`


nturinski:
TODO: Add tests

PR #1259, opened 2019-12-10
Add a confirmation message for copy connection string

burkeholland:
When executing the "Copy Connection String" command from the Command Palette, there is no feedback that the operation succeeded. This is a little awkward as the developer isn't sure if anything has happened.

This is less of an issue on the right-click since we are accustomed to right-click actions being successful by virtue of the right click occurring without error. Since we don't distinguish between when a the connection string copy occurs via right-click vs command palette, I suggest that we always show a notification that the connection string copy has occurred successfully.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -112,11 +112,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
             await node.openInPortal();
         });
         registerCommand('cosmosDB.copyConnectionString', async (actionContext: IActionContext, node?: MongoAccountTreeItem | DocDBAccountTreeItemBase) => {
+            const message = 'The connection string has been copied to the clipboard';
             if (!node) {
                 node = await ext.tree.showTreeItemPicker(accountContextValues, actionContext);
             }
 
             await copyConnectionString(node);
+            await vscode.window.showInformationMessage(message);
         });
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -112,11 +112,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
             await node.openInPortal();
         });
         registerCommand('cosmosDB.copyConnectionString', async (actionContext: IActionContext, node?: MongoAccountTreeItem | DocDBAccountTreeItemBase) => {
+            const message = 'The connection string has been copied to the clipboard';
             if (!node) {
                 node = await ext.tree.showTreeItemPicker(accountContextValues, actionContext);
             }
 
             await copyConnectionString(node);
+            await vscode.window.showInformationMessage(message);
```
EricJizbaMSFT: nit: Could we remove the `await`? With the `await` it'll skew the telemetry for this action because a lot of users miss the notification, never dismiss it (which is the only time `showInformationMessage` actually returns) and telemetry won't be sent

Here's similar code: https://github.com/microsoft/vscode-cosmosdb/blob/da8b567e5418b4cc7b3c31ac05a09ab99a613fcc/src/tree/SubscriptionTreeItem.ts#L66

nturinski: @burkeholland ping 😃

burkeholland: Sorry! The holidays got me, y'all. 

Removed the await.

PR #1263, opened 2019-12-14
Check for chaining before executing command

nturinski:
A workaround for this issue https://github.com/microsoft/vscode-cosmosdb/issues/1193

You can now use `getCollection` to retrieve collections with odd naming conventions.  The real solution is to edit the grammar file to understand [] notation.

`getCollection` doesn't work today because `getCollection().find()` was parsed with the command.name as find, and would enter that part of the conditional rather than attempting to execute the command.

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -112,7 +112,7 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 			throw new Error(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
 		}
 
-		if (command.name === 'find') {
+		if (command.name === 'find' && !command.chained) {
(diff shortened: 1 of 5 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -112,7 +112,7 @@ async function executeCommand(activeEditor: vscode.TextEditor, database: MongoDa
 			throw new Error(`Error near line ${err.range.start.line}, column ${err.range.start.character}: '${err.message}'. Please check syntax.`);
 		}
 
-		if (command.name === 'find') {
+		if (command.name === 'find' && !command.chained) {
```
EricJizbaMSFT: Maybe add a comment in the code for this one, or just link to the bug. It's not immediately clear why `chained` matters without context

test/mongoGetCommand.test.ts
```diff
@@ -459,6 +459,15 @@ suite("scrapbook parsing Tests", () => {
         });
     });
 
+    test("test chained command: chain command ends with find", () => {
```
EricJizbaMSFT: Please remove

Issue #1096, opened 2019-04-26
Inserting ISODate causes an error with mongodb


Diggerfdf:
I noticed a problem when trying to use ISODate to set a date with a mongodb database.

When I set up the date in the shell, everything is fine. But when I try via VScode Azure, I got an error. The hour always is inserted with an offset of + 3 hours. I don't need to use the UTC time. But I'm at Sao Paulo, Brazil UTC -3.

This is an example with my date and time of birth:

``db.test.insert({
    'datetime': ISODate('1975-07-23T04:10:00')
})``

This is the result

``{
	"result": {
		"ok": 1,
		"n": 1
	},
	"ops": [
		{
			"datetime": "1975-07-23T07:10:00.000Z",
			"_id": "5cc38d66da0aca11a90af39a"
		}
	],
	"insertedCount": 1,
	"insertedIds": [
		"5cc38d66da0aca11a90af39a"
	]
}``

The time is (...)T07:10:00.000Z Why this is happening?

I'm using Kubuntu 18.04 fully upgraded, Azure Cosmos DB Version 0.10.1 and VScode 1.33.1 installed through anaconda.

StephenWeatherford:
Other issues:
1) Date(1) gives the wrong result
1) ISODate(1) should give an error but does not

PrashanthCorp:
TODO: 
1. Check to see if there's a way to get Mongo's ISO function call. 
2. Failing that, send command to the shell.

PrashanthCorp:
@StephenWeatherford: Date(1) and new Date(1) give different results in node since the Function cal does not take any parameters -> Date(1) is equivalent to Date(). If a user chooses to use `Date ` instead of `new Date` we shouldn't give the users an error.

nturinski:
Hi @Diggerfdf

Thanks very much for your patience.  This issue should be resolved now; please let me know if there are still issues with inserting ISODates.

Thank you and have a great weeekend! 😃

PR #1281, opened 2020-01-10
Include ISODate 📆 for constructorCall 👷‍♀️📞 and handle appropriately

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1096

The error was happening because ISODate was being handled the same as a Date object.  The string representation for a Date is `1975-07-23T04:10:00.000` as opposed to ISODate's `1975-07-23T04:10:00.000Z`.  So when Mongo inserted the date, it just assumed the timezone was whatever the computer's local timezone was set to.

Edit: Both `db.test.insert({ 'datetime': ISODate('1975-07-23T04:10:00') })` and `db.test.insert({ 'datetime': ISODate('1975-07-23T04:10:00Z') })` are valid timestrings.

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -348,35 +348,53 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
 		}
 		let argumentContextArray: mongoParser.ArgumentContext[] = filterType(argumentsToken.children, mongoParser.ArgumentContext);
 
-		let functionMap = { "ObjectId": this.objectIdToObject, "ISODate": this.dateToObject, "Date": this.dateToObject };
+		let functionMap = { "ObjectId": this.objectIdToObject, "ISODate": this.isodateToObject, "Date": this.dateToObject };
 		if (argumentContextArray.length > 1) {
 			let err: IParsedError = parseError(`Too many arguments. Expecting 0 or 1 argument(s) to ${constructorCall}`);
 			this.addErrorToCommand(err, ctx);
 			return {};
 		}
+
 		if (constructorCall.text in functionMap) {
 			let args = [ctx, argumentContextArray.length ? argumentContextArray[0].text : undefined];
 			return functionMap[constructorCall.text].apply(this, args);
 		}
+
 		let unrecognizedNodeErr: IParsedError = parseError(`Unrecognized node type encountered. Could not parse ${constructorCall.text} as part of ${child.text}`);
 		this.addErrorToCommand(unrecognizedNodeErr, ctx);
 		return {};
 	}
 
-	private dateToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, tokenText?: string): Object {
+	private dateToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, tokenText?: string, isIsodate: boolean = false): Date | {} {
 		let constructedObject: Date;
 		if (!tokenText) { // usage : ObjectID()
 			constructedObject = new Date();
 		} else {
 			try {
-				constructedObject = new Date(stripQuotes(tokenText));
+				tokenText = stripQuotes(tokenText);
+
+				// if the tokenText was an isodate, it needs a Z suffix for indication
+				if (isIsodate) {
+					tokenText += 'Z';
+				}
+
+				constructedObject = new Date(tokenText);
 			} catch (error) {
 				let err: IParsedError = parseError(error);
 				this.addErrorToCommand(err, ctx);
 				return {};
 			}
 		}
-		return { $date: constructedObject.toString() };
+
+		return constructedObject;
+	}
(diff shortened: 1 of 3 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -348,35 +348,55 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
 
-		let functionMap = { "ObjectId": this.objectIdToObject, "ISODate": this.dateToObject, "Date": this.dateToObject };
+		let functionMap = { "ObjectId": this.objectIdToObject, "ISODate": this.isodateToObject, "Date": this.dateToObject };
 		if (argumentContextArray.length > 1) {
 			let err: IParsedError = parseError(`Too many arguments. Expecting 0 or 1 argument(s) to ${constructorCall}`);
 			this.addErrorToCommand(err, ctx);
 			return {};
 		}
+
 		if (constructorCall.text in functionMap) {
```
EricJizbaMSFT: This `functionMap` thing is really weird to me. I feel like it'll make our lives easier if you just do a simple `switch` like this:

```typescript
const tokenText: string | undefined = argumentContextArray.length ? argumentContextArray[0].text : undefined;
switch (constructorCall.text) {
	case 'ObjectId':
		return this.objectIdToObject(ctx, tokenText);
	case 'ISODate':
		return this.dateToObject(ctx, tokenText, true /* isIsoDate */);
	case 'Date':
		return this.dateToObject(ctx, tokenText);
	default:
		// ...
}
```

test/mongoGetCommand.test.ts
```diff
@@ -459,14 +459,50 @@ suite("scrapbook parsing Tests", () => {
...
+                    "date": {
+                        "$date": "2018-05-01T00:00:00.000Z"
+                    },
+                    "name": "Stephen"
+                }
+            ]
+        });
+    });
+
+    test("ISODate 3", () => {
```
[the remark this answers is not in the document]
nturinski: image

src/mongo/MongoScrapbook.ts
```diff
@@ -347,37 +347,68 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+		}
+	}
+
+	private dateToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, tokenText?: string): { $date: Date | {} } {
+		let stringDate: Date | {} = this.tryToConstructDate(ctx, tokenText);
+		if (stringDate instanceof Date) {
+			return { $date: stringDate.toISOString() };
+		}
+
+		return { $date: stringDate };
```
EricJizbaMSFT: I think the return type of `dateToObject` should be `{ $date: string } | {}`. 

No need to add the "$date" stuff if it returned an empty object (aka in the error case below)

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -183,7 +183,9 @@ export class MongoCollectionTreeItem extends AzureParentTreeItem<IMongoTreeRoot>
 		if (!document) {
 			throw new Error("The insert command requires at least one argument");
 		}
-		const insertResult = await this.collection.insert(document);
+
+		// insert is deprecated, so use insertOne instead
+		const insertResult = await this.collection.insertOne(document);
```
EricJizbaMSFT: Was this intentional? Seems unrelated

src/mongo/MongoScrapbook.ts
```diff
@@ -347,37 +347,68 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
...
+				let unrecognizedNodeErr: IParsedError = parseError(`Unrecognized node type encountered. Could not parse ${constructorCall.text} as part of ${child.text}`);
+				this.addErrorToCommand(unrecognizedNodeErr, ctx);
+				return {};
+		}
+	}
+
+	private dateToObject(ctx: mongoParser.ArgumentContext | mongoParser.PropertyValueContext, tokenText?: string): { $date: Date | {} } {
+		let stringDate: Date | {} = this.tryToConstructDate(ctx, tokenText);
+		if (stringDate instanceof Date) {
+			return { $date: stringDate.toISOString() };
```
EricJizbaMSFT: Previously it did `toString` instead of `toISOString`. Honestly I'm not sure what's the expected behavior here, but I would err on the side of keeping it consistent with previous releases

src/mongo/tree/MongoCollectionTreeItem.ts
nturinski: It was just something I noticed when trying to resolve this issue (since I thought the bug might have stemmed from using an unsupported insert command).  I can remove it though

EricJizbaMSFT: I'd prefer if you removed it just so I don't have to investigate that file haha. At the moment I'm not exactly sure what's going on there

PR #1316, opened 2020-01-28
Turn on several auto-fixable lint rules 🧹

EricJizbaMSFT:
I basically enabled all lint rules we have in functions repo, auto-fixed everything I could, and re-disabled any rules that were still causing problems.

The kind of rules this fixed:
1. Spaces instead of tabs
1. Order of public/private methods
1. Semicolons
1. Don't await a non-promise
1. No unnecessary variable


package.json
```diff
@@ -1,955 +1,956 @@
...
+                    ]
+                }
+            }
+        }
+    },
+    "scripts": {
+        "vscode:prepublish": "npm run webpack-prod",
+        "build": "tsc -p ./",
+        "compile": "tsc -watch -p ./",
+        "pa
```
nturinski: Why did this get added as a dependency?  It always existed in the `extension.bundle.ts`

EricJizbaMSFT: Because one of the rules that got turned on was "no-implicit-dependencies" https://palantir.github.io/tslint/rules/no-implicit-dependencies/

nturinski:
Just curious, but how come static functions got changed to just be functions existing outside of the class?

EricJizbaMSFT:
Because some rule re-ordered private static methods above public non-static methods and that bothered me. Also I just prefer functions over static methods

PR #1337, opened 2020-02-26
Tree view for postgres

### LINKED ISSUE #1336: Tree View for Azure PostgreSQL accounts

neelip:
Fixes #1336 
- Complete tree view for the azure postgres accounts, doesn't include --> context menus and document views
- Need to add account connection details as env variables POSTGRES_USERNAME, POSTGRES_PASSWORD, and POSTGRES_SSL to access the postgres databases; POSTGRES_SSL should the string "true" or "false"


src/postgres/ClientConfigClass.ts
```diff
@@ -0,0 +1,41 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { ClientConfig } from 'pg';
+
+export class ClientConfigClass implements ClientConfig {
```
EricJizbaMSFT: I think this class is overkill. It appears it's only _really_ used in `PostgreSQLDatabaseTreeItem` here:
```
public async connectToDb(): Promise<Schema[]> {
        const accountConnection = await connectToPostgresClient(this.clientConfig);
        const db = await pgStructure(accountConnection);
        return db.schemas;
    }
```

But we would normally create a simple object and pass it in like this:
```
public async connectToDb(): Promise<Schema[]> {
        const accountConnection = await connectToPostgresClient({
            host: this.host,
            database: this.databaseName,
            // etc.
        });
        const db = await pgStructure(accountConnection);
        return db.schemas;
    }
```

src/postgres/config.ts
```diff
@@ -0,0 +1,6 @@
+// Enter these values to enable postgres account view
+export const config = {
+    user: '<username>@<postgres_account_name>', //Add username@accountname here, username is the Azure username
```
EricJizbaMSFT: We prefer to keep master brach in a ship-ready state and this username/password logic seems pretty far from ship-ready. We have two options:
1. Wait for you to implement better username/password logic to merge this PR
1. Hide all of the Postgres stuff behind a feature flag, implement username/password stuff in a separate PR, and remove the feature flag once that's done

I guess I would prefer option 2 since this PR is already pretty big and I don't want to delay merging even longer. Here's an example PR where I'm _removing_ a feature flag: https://github.com/microsoft/vscode-azurestorage/pull/562. The important part is just the VS Code setting `azureStorage.enableViewInFileExplorer` which you should emulate

src/postgres/connectToPostgresClient.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Client } from 'pg';
+
+export async function connectToPostgresClient(clientConfig): Promise<Client> {
+
+    const client = new Client(clientConfig);
+    try {
+        return client;
+    } catch (err) {
+        const error = <{ message?: string, name?: string }>err;
```
EricJizbaMSFT: I'm guessing you copied this function from somewhere else (mongo?) but it's actually not doing anything for Postgres. You're catching the error and immediately throwing it. You can remove the try/catch block and since you're only left with one line of code I would just get rid of `connectToPostgresClient` entirely

src/postgres/tree/IPostgreSQLTreeRoot.ts
```diff
@@ -0,0 +1,11 @@
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ISubscriptionContext } from "vscode-azureextensionui";
+
+export interface IPostgreSQLTreeRoot extends ISubscriptionContext {
+    isEmulator: boolean;
```
EricJizbaMSFT: Uh does postgres support an emulator? I think that's Cosmos specific, in which case you can remove all emulator code from Postgres

src/postgres/tree/PostgreSQLAccountTreeItem.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.host = databaseAccount.fullyQualifiedDomainName;
+    }
+
+    // overrides ISubscriptionContext with an object that also has Mongo info
+    public get root(): IPostgreSQLTreeRoot {
+        return this._root;
+    }
+
+    public get iconPath(): string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } {
+        return getThemeAgnosticIconPath('CosmosDBAccount.svg');
```
EricJizbaMSFT: You should get a Postgres icon for this instead of cosmos. See here: https://cds-icons.azurewebsites.net/

src/tree/SubscriptionTreeItem.ts
```diff
@@ -27,14 +31,32 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
 
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
+
+        let treeItemPostgres: AzExtTreeItem[];
+        let treeItem: AzExtTreeItem[];
+
+        //Postgres
+        const clientPostgres: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
+        const accountsPostgres: ServerListResult = await clientPostgres.servers.list();
+        const databasesPostgres = await clientPostgres.databases.listByServer("myResourceGroup", accountsPostgres[0].name);
```
EricJizbaMSFT: As a general rule, you want to delay calls until you actually need the information. In this case, you don't need to list the databases until the user expands the `PostgreSQLAccountTreeItem`. Imagine if the user has hundreds of accounts and each one is listing all of it's databases here - that could take a long time even if the user never expands a single account tree item. This could should probably live in `PostgreSQLAccountTreeItem.loadMoreChildrenImpl`.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -27,14 +31,32 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         const accounts: DatabaseAccountsListResult = await client.databaseAccounts.list();
-        return await this.createTreeItemsWithErrorHandling(
+        treeItem = await this.createTreeItemsWithErrorHandling(
             accounts,
             'invalidCosmosDBAccount',
             async (db: DatabaseAccount) => await this.initChild(client, db),
             (db: DatabaseAccount) => db.name
         );
+
+        treeItem.push(treeItemPostgres.pop());
```
EricJizbaMSFT: This would only add the first tree item. Something like this will add all items: `treeItems.push(...treeItemPostgres)`

src/tree/SubscriptionTreeItem.ts
```diff
@@ -71,32 +93,46 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return typeof contextValue !== 'string' || !/attached/i.test(contextValue);
     }
 
-    private async initChild(client: CosmosDBManagementClient, databaseAccount: DatabaseAccount): Promise<AzureTreeItem> {
-        const experience = tryGetExperience(databaseAccount);
-        const resourceGroup: string = azureUtils.getResourceGroupFromId(databaseAccount.id);
-        const accountKindLabel = getExperienceLabel(databaseAccount);
-        const label: string = databaseAccount.name + (accountKindLabel ? ` (${accountKindLabel})` : ``);
+    private async initChild(client: CosmosDBManagementClient | PostgreSQLManagementClient, databaseAccount?, databasesPostgres?): Promise<AzureTreeItem> {
```
EricJizbaMSFT: I would rename `initChild` to `initCosmosDBChild` and make this method exclusive to Cosmos. The postgres logic looks simple and unrelated

src/vscode-cosmosdb.api.d.ts
```diff
@@ -59,7 +59,7 @@ export interface DatabaseTreeItem extends DatabaseAccountTreeItem {
 
 export type CosmosDBResourceType = 'DatabaseAccount' | 'Database';
 
-export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table';
+export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table' | 'Postgres';
```
EricJizbaMSFT: I would keep this file as-is for now. It defines the API we use between the App Service extension and the Cosmos extension and we want to be careful about any changes because it's a lot harder to change an API after the fact. This won't be needed until you add support for the "Connections" scenario discussed in standup

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,49 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as _ from 'underscore';
```
EricJizbaMSFT: Remove this unused import. I thought we had lint rules that would automatically do this for you...I might look into that.

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as vscode from 'vscode';
+import { AzureTreeItem } from 'vscode-azureextensionui';
+import { getThemeAgnosticIconPath } from '../../constants';
+import { getDocumentTreeItemLabel } from '../../utils/vscodeUtils';
+import { IPostgreSQLTreeRoot } from './IPostgreSQLTreeRoot';
+import { PostgreSQLSchemaTreeItem } from './PostgreSQLSchemaTreeItem';
+
+export class IPostgresTable {
```
EricJizbaMSFT: Yeah we don't use classes for something as simple as this. Just use an interface and an object

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,49 @@
...
+    public _id: string;
+    constructor(id) {
+        this._id = id;
+    }
+}
+
+export class PostgreSQLTableTreeItem extends AzureTreeItem<IPostgreSQLTreeRoot> {
+    public static contextValue: string = "PostgresTable";
+    public readonly contextValue: string = PostgreSQLTableTreeItem.contextValue;
+    // public readonly commandId: string = 'cosmosDB.openDocument';
```
EricJizbaMSFT: remove commented line

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,49 @@
...
+    public document: IPostgresTable;
+    public readonly parent: PostgreSQLSchemaTreeItem;
+
+    private _label;
+
+    constructor(parent: PostgreSQLSchemaTreeItem, document: IPostgresTable) {
```
EricJizbaMSFT: I'm confused by the naming of table/document in this file. Please clean that up

src/experiences.ts
```diff
@@ -28,29 +29,45 @@ export function getExperienceFromApi(api: API): Experience {
     return info;
 }
 
-export function getExperienceLabel(account: DatabaseAccount): string {
-    const experience: Experience | undefined = tryGetExperience(account);
+export function getExperienceLabel(account?): string {
```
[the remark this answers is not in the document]
EricJizbaMSFT: Turned on the lint rule in this PR: https://github.com/microsoft/vscode-cosmosdb/pull/1346

src/postgres/tree/IPostgreSQLTreeRoot.ts
neelip: IPostgreSQLTreeRoot only contains isEmulator for now, so it will be empty once I remove that variable. Should I keep this class for any future additions or just use the super-type for now in its place?

src/postgres/tree/PostgreSQLAccountTreeItem.ts
neelip: So Postgres structure is Server > Database > Schema > Table...Is it necessary to change the icons for Schema and Table as well?

src/experiences.ts
```diff
@@ -89,6 +106,7 @@ const CoreExperience: Experience = { api: API.Core, longName: "Core", descriptio
 const MongoExperience: Experience = { api: API.MongoDB, longName: "Azure Cosmos DB for MongoDB API", shortName: "MongoDB", kind: DBAccountKind.MongoDB, tag: "Azure Cosmos DB for MongoDB API" };
 const TableExperience: Experience = { api: API.Table, longName: "Azure Table", shortName: "Table", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableTable', tag: "Azure Table" };
 const GremlinExperience: Experience = { api: API.Graph, longName: "Gremlin", description: "(graph)", shortName: "Gremlin", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableGremlin', tag: "Gremlin (graph)" };
+const PostgresExperience: Experience = { api: API.Postgres, longName: "PostgreSQL", description: "PostgreSQL", shortName: "Postgres", kind: DBAccountKind.Postgres, tag: "Postgres" };
```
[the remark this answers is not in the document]
neelip: I created two separate files one for each cosmos and postgres... and handling the common variables in another file called experience.ts. Does this sound reasonable?

src/postgres/tree/PostgreSQLTableTreeItem.ts
neelip: Table for postgres is similar to the document for the cosmosdb account types, so the functionality we use for documents works for tables in postgres. However, If we use "Table" instead of "Document", it wont align with some methods we are using...such as "getDocumentTreeItemLabel".  Should we keep using "Document" instead of "Table" to keep things consistent across other types or use "Table"?

neelip: Same thing happens with "Server" which is similar to the "DatabaseAccount/Account" in other types.

src/postgres/tree/IPostgreSQLTreeRoot.ts
EricJizbaMSFT: Just use the super-type

src/postgres/tree/PostgreSQLAccountTreeItem.ts
EricJizbaMSFT: That would be great if you can find them. But yeah my only _requirement_ is changing the icon here in `PostgreSQLAccountTreeItem.ts`

src/experiences.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah that sounds fine. Also don't add Postges here if you don't need to. I believe a lot of this stuff is only used when creating a Cosmos DB account, which you are going to support in a separate PR

src/postgres/tree/PostgreSQLTableTreeItem.ts
EricJizbaMSFT: I understand there will be some confusion around the names - just try your best. But also - make sure the stuff you're sharing actually makes sense. For example, does `getDocumentTreeItemLabel` still make sense to use for Postgres? I'm guessing it might be pretty specific to CosmosDB

I would've guess Table was more analogous to Collection. And a document was analogous to a row in the table

src/postgres/tree/PostgreSQLSchemaTreeItem.ts
```diff
@@ -0,0 +1,45 @@
...
+        return getThemeAgnosticIconPath('Collection.svg');
+    }
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<PostgreSQLTableTreeItem[]> {
+        const tables: Table[] = this.schema.tables;
+        return tables.map(table => new PostgreSQLTableTreeItem(this, <IPostgresTable>{ _id: table.oid, _name: table.name }));
```
EricJizbaMSFT: I would just pass `table` as-is instead of creating a new type `IPostgresTable`

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,42 @@
...
+    private _label: string;
+
+    constructor(parent: PostgreSQLSchemaTreeItem, table: IPostgresTable) {
+        super(parent);
+        this.table = table;
+        this._label = table._name;
+    }
+
+    public get id(): string {
+        // tslint:disable-next-line:no-non-null-assertion
```
EricJizbaMSFT: Just remove the non-null-assertion, it's not necessary

src/postgres/tree/PostgreSQLTableTreeItem.ts
```diff
@@ -0,0 +1,42 @@
...
+        return String(this.table!._id);
+    }
+
+    public get label(): string {
+        return this._label;
```
EricJizbaMSFT: No need for _label member. Just use `this.table._name`

src/postgres/tree/PostgreSQLSchemaTreeItem.ts
```diff
@@ -0,0 +1,45 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { Schema, Table } from 'pg-structure';
+import * as _ from 'underscore';
+import * as vscode from 'vscode';
+import { AzureParentTreeItem, ISubscriptionContext } from 'vscode-azureextensionui';
+import { getThemeAgnosticIconPath } from '../../constants';
+import { IPostgresTable, PostgreSQLTableTreeItem } from './PostgreSQLTableTreeItem';
+
+export class PostgreSQLSchemaTreeItem extends AzureParentTreeItem<ISubscriptionContext> {
```
EricJizbaMSFT: it's a little weird that in some cases you use `postgres` (like the context value) and some you use `postgresSQL` (for class names). Can we pick a consistent one to use internally? I guess `postgres` seems easier since it has fewer characters. Would also be fine with `pg` if we want to use the least number of characters. Of course, for any user-visible string, it should be the official "PostgreSQL"

src/CosmosDBExperiences.ts
```diff
@@ -5,31 +5,18 @@
...
     let info = experiencesMap.get(api);
     if (!info) {
         info = { api: api, shortName: api, longName: api, kind: DBAccountKind.GlobalDocumentDB, tag: api };
     }
     return info;
 }
 
-export function getExperienceLabel(account: DatabaseAccount): string {
-    const experience: Experience | undefined = tryGetExperience(account);
+export function getExperienceLabel_cosmosdb(account: DatabaseAccount): string {
```
EricJizbaMSFT: I've never seen a function name with underscores like `getExperienceLabel_cosmosdb`. We usually stick to just camelCase. In this example, I probably would've done getCosmosDBExperienceLabel

src/tree/SubscriptionTreeItem.ts
```diff
@@ -27,14 +33,33 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
+
+        let treeItemPostgres: AzExtTreeItem[];
+        let treeItem: AzExtTreeItem[];
+
+        //Postgres
+        const clientPostgres: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
```
EricJizbaMSFT: Suggest renaming `clientPostgres` to `pgClient`

src/tree/SubscriptionTreeItem.ts
```diff
@@ -27,14 +33,33 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+        const accountsPostgres: ServerListResult = await clientPostgres.servers.list();
```
EricJizbaMSFT: Stay consistent with the `server` terminology. How about `pgServers` for this variable

src/tree/SubscriptionTreeItem.ts
```diff
@@ -99,4 +124,12 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             }
         }
     }
+
+    private async initPostgresChild(account: Server, databases: Databases): Promise<AzureTreeItem> {
+        const accountKindLabel = getExperienceLabel_postgres(account);
```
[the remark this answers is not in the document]
neelip: Do you mean revert back to a single experience.ts file like before with all the cosmos experiences code in it?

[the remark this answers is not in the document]
EricJizbaMSFT: Yes it should be a single file, but maybe make the name `CosmosDBExperiences.ts` instead of `experience.ts`. The whole point is that "Experience" is a concept unique to Cosmos DB. I don't see us ever using that stuff for Postgres and all of the current code in `PostgresExperiences.ts` is unnecessary

src/postgres/tree/PostgreSQLAccountTreeItem.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export class PostgreSQLAccountTreeItem extends AzureParentTreeItem<ISubscriptionContext> {
+    public static contextValue: string = "cosmosDBPostgresServer";
+    public readonly contextValue: string = PostgreSQLAccountTreeItem.contextValue;
+    public readonly childTypeLabel: string = "Database";
+    public readonly id: string;
+    public readonly label: string;
+    public readonly connectionString: string;
+    public readonly host: string;
+    public readonly databases: Databases;
```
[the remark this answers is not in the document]
neelip: PostgreSQLManagementClient is initialized in Subscription...Tree already and is used only once to get the server object and the databases object. I feel like its redundant to use PostgreSQLManagementClient again in this file. If we are trying to reduce the number of property variables, we can just pass client instead of server and databases and then access them later in the file as needed?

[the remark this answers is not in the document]
neelip: I fixed this and along with removing initPostgres function from Subscription...Tree.ts, there are only 3 property variables for this file now --> server, databases and label and only 2 parameters for the constructor. Is that better now, or should I pass client instead of both server and databases?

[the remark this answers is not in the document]
EricJizbaMSFT: After more thought, here's my reasoning:

1. The name "databases" confuses me - I would expect that to be the result of `listDatabases`, not the client itself. Something like "databaseClient" would make more sense, but the type itself is called "Databases" and you can't change that
1. I try to avoid caching things like clients because they deal with credentials. I don't know how it works under the covers, but theoretically there could be a refresh token that only gets updated when you create a new client and thus it could expire if you keep an old client around.
1. That's just how we do it elsewhere in our repos 🤷‍♂️ This is the worst reason in the list, but if either option results in about the same lines of code, it serves as a decent tie-breaker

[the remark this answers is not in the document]
neelip: Makes sense. Databases is confusing as a name especially being passed to the PostgreSQLAccount... class. I fixed it with only passing client and the server to the PostgresSQLAccount... and handling databases object in the loadMoreChildren... method.

src/postgres/tree/PostgreSQLTableTreeItem.ts
neelip: I am getting a "Forbidden non null assertion" when I remove it

neelip: Never mind, fixed it

src/CosmosDBExperiences.ts
neelip: I fixed it with the code in the master, expect the file name is CosmosDBExperiences and removed PostgresExperiences given an earlier suggestion.  Is required to have method names also specific to cosmosdb or can we keep them as is?

src/postgres/tree/PostgreSQLSchemaTreeItem.ts
neelip: Should we have the class names start with PG... as well or is Postgres ok?

EricJizbaMSFT: I don't care if you use `Pg` or `Postgres`, but I would prefer if it was consistent

package.json
```diff
@@ -579,6 +579,10 @@
                     "command": "cosmosDB.openInPortal",
```
EricJizbaMSFT: You still have `.DS_Store` files in this PR. Please remove them - you might need to google how to do it if git is giving you problems (sometimes happens with files that you added before you changed your gitignore)

package.json
```diff
@@ -845,13 +853,20 @@
...
         "documentdb": "^1.14.2",
         "fs-extra": "^4.0.2",
         "mongodb": "^3.3.2",
         "mongodb-extended-json": "^1.10.0",
         "ms-rest": "^2.2.1",
         "ms-rest-azure": "^2.3.1",
+        "node-uuid": "1.4.8",
+        "pg": "^7.18.1",
+        "pg-structure": "^5.10.7",
+        "socket.io": "^1.7.3",
```
EricJizbaMSFT: Some of these were removed in https://github.com/microsoft/vscode-cosmosdb/pull/1362. Make sure the diff here only shows the postgres packages you need

src/extension.ts
```diff
@@ -62,7 +63,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         registerGraphCommands();
         const codeLensProvider = registerMongoCommands(editorManager);
 
-        const accountContextValues: string[] = [GraphAccountTreeItem.contextValue, DocDBAccountTreeItem.contextValue, TableAccountTreeItem.contextValue, MongoAccountTreeItem.contextValue];
+        const accountContextValues: string[] = [GraphAccountTreeItem.contextValue, DocDBAccountTreeItem.contextValue, TableAccountTreeItem.contextValue, MongoAccountTreeItem.contextValue, PostgresServerTreeItem.contextValue];
```
EricJizbaMSFT: There's several commands here that postgres doesn't support. Suggest splitting it up like this:
```
const cosmosDBAccountContextValues: string[] = [GraphAccountTreeItem.contextValue, DocDBAccountTreeItem.contextValue, TableAccountTreeItem.contextValue, MongoAccountTreeItem.contextValue];
const allTopLevelContextValues: string[] = [...cosmosDBAccountContextValues, PostgresServerTreeItem.contextValue];
```

I'm not a huge fan of the variable names I came up with, but you get the idea. You should only pass `allTopLevelContextValues` to `showTreeItemPicker` for commands that postgres also supports - I believe only "Open in Portal".

For example, if you run "Cosmos DB: Delete Account" from the command palette, a postgres account should not show up

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -0,0 +1,58 @@
...
+    public readonly client: PostgreSQLManagementClient;
+    public readonly databases: Databases;
+    public readonly label: string;
+    public readonly server: Server;
+
+    constructor(parent: AzureParentTreeItem, client: PostgreSQLManagementClient, server: Server) {
+        super(parent);
+        this.client = client;
+        this.server = server;
+        this.label = server.name + ` (PostgreSQL)`;
```
EricJizbaMSFT: There's a separate `description` property that you should use for `PostgreSQL` instead of adding it to the label. It will handle the formatting for you

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -0,0 +1,58 @@
...
+import { azureUtils } from '../../utils/azureUtils';
+import { PostgresDatabaseTreeItem } from './PostgresDatabaseTreeItem';
+import { PostgresSchemaTreeItem } from './PostgresSchemaTreeItem';
+import { PostgresTableTreeItem } from './PostgresTableTreeItem';
+
+export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionContext> {
+    public static contextValue: string = "postgresServer";
+    public readonly contextValue: string = PostgresServerTreeItem.contextValue;
+    public readonly childTypeLabel: string = "Database";
+    public readonly client: PostgreSQLManagementClient;
```
EricJizbaMSFT: To continue the previous thread, I still don't want you to cache `client` here in this class. This is the main reason from before:

> I try to avoid caching things like clients because they deal with credentials. I don't know how it works under the covers, but theoretically there could be a refresh token that only gets updated when you create a new client and thus it could expire if you keep an old client around.

To expand on that, consider this possibility:
1. User expands their subscription tree item, causing us to list pg servers and create the client
1. User walks away for a day
1. User comes back and expands a pg server tree item

It's still using the client you created a day ago. I'm okay passing the client around if it's used immediately, but tree expansion does not fit under that case. It's easy enough to just create a new client below and actually results in

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -0,0 +1,58 @@
...
+    public readonly databases: Databases;
```
EricJizbaMSFT: Remove `databases` since it's no longer used

src/postgres/tree/PostgresServerTreeItem.ts
neelip: Do you mean the "description" property for Experience interface or one from the AzureExtTree classes?  Also this label property for Cosmos accounts is created in SubscriptionTreeItem as such...
```
       const accountKindLabel = getExperienceLabel(databaseAccount);
       const label: string = databaseAccount.name + (accountKindLabel ? ` (${accountKindLabel})` : ``);
```
I am not sure I understand how to incorporate description to create this label.

EricJizbaMSFT: AzExtTreeItem class:

https://github.com/microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L169

EricJizbaMSFT: `SubscriptionTreeItem` is probably doing it wrong. Here's a tree item that does it: https://github.com/microsoft/vscode-azureappservice/blob/138adfd6ff9dc65751a141c7b313c05968ff9127/src/explorer/SiteTreeItem.ts#L55

neelip: Is description the same as the label that we see next to each account name in the tree view such "<Account name> (SQL)" ? I am not seeing any association to these two in given code..

EricJizbaMSFT: Yes - per [the docs](https://github.com/microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L167)
> Additional information about a tree item that is appended to the label with the format `label (description)`

It also displays a little different in the treeItemPicker if `description` is used

package-lock.json
```diff
@@ -1317,7 +1369,7 @@
         },
         "chalk": {
             "version": "1.1.3",
-            "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
```
EricJizbaMSFT: You see how it's changing a bunch of https urls to http? That's due to an old bug in npm. Make sure you're on an up-to-date version of npm (I'm on 6.13.4) and follow these steps to fix: https://github.com/npm/npm/issues/20106#issuecomment-383770315

neelip: Updated npm to latest version in the next commit

PR #1365, opened 2020-03-16
Don't allow trailing whitespace in connection strings

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/994

### DIFF
--- a/src/tree/AttachedAccountsTreeItem.ts
+++ b/src/tree/AttachedAccountsTreeItem.ts
@@ -137,7 +137,7 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
                 validateInput = AttachedAccountsTreeItem.validateDocDBConnectionString;
             }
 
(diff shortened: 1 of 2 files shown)


src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -64,20 +64,24 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
+
         if (value && value.match(/^mongodb(\+srv)?:\/\//)) {
             return undefined;
         }
+
         return MONGO_CONNECTION_EXPECTED;
     }
 
     private static validateDocDBConnectionString(value: string): string | undefined {
+        value = value.trim();
```
EricJizbaMSFT: This repo doesn't have stictNullChecks turned on so you have to be extra careful. I'm pretty sure `value` could be undefined and you should handle that appropriately

PR #1369, opened 2020-03-17
Easier debugging in a webpacked world 🕸📦🌎

EricJizbaMSFT:
Same as https://github.com/microsoft/vscode-azurefunctions/pull/1973

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -3,10 +3,22 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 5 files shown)


gulpfile.ts
```diff
@@ -3,11 +3,23 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// tslint:disable:no-implicit-dependencies (this allows the use of dev dependencies)
```
nturinski: ?

EricJizbaMSFT: I see squigglies so I fix da squigglies

Screen Shot 2020-03-17 at 3 46 54 PM

Issue #966, opened 2018-11-26
Errored attached databases cause connections node to fail to parse/load


fiveisprime:
Repro:

1. In app service extension, connect to a database on Azure
1. Start local mongo `$ docker run --rm -d -p 27017:27017/tcp mongo:latest`
1. In the cosmos extension, attach to `mongodb://127.0.0.1:27017`
1. Stop local mongo `$ docker stop <id>` or `$ docker stop $(docker ps -q)` if it's the only container running
1. Reload vs code
1. Expand connections node to see mongo error

![image](https://user-images.githubusercontent.com/1186948/49048652-5ea11600-f190-11e8-971d-d7deb33e3d6c.png)

EricJizbaMSFT:
The key here is that the error in app service is for a db account that's not actually the connected one

PR #1370, opened 2020-03-17
Allow connections node to load when the emulator isn't running

wwlorey:
I wasn't sure if this was the best way to approach this but I gave it a shot.

It lets you view other connections in appservice even if the emulator node fails to load.

Screen Shot 2020-03-17 at 2 20 55 PM

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/966


nturinski:
@EricJizbaMSFT I'm not sure if the try/catch should be in  `findTreeItem` or in `searchDbAccounts`.  Since `searchDbAccounts` is supposed to return undefined if it can't find it, I feel like the try/catch should be here, and in case of an error, it should just swallow the error and return undefined.  That way we don't have to look for a specific error `MongoNetworkError` in `findTreeItem`.

This would also help in case the connected account that errors is a Core SQL or Postgres account, etc.

EricJizbaMSFT:
I agree with Nathan 👍

src/commands/api/findTreeItem.ts
```diff
@@ -83,34 +72,44 @@ export async function findTreeItem(query: TreeItemQuery): Promise<DatabaseAccoun
...
-
-            return new DatabaseAccountTreeItemInternal(expected, dbAccount);
+        }
+    } catch (error) {
+        const parsedError: IParsedError = parseError(error);
+        if (parsedError.errorType === 'MongoNetworkError' && !parsedError.message.includes(expected.accountId)) {
+            // Ignore this error since it doesn't pertain to the account we're searching for
+            // https://github.com/microsoft/vscode-cosmosdb/issues/966
+        } else {
+            throw error;
```
nturinski: If you throw the error, we will have this same issue for when users have SQL Core or Postgres attached accounts that fail.  As of now, this only fixes an errors that Mongo throws.

I think you can just check if the expected accountId is included in the error and throw the error then.  If it's not, then you can return undefined in the catch block because we don't want unrelated errors to screw up the Connections node.

nturinski: image
 LOL uh-oh.  Got the anger sharks swimming in my head!!

EricJizbaMSFT: I feel like you guys might be nervous to swallow all errors here because I usually make comments like [this](https://github.com/microsoft/vscode-azurestorage/pull/618#discussion_r391340142) telling you to check `parsedError.errorType`, but sometimes there's exceptions to that rule. For example [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/src/templates/script/parseScriptTemplates.ts#L242). I think it's fine to swallow all errors in this case since you don't want _any_ error to block you from searching other dbAccounts

PR #1380, opened 2020-03-20
Turn on strictNullChecks

EricJizbaMSFT:
There's a few files I'm not that familiar with so I just used non-null-assertions, but otherwise this should bring Cosmos DB up to par with our other repos, mainly by utilizing `nonNullValue` and `nonNullProp`.


src/mongo/services/completionItemProvider.ts
```diff
@@ -93,86 +96,86 @@ export class CompletionItemsVisitor extends MongoVisitor<Promise<CompletionItem[
...
                 });
             });
     }
 
     private getFunctionName(ctx: ParseTree): string {
-        let parent = ctx.parent;
+        let parent = ctx.parent!;
         if (!(parent && parent instanceof mongoParser.ArgumentContext)) {
-            return null;
+            return null!;
```
wwlorey: Why is the trailing `!` required for these cases?

src/mongo/mongoConnectionStrings.ts
```diff
@@ -35,14 +36,13 @@ export function getDatabaseNameFromConnectionString(connectionString: string): s
     return undefined;
 }
 
-export function addDatabaseToAccountConnectionString(connectionString: string, databaseName: string): string | undefined {
+export function addDatabaseToAccountConnectionString(connectionString: string, databaseName: string): string {
     try {
         return connectionString.replace(mongoConnectionStringRegExp, `$1\/${databaseName}`);
     } catch (error) {
         // Shouldn't happen, but ignore if does
+        return connectionString;
```
wwlorey: Is this just better behavior in the case of an error?

EricJizbaMSFT: This function is only used once and it expects the resulting connection string to be of type `string`, not `string | undefined`. My choices were to return `connectionString` here or call `nonNullValue` there. I didn't write this code originally, but I felt like calling `nonNullValue` didn't jive with "Shouldn't happen, but ignore if does" since it would immediately throw an error. I figure the better way to "ignore" this error is to use the original connection string, which might still work even if it's in a weird format that we can't parse

PR #1381, opened 2020-03-23
Add support for Postgres username & password input

wwlorey:
This allows you to enter credentials and connect to accounts that have SSL disabled. I'll address SSL in a separate PR


src/postgres/tree/PostgresEnterCredentialsTreeItem.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
+import { localize } from "../../utils/localize";
+import { PostgresDatabaseTreeItem } from "./PostgresDatabaseTreeItem";
+
+export class PostgresEnterCredentialsTreeItem extends AzureTreeItem<ISubscriptionContext> {
```
nturinski: I don't think you'd need a whole treeItem class for this.  For connecting to Github, I used a generic treeItem like so: https://github.com/microsoft/vscode-azuretools/blob/e8e7cf2b8733f6b6ffc380a01aa2996a260caf3c/appservice/src/tree/DeploymentsTreeItem.ts#L84-L89

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -40,16 +60,107 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+                        username = server.username;
+                        password = await this._keytar.getPassword(this._serviceName, this._serverId) || undefined;
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (!username || !password) {
+            if (warnBeforePrompting) {
```
nturinski: What's the point of warning the user before prompting them?

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -40,16 +60,107 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            })];
+        }
+    }
+
+    public async getCredentials(forcePrompt: boolean, warnBeforePrompting: boolean): Promise<{ username: string, password: string }> {
+        let username: string | undefined;
+        let password: string | undefined;
+
+        if (!forcePrompt) {
+            const storedValue: string | undefined = ext.context.globalState.get(this._serviceName);
```
nturinski: I think this logic is complicated enough to warrant it's own function. Maybe like `getCredentialsFromKeytar` or something

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -40,16 +60,107 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+                commandId: 'cosmosDB.getPostgresCredentials'
+            })];
+        }
+    }
+
+    public async getCredentials(forcePrompt: boolean, warnBeforePrompting: boolean): Promise<{ username: string, password: string }> {
+        let username: string | undefined;
+        let password: string | undefined;
+
+        if (!forcePrompt) {
```
nturinski: How come we wouldn't use the cached password sometimes?

src/postgres/tree/PostgresDatabaseTreeItem.ts
wwlorey: Matt requested that we prompt before asking for credentials so what's happening is more obvious

wwlorey: But that warning didn't make sense when you entered credentials via the cmd palette command so I removed it in that case

src/postgres/tree/PostgresDatabaseTreeItem.ts
wwlorey: If you call `getPostgresCredentials` via the command palette it will override the cached credentials

package.json
```diff
@@ -337,6 +338,11 @@
                 "category": "Cosmos DB",
                 "command": "cosmosDB.openStoredProcedure",
                 "title": "Open Stored Procedure"
+            },
+            {
+                "category": "PostgreSQL",
+                "command": "cosmosDB.getPostgresCredentials",
+                "title": "Enter Server Credentials"
```
wwlorey: The old title was redundant: 
Screen Shot 2020-03-26 at 10 42 59 AM

New title:
Screen Shot 2020-03-26 at 10 42 27 AM

src/postgres/tree/PostgresDatabaseTreeItem.ts
nturinski: Uhh, but why?

wwlorey: It just seemed like weird behavior to select enter credentials and then fallback to the cached credentials without entering anything

wwlorey:
If the password is changed and you reload a database, here's the behavior:

Screen Shot 2020-03-26 at 11 30 16 AM

Clicking on the "Enter server credentials..." tree item calls `getPostgresCredentials` which updates the cache. Does that behavior look ok?

src/postgres/tree/PostgresDatabaseTreeItem.ts
nturinski: Ahh, yeah that makes sense.

src/postgres/getPostgresCredentials.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ext } from "../extensionVariables";
+import { nonNullProp } from "../utils/nonNull";
+import { PostgresDatabaseTreeItem } from "./tree/PostgresDatabaseTreeItem";
+
+export async function getPostgresCredentials(context: IActionContext, treeItem?: PostgresDatabaseTreeItem | GenericTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+    }
+
+    if (treeItem instanceof GenericTreeItem) {
```
nturinski: nit: This could be an else if since the treeItemPicker would never pick a GenericTreeItem type

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -4,24 +4,44 @@
...
     public readonly contextValue: string = PostgresDatabaseTreeItem.contextValue;
     public readonly childTypeLabel: string = "Schema";
     public readonly databaseName: string;
     public readonly parent: PostgresServerTreeItem;
 
+    private readonly _serviceName: string = "ms-azuretools.vscode-cosmosdb.postgresPasswords";
+    private _keytar: KeyTar | undefined;
+    private _usernameSuffix: string;
+    private _usernamePlaceholder: string;
+    private _usernameRegex: RegExp;
```
nturinski: I don't think you really need to declare this as a private variable.  Should be able to just define it when you use it to validate the username.

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -4,24 +4,44 @@
...
+    private _serverId: string;
+
     constructor(parent: PostgresServerTreeItem, databaseName: string) {
         super(parent);
         this.databaseName = databaseName;
+        this._keytar = tryGetKeyTar();
+        this._usernameSuffix = `@${this.parent.server.name}`;
```
nturinski: Probably don't need to make this a property this either since it's only used in the constructor.

wwlorey: It's used to make the username regex too is it ok to keep it a property?

nturinski: Oh right, haha you got rid of the regex property.

src/postgres/tree/PostgresDatabaseTreeItem.ts
EricJizbaMSFT: I'm with Nathan on this. Seems like the prompt in the input box should have the message being shown here

EricJizbaMSFT: If we want a separate prompt to make it more obvious what's going on, how about we count your nice tree item as that prompt:
Screen Shot 2020-03-27 at 11 22 53 AM

Aka don't prompt them automatically and just make them click that. It's the same amount of clicks as your current "Continue" button 😂.

PR #1385, opened 2020-03-26
Add SSL support for Postgres

wwlorey:


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -62,8 +65,12 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         try {
             const { username, password } = await this.getCredentials(false, true);
 
-            const sslString: string | undefined = process.env.POSTGRES_SSL;
-            const ssl: boolean = sslString === 'true';
+            const ssl: ConnectionOptions = {
+                // Always provide the certificate since it is accepted even when SSL is disabled
+                // Certificate source: https://aka.ms/AA7wnvl
+                ca: fse.readFileSync(path.join(getResourcesPath(), 'certificates', 'BaltimoreCyberTrustRoot.crt.pem')).toString()
```
EricJizbaMSFT: Always use the async version when possible. `await fse.readFile` instead of `fse.readFileSync`

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -62,8 +65,12 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         try {
             const { username, password } = await this.getCredentials(false, true);
 
-            const sslString: string | undefined = process.env.POSTGRES_SSL;
-            const ssl: boolean = sslString === 'true';
+            const ssl: ConnectionOptions = {
+                // Always provide the certificate since it is accepted even when SSL is disabled
```
EricJizbaMSFT: Hmm I think I'd prefer to make `BaltimoreCyberTrustRoot` a constant in this file instead of it's own file in resources. Someone looking through "resources" might be confused about seeing a random certificate without the context that exists here (that link is helpful). It would also make this code simpler, although maybe put the constant at the bottom of this file for readability

wwlorey:
@EricJizbaMSFT 🤦‍♂️😭

PR #1386, opened 2020-03-26
Setup Postgres firewall rules

### LINKED ISSUE #1382: Firewall Rules Setup

wwlorey:
Fixes #1382


package.json
```diff
@@ -869,6 +869,7 @@
         "ms-rest-azure": "^2.3.1",
         "pg": "^7.18.1",
         "pg-structure": "^5.10.7",
+        "public-ip": "^4.0.0",
```
wwlorey: I wasn't sure if adding this dependency was the best way to go about finding the machine's IP. Does anyone know of a better way?

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -73,4 +75,40 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+
+        if (!existingFirewallRule || existingFirewallRule.startIpAddress !== ip || existingFirewallRule.endIpAddress !== ip) {
+            const newFirewallRule: FirewallRule = {
+                startIpAddress: ip,
+                endIpAddress: ip
+            };
+
+            const options: vscode.ProgressOptions = {
+                location: vscode.ProgressLocation.Notification,
+                title: localize('configuringFirewall', 'Configuring firewall for server "{0}"...', serverName)
```
EricJizbaMSFT: I would prompt first before doing something like this since it's security related and could take a while

package.json
EricJizbaMSFT: How about this package, which has zero dependencies and way more weekly downloads (the main two things I look at when picking something to use): https://www.npmjs.com/package/ip

wwlorey: I looked into that package and a few others but they all gave the private IP instead of the public one. `public-ip` was the only package I could find that provides a public ip

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -73,4 +76,46 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+        } catch (error) {
+            if (parseError(error).errorType === 'ResourceNotFound') {
+                // This firewall rule is not configured yet. Ignore error
+            } else {
+                throw error;
+            }
+        }
+
+        if (!existingFirewallRule || existingFirewallRule.startIpAddress !== ip || existingFirewallRule.endIpAddress !== ip) {
+            await ext.ui.showWarningMessage(
```
EricJizbaMSFT: Haha I would phrase this more as "Your current setup is not supported. Would you like to change it?" with a "Yes change it" button instead of a "We're going to change this for you" with a "I guess I'm along for the ride" button

package.json
EricJizbaMSFT: Interesting distinction - did you by chance find any best practice docs around firewalls for Postgres? I'm wondering how "recommended" it is to do this kind of stuff... as far as I understand it the ip address that you're talking about could easily change based on the whim of your internet service provider

src/postgres/tree/PostgresServerTreeItem.ts
wwlorey: I'll wait to change this until https://github.com/microsoft/vscode-cosmosdb/pull/1397 is merged, it'll touch the same code in the database tree item

package.json
wwlorey: I was going off of how the portal configures the client IP address:

Screen Shot 2020-03-31 at 12 55 11 PM

That being said, the address could change. I'm not sure if it would be better to update the existing firewall rule in that case or create a new one. Right now we just update it

EricJizbaMSFT:
Can you try that merge commit again because that diff is
yikes

package-lock.json
```diff
@@ -4908,7 +5044,7 @@
         },
         "load-json-file": {
             "version": "1.1.0",
-            "resolved": "https://registry.npmjs.org/load-json-file/-/load-json-file-1.1.0.tgz",
+            "resolved": "http://registry.npmjs.org/load-json-file/-/load-json-file-1.1.0.tgz",
```
EricJizbaMSFT, line 5047: You're having that https to http problem again. Make sure to follow the steps https://github.com/npm/npm/issues/20106#issuecomment-383770315

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -70,6 +70,8 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
                 const host: string = nonNullProp(this.parent.server, 'fullyQualifiedDomainName');
                 const clientConfig: ClientConfig = { user: username, password, ssl, host, port: 5432, database: this.databaseName };
                 const accountConnection: Client = new Client(clientConfig);
+
+                await this.parent.checkAndConfigureFirewall();
```
EricJizbaMSFT, line 74: Okay well now that we merged the other PRs, this behavior doesn't match anymore. Can we make this more similar to "Enter server credentials"? Aka it doesn't prompt automatically and just shows a "GenericTreeItem" if I need to set it up

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -73,4 +76,46 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+            );
+
+            const newFirewallRule: FirewallRule = {
+                startIpAddress: ip,
+                endIpAddress: ip
+            };
+
+            const options: vscode.ProgressOptions = {
+                location: vscode.ProgressLocation.Notification,
+                title: localize('configuringFirewall', 'Adding firewall rule for IP "{0}" to server "{1}"...', ip, serverName)
```
EricJizbaMSFT, line 113: How long exactly does this take? Should we add text like "This may take several minutes" to give people an idea or is it shorter than that?

wwlorey: This generally takes about 30 seconds, probably short enough to avoid that text

src/postgres/configurePostgresFirewall.ts
```diff
@@ -0,0 +1,20 @@
...
+import { GenericTreeItem, IActionContext } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
+import { nonNullProp } from "../utils/nonNull";
+import { PostgresServerTreeItem } from "./tree/PostgresServerTreeItem";
+
+export async function configurePostgresFirewall(context: IActionContext, treeItem?: PostgresServerTreeItem | GenericTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = await ext.tree.showTreeItemPicker(PostgresServerTreeItem.contextValue, context);
+    } else if (treeItem instanceof GenericTreeItem) {
+        treeItem = nonNullProp(nonNullProp(treeItem, 'parent'), 'parent');
```
EricJizbaMSFT, line 15: nit: I would say this is the slightly cleaner way of getting around this problem: https://github.com/microsoft/vscode-azuretools/blob/b37737c0550d5e8af177228d33142d469f1a27ef/appservice/src/tree/LogFilesTreeItem.ts#L34

Aka set the commandArgs on the GenericTreeItem to the parent or whatever so that this file always gets a `PostgresServerTreeItem`

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -61,6 +62,14 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
 
         if (username && password) {
             try {
+                if (!await this.parent.isFirewallConfigured()) {
```
EricJizbaMSFT, line 65: I'd rather not check for this up front for two reasons:
1. It could have performance implications (slowing down people that already have the firewall configured)
1. It could block people if our logic is incorrect. For example, it seems like `isFirewallConfigured` only searches for a specific firewall rule named "azureDatabasesForVSCode-publicIp", but couldn't I configure a rule myself on the portal with a different name?

Instead, if you return the GenericTreeItem in the catch block (when a firewall-related error occurs), it should solve both of those concerns.

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -73,4 +77,44 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+            if (parseError(error).errorType === 'ResourceNotFound') {
+                // The firewall rule is not configured yet
+                return false;
+            } else {
+                throw error;
+            }
+        }
+    }
+
+    public async configureFirewall(): Promise<void> {
```
EricJizbaMSFT, line 100: Since you already have a `configurePostgresFirewall.ts` file - I would prefer to put all of this logic there. The tree files can grow pretty quickly if we're not careful 🌳🌳🌳🌳🌳🌳🌳🌳🌳

package.json
```diff
@@ -343,6 +344,11 @@
                 "category": "PostgreSQL",
                 "command": "cosmosDB.getPostgresCredentials",
                 "title": "Enter Server Credentials"
+            },
+            {
+                "category": "PostgreSQL",
+                "command": "cosmosDB.configurePostgresFirewall",
+                "title": "Configure Firewall"
```
EricJizbaMSFT, line 351: I like how you added this to the command palette, but I would add a confirmation message for this entry point. I think it's a good idea to make sure people didn't mis-click the wrong command/server before we change something security related. Plus it's nice to be able to see your current IP in the message

EricJizbaMSFT: In terms of the `GenericTreeItem` entry point, I'm somewhat indifferent on if that has the confirmation message. I feel it's not entirely necessary, but it might be simpler code-wise if they just both use it. If you do that, maybe change the message in the tree to something simpler and save the extra info (like my current IP) for the confirmation message. Something like this which would match the credentials message:
`Configure firewall to connect to "emjpg"...`

Issue #1390, opened 2020-03-27
Endless cycle of "Enter server credentials" ♾


EricJizbaMSFT:
Repro steps:
1. Expand a new postgres db
1. Enter correct credentials
    <img width="350" alt="Screen Shot 2020-03-27 at 11 04 13 AM" src="https://user-images.githubusercontent.com/11282622/77786253-b9fc4600-701a-11ea-9111-3075a65916db.png">
1. Make an error happen (I turned my wifi off)
    <img width="469" alt="Screen Shot 2020-03-27 at 11 13 34 AM" src="https://user-images.githubusercontent.com/11282622/77787111-37748600-701c-11ea-9213-88fa7c07aaa9.png">
1. Enter correct credentials
    <img width="350" alt="Screen Shot 2020-03-27 at 11 04 13 AM" src="https://user-images.githubusercontent.com/11282622/77786253-b9fc4600-701a-11ea-9111-3075a65916db.png">
1. The cycle continues
    <img width="469" alt="Screen Shot 2020-03-27 at 11 13 34 AM" src="https://user-images.githubusercontent.com/11282622/77787111-37748600-701c-11ea-9213-88fa7c07aaa9.png">

The point is not the error itself. The point is my credentials are correct so you should not prompt me to enter them again. This is a case where you should be checking `parseError(error).errorType` and only telling me to enter credentials for errors we're pretty sure are related to credentials.

EricJizbaMSFT:
Fixed in https://github.com/microsoft/vscode-cosmosdb/pull/1397

PR #1395, opened 2020-03-30
Only prompt for Postgres credentials when necessary

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1390

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -73,16 +73,18 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         } catch (error) {
             const parsedError: IParsedError = parseError(error);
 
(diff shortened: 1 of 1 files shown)


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -73,16 +73,18 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         } catch (error) {
             const parsedError: IParsedError = parseError(error);
 
-            if (parsedError.errorType !== 'UserCancelledError') {
+            if (parsedError.message.match(/password authentication failed for user/) || parsedError.errorType === 'UserCancelledError') {
```
EricJizbaMSFT: Can you check against errorType instead? This won't work for other languages, and is prone to problems if they even slightly change the message

wwlorey:
Closing in favor of https://github.com/microsoft/vscode-cosmosdb/pull/1397

PR #1396, opened 2020-03-30
Remove warning modal before entering Postgres credentials

wwlorey:
This uses the tree item as a prompt instead of the warning modal.

Screen Shot 2020-03-30 at 3 34 13 PM

From https://github.com/microsoft/vscode-cosmosdb/pull/1381#discussion_r399463021

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -21,6 +21,8 @@ interface IPersistedServer {
     username: string;
 }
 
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
I know we say it's usually better to split PRs up, but can you combine this with https://github.com/microsoft/vscode-cosmosdb/pull/1395? It's literally touching the same exact code and would be easier to make sense of the two together

wwlorey:
Closing in favor of https://github.com/microsoft/vscode-cosmosdb/pull/1397

PR #1397, opened 2020-03-31
Update warning and when to prompt for Postgres credentials

wwlorey:
Combining https://github.com/microsoft/vscode-cosmosdb/pull/1395 and https://github.com/microsoft/vscode-cosmosdb/pull/1396

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -20,7 +20,6 @@ interface IPersistedServer {
     id: string;
     username: string;
 }
-class CredentialsNotFoundError extends Error { }
 const invalidCredentialsErrorType: string = '28P01';
 
 export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionContext> {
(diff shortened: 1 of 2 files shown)


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -98,11 +104,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
 
         if (!username || !password) {
             if (warnBeforePrompting) {
-                await ext.ui.showWarningMessage(
-                    localize('mustEnterUsernameAndPassword', 'You must enter the username and password for server "{0}" to continue.', this.parent.label),
-                    { modal: true },
-                    { title: localize('continue', 'Continue') }
-                );
+                throw new PromptForCredentialsError();
```
EricJizbaMSFT: You're not using most of this function for this case. I suggest you use `getCredentialsFromKeytar` directly inside `loadMoreChildrenImpl`.

At that point, `getCredentials` is only called once from `getPostgresCredentials.ts` and it always prompts, thus I feel like the name would make more sense if it were something like `promptForCredentials` or `enterCredentials`

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -73,16 +75,20 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         } catch (error) {
             const parsedError: IParsedError = parseError(error);
 
-            if (parsedError.errorType !== 'UserCancelledError') {
-                // tslint:disable-next-line: no-floating-promises
-                ext.ui.showWarningMessage(localize('couldNotConnect', 'Could not connect to "{0}": {1}', this.parent.label, parsedError.message));
+            if (parsedError.errorType === invalidCredentialsErrorType || parsedError.errorType === 'UserCancelledError' || parsedError.errorType === 'PromptForCredentialsError') {
```
EricJizbaMSFT: No need to check `UserCancelledError` anymore since you're not prompting for credentials

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -73,77 +75,59 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         } catch (error) {
             const parsedError: IParsedError = parseError(error);
 
-            if (parsedError.errorType !== 'UserCancelledError') {
-                // tslint:disable-next-line: no-floating-promises
-                ext.ui.showWarningMessage(localize('couldNotConnect', 'Could not connect to "{0}": {1}', this.parent.label, parsedError.message));
+            if (parsedError.errorType === invalidCredentialsErrorType || parsedError.errorType === 'CredentialsNotFoundError') {
```
EricJizbaMSFT: The `CredentialsNotFoundError` pattern is a bit odd. I think we get better type safety and a more obvious flow if we do something like this:

```
const { username, password } = await this.getCredentialsFromKeytar();
if (username && password) {
    try {
        // return normal tree items
    } catch (error) {
        if (parsedError.errorType === invalidCredentialsErrorType) {
            // show warning message and fall down below
        } else {
            throw error;
        }
    }
}

// return "Enter Credentials" tree item
```

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -73,77 +75,59 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
         if (storedValue && this._keytar) {
             const servers: IPersistedServer[] = JSON.parse(storedValue);
             for (const server of servers) {
                 if (server.id === this._serverId) {
-                    username = server.username;
-                    password = await this._keytar.getPassword(this._serviceName, this._serverId) || undefined;
-                    break;
+                    return {
+                        username: server.username,
+                        password: <string>await this._keytar.getPassword(this._serviceName, this._serverId)
```
EricJizbaMSFT: Related to the above comment and type safety, it's not a great idea to cast this to `string`. For example - maybe someone looked through their keychain and deleted all passwords they didn't recognize - now this will return undefined/null and I feel like that should be treated the same way as if they didn't enter their password

Issue #1355, opened 2020-03-10
Create Postgres Account - Default Setup


neelip:

neelip:
1. Postgres API allows 4 different kinds of account creation --> Default, Restore, GeoRestore and Replica. Should we just support default for now?
2. For the Default create mode, there are few optional properties --> version, storage profile, sku, tags. How do we want to handle them?

neelip:
Should we store the username and password as the user inputs these to create an account? Or we only use these inputs to create an account and then late on they have to re-enter these details to access the databases?

wwlorey:
> Should we store the username and password as the user inputs these to create an account? Or we only use these inputs to create an account and then late on they have to re-enter these details to access the databases?

I think it would make sense to cache the username and password when the account is created so they don't have to be entered multiple times. We should be able to reuse [this function](https://github.com/microsoft/vscode-cosmosdb/blob/fd1bace91597078193ca8b3106f6c147cc76e1a4/src/postgres/tree/PostgresDatabaseTreeItem.ts#L88) with some tweaks. It may need to be moved to `PostgresServerTreeItem` since we won't have access to databases before creating the server

PR #1394, opened 2020-03-27
Enabled delete command for postgres database

neelip:
Fixes #1393

PR #1401, opened 2020-04-01
Move credential related functions to PostgresServerTreeItem

wwlorey:
This should help with https://github.com/microsoft/vscode-cosmosdb/issues/1355

Also it makes more sense to manage server credentials in `PostgresServerTreeItem` rather than the database tree item


src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -73,4 +87,62 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
                 return false;
         }
     }
+
+    public async promptForCredentials(): Promise<{ username: string, password: string }> {
```
EricJizbaMSFT, line 91: Since you're already moving this code around, I've got some suggestions
1. Move the code in this method (promptForCredentials) to `getPostgresCredentials.ts`. That file has very little code and this file does not. The only call into the tree item would be for `persistServer`, which you could make public
1. Rename `persistServer` to `setCredentials`, to match `getCredentialsFromKeytar ` (I'm indifferent on whether or not it has 'keytar' in the name)
1. Rename `getPostgresCredentials.ts` to `enterPostgresCredentials.ts` to match the title of the command in package.json
1. Move `enterPostgresCredentials.ts` into a commands folder (Related to https://github.com/microsoft/vscode-cosmosdb/pull/1394#discussion_r400557852)

PR #1404, opened 2020-04-03
Create Default Postgres Server Account - Separate Create Command

### LINKED ISSUE #1355: Create Postgres Account - Default Setup

neelip:
Fixes #1355 
- First prompt provides options to Postgres or Cosmos; Cosmos experience is not changed
- Default "Create" setup --> prompted for username and password; ssl is set to Enabled by default
- Includes --> server name, username and password validation as described in the azure portal
- Firewall setup added to the create prompts


src/tree/SubscriptionTreeItem.ts
```diff
@@ -60,34 +68,80 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         ];
-        LocationListStep.addStep(wizardContext, promptSteps);
-
-        const wizard = new AzureWizard(wizardContext, {
-            promptSteps,
-            executeSteps: [
-                new CosmosDBAccountCreateStep()
-            ],
-            title: 'Create new Cosmos DB account'
+        const wizardInitial = new AzureWizard(wizardContextInitial, {
```
wwlorey, line 76: It may be overkill to use a separate wizard here since it's only one prompt. 

Is it possible to combine `DatabaseAccountOptionsStep` with the rest of the wizard? That way we would have a back button

src/tree/SubscriptionTreeItem.ts
```diff
@@ -60,34 +68,80 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+            });
+            await wizard.prompt();
+            const serverName: string = nonNullProp(wizardContext, 'serverName');
+            context.showCreatingTreeItem(serverName);
+            await wizard.execute();
+
+            vscode.window.showInformationMessage(`Successfully created account "${wizardContext.serverName}".`);
+            const serverTree: PostgresServerTreeItem = new PostgresServerTreeItem(this, nonNullProp(wizardContext, 'server'));
+            let user: string = nonNullProp(wizardContext, 'adminUser');
+            const usernameSuffix: string = `@${serverName}`;
```
wwlorey, line 134: nit: There's a bit of overlap here with [enterPostgresCredentials.ts](https://github.com/microsoft/vscode-cosmosdb/blob/master/src/postgres/commands/enterPostgresCredentials.ts#L21). It could probably be combined into a function called "validateUsername" or something along those lines

package-lock.json
```diff
@@ -137,9 +137,9 @@
             },
             "dependencies": {
                 "@types/node": {
-                    "version": "13.9.8",
-                    "resolved": "https://registry.npmjs.org/@types/node/-/node-13.9.8.tgz",
-                    "integrity": "sha512-1WgO8hsyHynlx7nhP1kr0OFzsgKz5XDQL+Lfc3b1Q3qIln/n8cKD4m09NJ0+P1Rq7Zgnc7N0+SsMnoD1rEb0kA=="
+                    "version": "13.11.0",
```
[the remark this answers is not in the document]
neelip: I think its because it wasn't recently merged with master

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -7,43 +7,74 @@ import PostgreSQLManagementClient from "azure-arm-postgresql";
...
 import { nonNullProp } from "../../utils/nonNull";
 import { PostgresServerTreeItem } from "../tree/PostgresServerTreeItem";
 
-export async function configurePostgresFirewall(context: IActionContext, treeItem?: PostgresServerTreeItem): Promise<void> {
-    if (!treeItem) {
+export async function configurePostgresFirewall(context: IActionContext | IPostgresWizardContext, treeItem?: PostgresServerTreeItem, createMode?: boolean): Promise<string> {
+    const ip: string = await publicIp.v4();
+    if (!treeItem && createMode) {
+
+        const yesPick: IAzureQuickPickItem<string> = { data: 'Yes', label: 'Yes' };
```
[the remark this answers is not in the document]
EricJizbaMSFT: I really hate yes/no quick picks - they just looks so ugly to me haha. For a more logical reason, I also prefer if the pick/button reflects the action users take rather than forcing them to parse out the question to figure out what "yes" does. How about this instead:

placeHolder: `A firewall rule is required to access this server from your current IP.`
pick1: `Add firewall rule for IP "{0}"`
pick2: `$(clock) Skip for now`

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -7,43 +7,74 @@ import PostgreSQLManagementClient from "azure-arm-postgresql";
...
 import { azureUtils } from "../../utils/azureUtils";
 import { localize } from "../../utils/localize";
 import { nonNullProp } from "../../utils/nonNull";
 import { PostgresServerTreeItem } from "../tree/PostgresServerTreeItem";
 
-export async function configurePostgresFirewall(context: IActionContext, treeItem?: PostgresServerTreeItem): Promise<void> {
-    if (!treeItem) {
+export async function configurePostgresFirewall(context: IActionContext | IPostgresWizardContext, treeItem?: PostgresServerTreeItem, createMode?: boolean): Promise<string> {
+    const ip: string = await publicIp.v4();
+    if (!treeItem && createMode) {
```
EricJizbaMSFT, line 20: All of the stuff inside of this `if` case should be inside `PostgresServerFirewallStep.ts` instead. It's only related to the prompt step, not running the command. There should be other ways to share code if necessary

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -7,43 +7,74 @@ import PostgreSQLManagementClient from "azure-arm-postgresql";
...
+    if (treeItem) {
+        void setFirewallRule(treeItem);
+    }
+
+    return "";
+}
+
+export async function setFirewallRule(treeItem: PostgresServerTreeItem): Promise<void> {
+
+    const ip: string = await publicIp.v4();
```
EricJizbaMSFT, line 57: I believe you already have `ip` and can pass it in to this function instead of calling `publicIp.v4()` again, which may or may not add a delay

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -7,43 +7,74 @@ import PostgreSQLManagementClient from "azure-arm-postgresql";
...
-        startIpAddress: ip,
-        endIpAddress: ip
-    };
-
     const options: vscode.ProgressOptions = {
         location: vscode.ProgressLocation.Notification,
         title: localize('configuringFirewall', 'Adding firewall rule for IP "{0}" to server "{1}"...', ip, serverName)
     };
 
+    const newFirewallRule: FirewallRule = {
```
EricJizbaMSFT, line 68: I don't think moving this does anything? Please try to keep the diff as clean as possible for us reviewers and move this back

src/tree/SubscriptionTreeItem.ts
EricJizbaMSFT: Yeah we need to have a more in-depth discussion about the entry points for creating a Postgres server vs cosmos account. For example -
- Command palette: If I execute a Cosmos DB command from the command palette, it doesn't make sense for you to give me the option to create a postgres server.
- Plus button in title bar: Perhaps "PostgreSQL" could be added to the prompt where users currently pick between Mongo/SQL/Graph

I propose we keep it simple in this PR (there's plenty of other stuff to review) and ignore those two entry points for now. Instead, how about you just add a "Create PostgreSQL server..." command to the right click menu on subscriptions. I believe that would be the simplest option and should be easy to refactor in the future. That means you would not change `createChildImpl` here and would just add a `createPostgresServer.ts` file to the `commands` folder for now

src/tree/PostgresAccountWizard/PostgresServerCredStep.ts
```diff
@@ -0,0 +1,77 @@
+/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT, line 1: Huh I expected these files to be in the "postgres" folder. Can we put them somewhere in there?

src/tree/PostgresAccountWizard/IPostgresWizardContext.ts
```diff
@@ -0,0 +1,19 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { Server } from 'azure-arm-postgresql/lib/models';
+import { IResourceGroupWizardContext } from 'vscode-azureextensionui';
+
+export interface IPostgresWizardContext extends IResourceGroupWizardContext {
+
+    serverName?: string;
+    server?: Server;
+    resourceGroupName?: String;
```
EricJizbaMSFT, line 13: Remove this - it's never used.

src/tree/PostgresAccountWizard/IPostgresWizardContext.ts
```diff
@@ -0,0 +1,19 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Server } from 'azure-arm-postgresql/lib/models';
+import { IResourceGroupWizardContext } from 'vscode-azureextensionui';
+
+export interface IPostgresWizardContext extends IResourceGroupWizardContext {
+
+    serverName?: string;
```
EricJizbaMSFT, line 11: Rename this to `newServerName` to be consistent with our existing patterns and make clear it isn't related to an existing server

src/tree/PostgresAccountWizard/PostgresServerCreateStep.ts
```diff
@@ -0,0 +1,47 @@
...
+import { localize } from '../../utils/localize';
+import { nonNullProp } from '../../utils/nonNull';
+import { IPostgresWizardContext } from './IPostgresWizardContext';
+
+export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresWizardContext> {
+    public priority: number = 100;
+
+    public async execute(wizardContext: IPostgresWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const client: PostgreSQLManagementClient = createAzureClient(wizardContext, PostgreSQLManagementClient);
+        const createMessage: string = localize('creatingPostgresServer', 'Creating Postgres Server "{0}"... It should be ready in several minutes.', wizardContext.serverName);
```
EricJizbaMSFT, line 19: Always use "PostgreSQL" in user-visible strings

src/tree/PostgresAccountWizard/PostgresServerCreateStep.ts
```diff
@@ -0,0 +1,47 @@
...
+            properties: {
+                administratorLogin: user,
+                administratorLoginPassword: password,
+                sslEnforcement: "Enabled",
+                createMode: "Default"
+            },
+        };
+
+        wizardContext.server = await client.servers.create(rgName, serverName, options);
+        wizardContext.server = await client.servers.get(rgName, serverName);
```
EricJizbaMSFT, line 40: Do you have to do this separate `get` or does `create` already return what you need? I think you copied behavior from cosmos db, but we only do that because there seems to be a bug in that case

src/tree/SubscriptionTreeItem.ts
```diff
@@ -60,34 +68,80 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+                    new PostgresServerCreateStep()
+                ],
+                title: 'Create new Postgres account'
+            });
+            await wizard.prompt();
+            const serverName: string = nonNullProp(wizardContext, 'serverName');
+            context.showCreatingTreeItem(serverName);
+            await wizard.execute();
+
+            vscode.window.showInformationMessage(`Successfully created account "${wizardContext.serverName}".`);
```
EricJizbaMSFT, line 131: Always use "server" for Postgres - "account" is a cosmos term

src/tree/PostgresAccountWizard/PostgresServerCredStep.ts
```diff
@@ -0,0 +1,77 @@
...
+export class PostgresServerCredStep extends AzureWizardPromptStep<IPostgresWizardContext> {
+
+    public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
+        wizardContext.adminUser = (await ext.ui.showInputBox({
+            placeHolder: "Username",
+            prompt: 'Enter administrator username for the server.',
+            validateInput: validateUser,
+        })).trim();
+        const user = nonNullProp(wizardContext, 'adminUser');
+        wizardContext.adminPassword = (await ext.ui.showInputBox({
```
EricJizbaMSFT, line 20: Each call to `ext.ui.showInputBox` has to be a separate prompt step otherwise the back button won't work properly

src/tree/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+async function validatePostgresAccountName(name: string, client: PostgreSQLManagementClient): Promise<string | undefined> {
+    name = name ? name.trim() : '';
+
+    const min = 3;
+    const max = 63;
+
+    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
+    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
```
EricJizbaMSFT, line 41: You should check the length first  and return before ever calling `checkNameAvailability` if it's not valid. We want to avoid the delay related to `checkNameAvailability` when possible

src/postgres/commands/configurePostgresFirewall.ts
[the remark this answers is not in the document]
neelip: What do you mean by "$(clock)"?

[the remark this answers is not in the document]
EricJizbaMSFT: It's an octicon: https://octicons.github.com/

src/tree/PostgresAccountWizard/IPostgresWizardContext.ts
neelip: What is the purpose of this? Given that the usage of this interface is for creation purposes, do we need to specify that it is a new server?

src/tree/PostgresAccountWizard/PostgresServerCredStep.ts
neelip: Do you mean the whole PostgresAccountWizard folder?

EricJizbaMSFT: Yes

src/tree/PostgresAccountWizard/IPostgresWizardContext.ts
EricJizbaMSFT: Yes it is best to follow existing naming patterns even if the situation doesn't apply to this specific wizard (yet). There are many wizard steps (i.e. resource group) where the user can create a new one or select an existing one - which is why we follow this naming pattern

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -7,43 +7,74 @@ import PostgreSQLManagementClient from "azure-arm-postgresql";
...
+        const yesPick: IAzureQuickPickItem<string> = { data: 'Yes', label: 'Yes' };
+        const noPick: IAzureQuickPickItem<string> = { data: 'No', label: 'No' };
+        const picks: IAzureQuickPickItem<string>[] = [yesPick, noPick];
+
+        const wizardContext: IPostgresWizardContext = <IPostgresWizardContext>context;
+
+        const result = await ext.ui.showQuickPick(picks, { placeHolder: `Would you like to add firewall rule for your IP ${ip} to this server?` });
+
+        if (result.label === "Yes") {
+            wizardContext.firewall = true;
```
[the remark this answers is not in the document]
neelip: The current commit should have all of the firewall code for create in the PostgresServerFirewallStep.ts. 
The only difference between the original configure... file and this version is that the prompt part and setting part are split between two functions. 

The setting of the firewall rule requires PostgresServerTreeItem and in the case of create account this is not available until after the CreateStep. So the PostgresServerFirewallStep is only to prompt and the given response is stored in the firewall context and retrieved after the CreateStep to setup the firewall rule based on the response. 

I agree that the firewall context seems unnecessary, do you have any suggestions for working around it?

src/tree/SubscriptionTreeItem.ts
neelip: The database login part doesn't seem to need extensive username validation compared to the create step. 
Do you mean combining just to simplify the username section in SubscriptionTreeItem or combining enterPostgresCredentials.ts with PostgresServerCredStep...?

wwlorey: I was talking about [these lines](https://github.com/microsoft/vscode-cosmosdb/blob/6c8f5d8e3a670efffb207f8cf422950301d5cd31/src/tree/SubscriptionTreeItem.ts#L137-L140) being duplicated but its probs fine as it is.

src/postgres/commands/configurePostgresFirewall.ts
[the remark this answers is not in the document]
wwlorey: Ah I see what you're saying - it does make sense to have `firewall` in the context so you know if it needs to be configured once the tree item is created

src/tree/SubscriptionTreeItem.ts
neelip: Oh ok, for these lines I feel like we might have to change the enterPostgresCredentials code, I feel like that it's unnecessary given that the rest of the function is not useful in this scenario.

src/postgres/commands/PostgresAccountWizard/PostgresServerFirewallStep.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
+
+        const ip: string = await publicIp.v4();
+        const yes: QuickPickItem = { label: localize('addFirewallRule', 'Add firewall rule for IP "{0}"', ip) };
+        const no: QuickPickItem = { label: localize('skipFireWallRule', '$(clock) Skip for now') };
+
+        const placeHolder: string = localize('addFirewallForNewServer', 'A firewall rule is required to access this server from your current IP.');
+
+        wizardContext.firewall = await ext.ui.showQuickPick([yes, no], { placeHolder }) === yes;
```
EricJizbaMSFT, line 23: The name `firewall` by itself doesn't imply to me that this variable is a boolean. I would suggest `includeFirewall` or `addFirewall`

src/postgres/commands/PostgresAccountWizard/PostgresServerFirewallStep.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    }
+
+    public shouldPrompt(wizardContext: IPostgresWizardContext): boolean {
+        return !wizardContext.firewall;
```
EricJizbaMSFT, line 28: This should be `wizardContext.firewall === undefined`

It's a very slight difference, but say someone sets `firewall` to `false` before they start the wizard. That means they don't want a firewall _and_ don't want to prompt here

src/tree/SubscriptionTreeItem.ts
EricJizbaMSFT: This comment doesn't look resolved yet. It's the most complicated comment IMO, so happy to discuss over a Teams call if you want

neelip: It's already fixed, I just forgot to click resolve here.

neelip: It is in the recent commit, I forgot to resolve it here.

src/postgres/commands/PostgresAccountWizard/PostgresServerCredStepPW.ts
```diff
@@ -0,0 +1,53 @@
...
+            validateInput: (password: string) => validatePassword(user, password),
+        })).trim();
+    }
+
+    public shouldPrompt(wizardContext: IPostgresWizardContext): boolean {
+        return !wizardContext.adminPassword;
+    }
+}
+
+async function validatePassword(username: string, password: string): Promise<string | undefined> {
```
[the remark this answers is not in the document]
neelip: I couldn't find any document for this. I set it up based on how it appears  on azure portal.

src/postgres/commands/PostgresAccountWizard/PostgresServerCredStepPW.ts
```diff
@@ -0,0 +1,53 @@
...
+    let numOccurence = 0;
+
+    regex.map(substring => {
+        if (password.match(substring)) {
+            numOccurence++;
+        }
+    });
+
+    if (password.length < min || password.length > max) {
+        return `Password must be between ${min} and ${max} characters.`;
```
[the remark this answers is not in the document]
neelip: Sorry about that, I tried to do a search to find the strings but looks like it didn't fix everything.

src/postgres/commands/PostgresAccountWizard/PostgresServerCredStepPW.ts
```diff
@@ -0,0 +1,53 @@
...
+export class PostgresServerCredStepPW extends AzureWizardPromptStep<IPostgresWizardContext> {
+
+    public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
+        const user = nonNullProp(wizardContext, 'adminUser');
+        wizardContext.adminPassword = (await ext.ui.showInputBox({
+            placeHolder: "Password",
+            prompt: 'Enter administrator password for the server.',
+            password: true,
+            validateInput: (password: string) => validatePassword(user, password),
+        })).trim();
```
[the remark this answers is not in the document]
neelip: You're right, trailing spaces are actually allowed for postgres.

src/postgres/commands/PostgresAccountWizard/PostgresServerCredStepUser.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+async function validateUser(username: string): Promise<string | undefined> {
+    username = username ? username.trim() : '';
+
+    const min = 1;
+    const max = 63;
+
+    const restricted = ['azure_superuser', 'azure_pg_admin', 'admin', 'administrator', 'root', 'guest', 'public'];
```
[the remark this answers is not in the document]
neelip: Same as for password checks, I replicated what is present directly on the portal when creating account. I will try to see if I can find any documentation on it.

src/postgres/commands/PostgresAccountWizard/PostgresServerFirewallStep.ts
```diff
@@ -0,0 +1,30 @@
...
+import { AzureWizardPromptStep } from 'vscode-azureextensionui';
+import { ext } from '../../../extensionVariables';
+import { localize } from '../../../utils/localize';
+import { IPostgresWizardContext } from './IPostgresWizardContext';
+
+export class PostgresServerFirewallStep extends AzureWizardPromptStep<IPostgresWizardContext> {
+
+    public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
+
+        const ip: string = await publicIp.v4();
```
EricJizbaMSFT, line 17: If you await this here and it takes a while, the user won't see anything while its executing. Instead, we want to pass the _promise_ to `showQuickPick` and VS Code will show a little loading animation while the promise executes. Here's how I usually do it:

Define a separate method to get the picks:
```
private async getPicks(): Promise<IAzureQuickPickItem<boolean>[]> {
    const ip: string = await publicIp.v4();
    return [
        { label: localize('addFirewallRule', 'Add firewall rule for IP "{0}"', ip), data: true },
        { label: localize('skipFireWallRule', '$(clock) Skip for now'), data: false }
    ];
}
```

When calling `showQuickPick`, call `getPicks` _without_ `await`
```
wizardContext.addFirewall = (await ext.ui.showQuickPick(this.getPicks(), { placeHolder })).data;
```
NOTE: I also refactored to use `data` instead of checking `=== yes` since we don't know the value of `yes` until after the promise finishes

src/postgres/commands/createPostgresServer.ts
```diff
@@ -0,0 +1,63 @@
...
+    let user: string = nonNullProp(wizardContext, 'adminUser');
+    const usernameSuffix: string = `@${serverName}`;
+    if (!user.includes(usernameSuffix)) {
+        user += usernameSuffix;
+    }
+    const password: string = nonNullProp(wizardContext, 'adminPassword');
+    void serverTree.setCredentials(user, password);
+    if (wizardContext.addFirewall) {
+        const ip: string = await publicIp.v4();
+        void setFirewallRule(serverTree, ip);
```
EricJizbaMSFT, line 60: Did you add `void` because you didn't want to await this? If so - why didn't you want to await this?

Also - we usually disable the lint rule instead of using `void` - it's much easier to tell at a glance what's going on:
```
// tslint:disable-next-line: no-floating-promises
```

src/postgres/commands/createPostgresServer.ts
```diff
@@ -0,0 +1,63 @@
...
+    const serverTree: PostgresServerTreeItem = new PostgresServerTreeItem(node, nonNullProp(wizardContext, 'server'));
+    let user: string = nonNullProp(wizardContext, 'adminUser');
+    const usernameSuffix: string = `@${serverName}`;
+    if (!user.includes(usernameSuffix)) {
+        user += usernameSuffix;
+    }
+    const password: string = nonNullProp(wizardContext, 'adminPassword');
+    void serverTree.setCredentials(user, password);
+    if (wizardContext.addFirewall) {
+        const ip: string = await publicIp.v4();
```
EricJizbaMSFT, line 59: Add `ip` to the wizard context and set it in `PostgresServerFirewallStep` so that you don't have to `await` this call again

src/postgres/commands/createPostgresServer.ts
```diff
@@ -0,0 +1,63 @@
...
+        ],
+        title: localize('createPostgresServerPrompt', 'Create new PostgreSQL server')
+    });
+    await wizard.prompt();
+    const serverName: string = nonNullProp(wizardContext, 'newServerName');
+    await wizard.execute();
+
+    vscode.window.showInformationMessage(localize('createdServerMsg', 'Successfully created server "{0}".', wizardContext.newServerName));
+    const serverTree: PostgresServerTreeItem = new PostgresServerTreeItem(node, nonNullProp(wizardContext, 'server'));
+    let user: string = nonNullProp(wizardContext, 'adminUser');
```
EricJizbaMSFT, line 51: All of this credential and firewall stuff should be a part of the wizard. The benefits include consistent progress notifications, a little extra telemetry related to steps, etc.

You could add it directly to `PostgresServerCreateStep`, but I would lean towards a separate execute step for each since we might convert `enterPostgresCredentials` and `configurePostgresFirewall` to wizards eventually (but probably not this PR for the sake of simplicity)

src/tree/DatabaseAccountOptionsStep.ts
```diff
@@ -0,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { IDBAWizardContext } from './IDBAWizardContext';
+
+export class DatabaseAccountOptionsStep extends AzureWizardPromptStep<IDBAWizardContext> {
```
EricJizbaMSFT, line 10: Remove this file and `IDBAWizardContext` in this PR since we're handling entry points separately

src/postgres/commands/createPostgresServer.ts
neelip: The setFirewallRule can only be called after we have the new Server Tree item created, so can't really set ip in PostgresServerFirewallStep. PostgresServer..Step is only prompting and setting the addFirewall boolean accordingly. But can add ip to setFirewallRule() method, I am not sure if that would help though

EricJizbaMSFT: I'm saying do something like this in `PostgresServerFirewallStep`:
```
wizardContext.publicIp = await publicIp.v4();
```

And whenever you call `setFirewallRule` (_not_ in `PostgresServerFirewallStep`), you would just use `wizardContext.publicIp` instead of calling `publicIp.v4` again

neelip: ok so directly add it to wizardContext, makes sense... in this case we can remove it from configurePostgresFirewall as well

src/postgres/commands/createPostgresServer.ts
neelip: Actually I think we need await at serverTree.setCredentials and setFirewallRule, given they depend on wizard execution.

src/postgres/commands/PostgresAccountWizard/IPostgresWizardContext.ts
```diff
@@ -0,0 +1,27 @@
...
+// tslint:disable-next-line: no-implicit-dependencies
+import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
+import { IResourceGroupWizardContext } from 'vscode-azureextensionui';
+import { SubscriptionTreeItem } from '../../../tree/SubscriptionTreeItem';
+
+export interface IPostgresWizardContext extends IResourceGroupWizardContext {
+
+    newServerName?: string;
+    server?: Server;
+    resourceGroup?: ResourceGroup;
```
EricJizbaMSFT, line 16: nit: remove resourceGroup since it's already on `IResourceGroupWizardContext`

src/postgres/commands/PostgresAccountWizard/PostgresServerSetFirewallStep.ts
```diff
@@ -0,0 +1,44 @@
...
+        const addFirewallMessage: string = localize('configuringFirewall', 'Adding firewall rule for IP "{0}" to server "{1}"...', ip, serverName);
+        progress.report({ message: addFirewallMessage });
+
+        wizardContext.addedFirewall = true;
+
+        await client.firewallRules.createOrUpdate(resourceGroup, serverName, firewallRuleName, newFirewallRule);
+
+    }
+
+    public shouldExecute(wizardContext: IPostgresWizardContext): boolean {
```
EricJizbaMSFT, line 41: This should be:
```
return !!wizardContext.addFirewall;
```

Otherwise I think the firewall will always be added even if I select "Skip for now". You can remove `addedFirewall` as a property from the wizard context

src/postgres/commands/PostgresAccountWizard/PostgresServerSetCredentialsStep.ts
```diff
@@ -0,0 +1,38 @@
...
+
+        const password: string = nonNullProp(wizardContext, 'adminPassword');
+        const server: Server = nonNullProp(wizardContext, 'server');
+        wizardContext.addedCredentials = true;
+
+        await setPostgresCredentials(user, password, nonNullProp(server, 'id'));
+    }
+
+    public shouldExecute(wizardContext: IPostgresWizardContext): boolean {
+        return wizardContext.addedCredentials === undefined;
```
EricJizbaMSFT, line 36: Unlike the firewall step, there's no prompt where people can skip this so you can just do `return true;`. It's not really worth having a separate `addedCredentials` property just for `shouldExecute`

src/postgres/commands/PostgresAccountWizard/PostgresServerSetFirewallStep.ts
```diff
@@ -0,0 +1,44 @@
...
+import { IPostgresWizardContext } from './IPostgresWizardContext';
+
+export class PostgresServerSetFirewallStep extends AzureWizardExecuteStep<IPostgresWizardContext> {
+    public priority: number = 250;
+
+    public async execute(wizardContext: IPostgresWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+
+        const ip: string = nonNullProp(wizardContext, 'publicIp');
+        const node: SubscriptionTreeItem = nonNullProp(wizardContext, 'subscriptonTreeItem');
+        const client: PostgreSQLManagementClient = createAzureClient(node.root, PostgreSQLManagementClient);
```
EricJizbaMSFT, line 22: You can use `wizardContext` instead of `node.root` here since it already has the subscription info necessary to create an azure client. Then you can remove `subscriptonTreeItem` from the wizard context

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -39,7 +39,7 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     constructor(parent: AzureParentTreeItem, server: Server) {
         super(parent);
         this.server = server;
-        this._keytar = tryGetKeyTar();
+        this._keytar = ext.keytar;
```
EricJizbaMSFT, line 42: Since keytar is on `ext` now, we don't need to store it in this class and we can just use `ext.keytar` directly

src/postgres/commands/setPostgresCredentials.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ext } from "../../extensionVariables";
+
+interface IPersistedServer {
+    id: string;
+    username: string;
+}
+
+export async function setPostgresCredentials(username: string, password: string, serverId: string): Promise<void> {
+    if (ext.keytar) {
+        const serviceName: string = "ms-azuretools.vscode-cosmosdb.postgresPasswords";
```
EricJizbaMSFT, line 15: I see you copied this code from `PostgresServerTreeItem`, but I think you forgot to delete the original code from that file

package-lock.json
```diff
@@ -4,6 +4,24 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
+        "@azure/cosmos": {
```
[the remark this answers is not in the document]
neelip: I am not sure why this file had changes this time. I have been following the "rm -rf node_modules/", "npm cache clean --force" and "npm i" steps to make sure this file doesn't have unnecessary changes

src/postgres/commands/setPostgresCredentials.ts
neelip: I haven't completed all the changes yet. I was hoping to get feedback on these new execute steps that is if I am going in the right direction from what we discussed, before I make rest of the changes. I will push rest of the changes then

package-lock.json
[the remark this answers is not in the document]
EricJizbaMSFT: You only need to do that if you see it changing "https" to "http". It's related to a bug in older versions of npm, so it really shouldn't happen going forward since I believe your npm is up to date now

src/postgres/commands/setPostgresCredentials.ts
EricJizbaMSFT: Yep the execute steps look good overall

neelip: I was planning to leave the setFirewallRule function as is in configurePostgresFirewall.ts even though duplicate code exists in the execute wizard ...SetFirewallStep. What do you think, should I remove that and make changes to configurePostgresFirewall.ts as well?

EricJizbaMSFT: Yes please remove the duplicate code. I believe minimal changes should be required to re-use this function

neelip: Do you mean reverting only configurePostgresFirewall.ts or reverting configurePostgresFirewall.ts and removing PostgresServerSetFirewallStep.ts as well?

EricJizbaMSFT: I'm only talking about changes you made in [this last commit](https://github.com/microsoft/vscode-cosmosdb/pull/1404/commits/d27d1606cf37be50ef5d445a933e2c3c22e7acb9)

So you should definitely keep `PostgresServerSetFirewallStep.ts`, but as an example you should change this:
```
const resourceGroup: string = nonNullProp(nonNullProp(wizardContext, 'resourceGroup'), 'name');
```

back to this:

```
const resourceGroup: string = nonNullProp(wizardContext, 'newResourceGroupName');
```

since that change was related to [this comment](https://github.com/microsoft/vscode-cosmosdb/pull/1404#discussion_r414127094)

PR #1406, opened 2020-04-05
 Change the minimum database throughput value from 400

ymasaoka:
Changed minThroughputPartitioned so that RU can be specified from 400 even in non-Fixed state.


EricJizbaMSFT:
This looks good to me - I wanted to do a quick test before approving/merging, but I can't create a Cosmos DB account right now due to region limitations

ymasaoka:
I see. Now that Azure is under resource restrictions due to the effects of coronavirus, I want you to try again after a certain period of time.

Issue #1407, opened 2020-04-06
Support Postgres Functions


wwlorey:
Stored procedures and functions were both created using the `CREATE FUNCTION` syntax until Postgres 11. The newest versions added support for `CREATE PROCEDURE` which also allows you to run transactions.

I think it would be best to add support for functions and stored procedures separately and only show procedures for v11. This would be consistent with how GUIs like pgAdmin handle this:

<img width="257" alt="Screen Shot 2020-04-06 at 1 11 20 PM" src="https://user-images.githubusercontent.com/22795803/78601526-c5fdba00-7809-11ea-9a83-42a05d7f0269.png"><img width="264" alt="Screen Shot 2020-04-06 at 1 13 55 PM" src="https://user-images.githubusercontent.com/22795803/78601533-c8601400-7809-11ea-9995-a73172671e6e.png">

@fiveisprime, @EricJizbaMSFT, @nturinski what do you think of this approach?

fiveisprime:
Agreed. So this would mean that v10 databases won't show "Procedures," right?

wwlorey:
Right, only v11 and up would show procedures

wwlorey:
Splitting stored procs into a separate issue: https://github.com/microsoft/vscode-cosmosdb/issues/1471

PR #1408, opened 2020-04-07
Add "Tables" tree item for Postgres

wwlorey:
Adding this so we can support functions & stored procs too (related: https://github.com/microsoft/vscode-cosmosdb/issues/1407)

I had trouble finding good icons for the tables group and for individual tables so I left them generic for now.

Before ➡️After
Screen Shot 2020-04-07 at 12 47 41 PMScreen Shot 2020-04-07 at 12 42 50 PM

### DIFF
--- a/src/postgres/tree/PostgresSchemaTreeItem.ts
+++ b/src/postgres/tree/PostgresSchemaTreeItem.ts
@@ -13,6 +13,7 @@ export class PostgresSchemaTreeItem extends AzureParentTreeItem<ISubscriptionCon
     public static contextValue: string = "postgresSchema";
     public readonly contextValue: string = PostgresSchemaTreeItem.contextValue;
     public readonly schema: Schema;
+    public autoSelectInTreeItemPicker: boolean = true;
 
     private _tablesTreeItem: PostgresTablesTreeItem;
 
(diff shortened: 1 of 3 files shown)


src/postgres/tree/PostgresSchemaTreeItem.ts
```diff
@@ -3,22 +3,23 @@
...
 import { AzureParentTreeItem, ISubscriptionContext } from 'vscode-azureextensionui';
 import { getThemeAgnosticIconPath } from '../../constants';
-import { PostgresTableTreeItem } from './PostgresTableTreeItem';
+import { PostgresTablesTreeItem } from './PostgresTablesTreeItem';
 
 export class PostgresSchemaTreeItem extends AzureParentTreeItem<ISubscriptionContext> {
     public static contextValue: string = "postgresSchema";
     public readonly contextValue: string = PostgresSchemaTreeItem.contextValue;
-    public readonly childTypeLabel: string = "Table";
     public readonly schema: Schema;
```
EricJizbaMSFT, line 15: Probably want to set `autoSelectInTreeItemPicker` just like [here](https://github.com/microsoft/vscode-azurestorage/blob/3ea68d7f2f868555c48b04d0b5395544ed4b1668/src/tree/StorageAccountTreeItem.ts#L49)

We don't want the "Tables" tree item to show up in the command palette

PR #1411, opened 2020-04-09
Add tree view for Postgres functions

wwlorey:


wwlorey:
I'll push tweaks to this once https://github.com/microsoft/vscode-cosmosdb/pull/1412 is merged

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -94,10 +99,52 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         credentialsTreeItem.commandArgs = [this.parent];
         return [credentialsTreeItem];
     }
+
     public async deleteTreeItemImpl(): Promise<void> {
         const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
         await client.databases.deleteMethod(azureUtils.getResourceGroupFromId(this.fullId), this.parent.name, this.databaseName);
     }
+
+    private async listFunctions(clientConfig: ClientConfig): Promise<IPostgresFunction[]> {
```
EricJizbaMSFT, line 108: This should happen inside `PostgresFunctionsTreeItem.loadMoreChildrenImpl`. We want to lazy load things in the tree as much as possible, which means you don't load the functions unless they expand the "Functions" tree item

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -133,4 +133,9 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
             await this._keytar.setPassword(this._serviceName, this._serverId, password);
         }
     }
+
+    public supportsStoredProcedures(): boolean {
+        const version: string | undefined = this.server.version;
+        return !!version && parseFloat(version) >= 11;
```
EricJizbaMSFT, line 139: Does postgres use semver? If so, I suggest you use the [semver package](https://www.npmjs.com/package/semver) instead of `parseFloat`. Theortically `parseFloat` would fail for versions like "11.0.0-alpha"

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -94,10 +99,52 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            }
+
+            allNames.add(row.name);
+        }
+
+        return rows.map(row => {
+            return {
+                name: row.name,
+                oid: row.oid,
+                description: duplicateNames.has(row.name) ? row.schema : '',
```
EricJizbaMSFT, line 143: This method is about getting the "raw" functions themselves, not converting them into tree items. Since `description` is a treeItem-specific property, I would prefer if you passed along `isDuplicate` and `schema` on this object and defined `description` in the tree item itself. (I feel like we might need `schema` in the future anyways)

Something like
```
public get description(): string | undefined {
    return this.func.isDuplicate ? this.func.schema : undefined;
}
```

src/postgres/tree/PostgresFunctionTreeItem.ts
```diff
@@ -0,0 +1,39 @@
...
+export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext> {
+    public static contextValue: string = 'postgresFunction';
+    public readonly contextValue: string = PostgresFunctionTreeItem.contextValue;
+    public readonly schema: string;
+    public readonly name: string;
+    public readonly id: string;
+    public readonly isDuplicate: boolean;
+    public definition: string;
+
+    constructor(parent: PostgresFunctionsTreeItem, schema: string, name: string, oid: number, definition: string, isDuplicate: boolean) {
```
EricJizbaMSFT, line 19: I would make an interface for this type: `{ schema: string, name: string, oid: number, definition: string }` and pass the row object into the constructor instead of passing each property individually. I try to avoid having so many parameters with the same type (`string`) because it can be easy to get them out of order and cause a bug

PR #1412, opened 2020-04-09
Remove Schema tree

### LINKED ISSUE #1374: Remove Schema Node

neelip:
Fixes #1374


wwlorey:
Since you can create tables with the same name under different schemas, it's possible to have multiple tables with the same name listed without a way to differentiate between the two:

Screen Shot 2020-04-10 at 10 14 23 AM

Since the table ID is the OID, tables with the same name don't cause an error. But it may be helpful for the table label to be formatted as `schema.table_name`

EricJizbaMSFT:
I agree we need to handle the case Will described, but instead of `schema.table_name`, I would prefer setting `label` to `table_name` and `description` to `schema`. That being said, I think adding `schema` as the `description` property will look ugly in the majority of cases, so I would suggest you only do it if you detect duplicate table names in the tree. So it would look like this:


- Tables
  - table1
  - dupeTable (schema1)
  - dupeTable (schema2)
  - table2

But as Will already approved - you can totally do that as a separate PR

neelip:
That's a good idea, we can avoid extra labels unnecessarily being too long

PR #1414, opened 2020-04-09
Create Postgres Database

### LINKED ISSUE #1377: Create Postgres Database

neelip:
Fixes #1377


src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -87,6 +88,22 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
         }
     }
 
+    public async createChildImpl(context: ICreateChildImplContext): Promise<PostgresDatabaseTreeItem> {
+        const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
+        const fullID: string = nonNullProp(this, 'fullId');
+        const resourceGroup: string = azureUtils.getResourceGroupFromId(fullID);
```
EricJizbaMSFT, line 94: You call `getResourceGroupFromId` quite a few times now. How about you add a `resourceGroup` property to this tree item class and set it once in the constructor

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -87,6 +88,22 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+        const databaseName = await ext.ui.showInputBox({
+            placeHolder: "Database Name",
+            prompt: "Enter the name of the database",
+            validateInput: (name: string) => validateDatabaseName(name, client, this.name, resourceGroup)
+        });
+        context.showCreatingTreeItem(databaseName);
+        const database: Database = { name: databaseName };
+        await client.databases.createOrUpdate(resourceGroup, this.name, databaseName, database);
+        await this.refresh();
```
EricJizbaMSFT, line 103: This line will cause us to re-load all databases (which could be slow) and should be unnecessary

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -134,3 +151,15 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
         }
     }
 }
+
+async function validateDatabaseName(name: string, client: PostgreSQLManagementClient, serverName: string, resourceGroup: string): Promise<string | undefined | null> {
+    const currentDatabaseList: DatabaseListResult = await client.databases.listByServer(resourceGroup, serverName);
```
EricJizbaMSFT, line 156: Similar comment to other PR. You want to do all sync calls first before you ever attempt an async call like this which could take longer. This method is executed every time the user types a character.

In fact, it would be best if we could find a way not to re-fetch the list of databases at all in this validate function. For example, if they've already expanded the tree, we should already know which databases there are.

neelip: So, maybe caching the names of the current databases when we call loadMoreChildren...? Or creating a property for the list of databases, and then use it for loadMoreChildren and createChild/validate function?

EricJizbaMSFT: There's already a caching mechanism for tree items. Just use `getCachedChildren`. You can create the promise before `ext.ui.showInputBox({`:
```
const getChildrenTask: Promise<AzExtTreeItem[]> = this.getCachedChildren(context);
```

pass the task in to `validateDatabaseName`, but only `await` it after you've validated the other quick stuff like length

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -145,3 +159,16 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
     }
 }
+
+async function validateDatabaseName(name: string, getChildrenTask: Promise<AzExtTreeItem[]>): Promise<string | undefined | null> {
+    if (!name || name.length < 1) {
+        return localize('NameCannotBeEmpty', 'Name cannot be empty.');
+    }
+    const currDatabaseList = await getChildrenTask;
+    const currDatabaseNames: string[] = [];
+    currDatabaseList.map(database => currDatabaseNames.push(database.label));
```
EricJizbaMSFT, line 169: As a general rule, `label` should only be used for display purposes. We've had bugs in the past because `label` isn't always the same as name. In this case, I would suggest something like this:
```
for (const db of currDatabaseList) {
    if (db instanceof PostgresDatabaseTreeItem) {
        currDatabaseNames.push(db.databaseName);
    }
}
```
I'm using `databaseName` instead of `label` and the `instanceof` check helps in the rare case where the children includes an unexpected tree item. For example, `createTreeItemsWithErrorHandling` can include an `InvalidTreeItem` if one child has an error, but the rest are fine

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -145,3 +159,16 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+    if (currDatabaseNames.includes(name)) {
```
EricJizbaMSFT, line 170: Are database names case sensitive? If not, you should do `toLowerCase` on both `currDatabaseNames` and `name` to prevent the same name with different casing

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -145,3 +159,16 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
         return !!version && gte(version, '11.0.0');
     }
 }
+
+async function validateDatabaseName(name: string, getChildrenTask: Promise<AzExtTreeItem[]>): Promise<string | undefined | null> {
+    if (!name || name.length < 1) {
```
EricJizbaMSFT, line 164: nit: `!name` already handles the empty string case. You don't have to check the string length

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -33,12 +34,14 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     private readonly _serviceName: string = "ms-azuretools.vscode-cosmosdb.postgresPasswords";
     private _keytar: KeyTar | undefined;
     private _serverId: string;
+    private _resourceGroup: string;
```
EricJizbaMSFT, line 37: Just make this public - I'm sure we'll need the resourceGroup for commands outside this file as well

src/postgres/tree/PostgresServerTreeItem.ts
neelip: They are case sensitive. Actually, except for "Name cannot be empty", there aren't any other restrictions to database names. Only pgAdmin has one extra check to see if a database name already exists under a server. We can still create duplicate database names from the extension if we allow. Should we follow pgAdmin and restrict duplicates?

PR #1415, opened 2020-04-13
Delete SQL Database

v-wuzhai:


test/nightly/azureResourceSQL.test.ts
```diff
@@ -74,6 +71,15 @@ suite('SQL action', async function (this: Mocha.Suite): Promise<void> {
         assert.ifError(await getDocDBCollectionMeta(accountName, databaseName, collectionId2));
     });
 
+    test('Delete SQL Database', async () => {
+        assert.ok(await getDatabaseMeta());
+        const testInputs: (string | RegExp)[] = [testAccount.getSubscriptionContext().subscriptionDisplayName, `${accountName} (SQL)`, databaseName, DialogResponses.deleteResponse.title];
+        await testUserInput.runWithInputs(testInputs, async () => {
+            await vscode.commands.executeCommand('cosmosDB.deleteDocDBDatabase');
+        });
+        assert.ifError(await getDatabaseMeta());
```
nturinski, line 80: I think this needs to be changed according to this:
https://nodejs.org/api/assert.html#assert_assert_iferror_value

Since `getDatabaseMeta` will return string | undefined, the behavior won't change whether the delete succeeded or not.

PR #1417, opened 2020-04-13
Add support for editing & updating Postgres functions

wwlorey:
Screen Shot 2020-04-13 at 3 19 11 PM

The filename includes function name, server name, and schema name to uniquely identify functions. Otherwise, you couldn't open two functions named "test" from two different servers/schemas at the same time.

### DIFF
--- a/src/postgres/tree/PostgresFunctionTreeItem.ts
+++ b/src/postgres/tree/PostgresFunctionTreeItem.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Client } from "pg";
 import { AzureTreeItem, ISubscriptionContext, TreeItemIconPath } from "vscode-azureextensionui";
 import { getThemeAgnosticIconPath } from "../../constants";
 import { IPostgresFunctionsQueryRow, PostgresFunctionsTreeItem } from "./PostgresFunctionsTreeItem";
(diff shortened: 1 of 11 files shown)


src/postgres/tree/PostgresFunctionTreeItem.ts
```diff
@@ -36,4 +39,12 @@ export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext
     public get iconPath(): TreeItemIconPath {
         return getThemeAgnosticIconPath('Collection.svg');
     }
+
+    public async update(newDefinition: string): Promise<string> {
+        this.definition = newDefinition;
+        const client = new Client(this.parent.clientConfig);
+        await client.connect();
+        await client.query(this.definition);
```
EricJizbaMSFT, line 47: query updates the definition? Seems odd to me

src/postgres/tree/PostgresFunctionTreeItem.ts
```diff
@@ -36,4 +39,12 @@ export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext
     public get iconPath(): TreeItemIconPath {
         return getThemeAgnosticIconPath('Collection.svg');
     }
+
+    public async update(newDefinition: string): Promise<string> {
```
EricJizbaMSFT, line 43: Would prefer to have this code directly in `PostgresFunctionEditor.ts`. A whole lot of code is _related_ to tree items, but shouldn't be put in the class itself because it would get out of hand quickly

src/postgres/tree/PostgresFunctionTreeItem.ts
wwlorey: So `definition` is technically the query used to create/replace a function. So executing the definition query will update it on the server.

Would it be helpful to change the `definition` var name?

EricJizbaMSFT: 🤷‍♂️

PR #1420, opened 2020-04-15
Add support for creating Postgres functions

wwlorey:


src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -47,29 +105,39 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+
+        const queryResult: QueryResult = await client.query(schemaQuery);
+        return queryResult.rows ? queryResult.rows.map(row => row.schema) : [];
     }
 
-    public isAncestorOfImpl(contextValue: string): boolean {
-        return contextValue === PostgresFunctionTreeItem.contextValue;
+    private validateFunctionName(name: string, existingFunctions: Set<string>): string | undefined {
+        if (existingFunctions.has(name)) {
+            return localize('functionAlreadyExists', 'A function named "{0}" already exists.', name);
```
nturinski, line 129: I would write a function named "{0}" with schema "{1}" already exists since it might be confusing as to why you can have duplicates sometimes but not others.

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -144,4 +145,23 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
         const version: SemVer | null = coerce(this.server.version);
         return !!version && gte(version, '11.0.0');
     }
+
+    public validateIdentifier(identifier: string, identifierLabel: string): string | undefined {
```
nturinski, line 149: Why do you need a separate identifierLabel?  Is this not just used to validate names?  Or are there plans to expand its usage?  Would these rules apply to all properties?

src/postgres/commands/createPostgresFunction.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ext } from "../../extensionVariables";
+import { PostgresFunctionsTreeItem } from "../tree/PostgresFunctionsTreeItem";
+
+export async function createPostgresFunction(context: IActionContext, treeItem?: PostgresFunctionsTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = <PostgresFunctionsTreeItem>await ext.tree.showTreeItemPicker(PostgresFunctionsTreeItem.contextValue, context);
+    }
+
+    const child = await treeItem.createChild(context);
+    await commands.executeCommand("cosmosDB.openPostgresFunction", child);
```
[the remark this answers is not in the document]
wwlorey: Lol that makes too much sense 😜

src/postgres/tree/PostgresFunctionTreeItem.ts
```diff
@@ -16,19 +16,21 @@ export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext
...
         super(parent);
         this.schema = row.schema;
         this.name = row.name;
-        this.id = String(row.oid);
         this.definition = row.definition;
         this.isDuplicate = isDuplicate;
     }
 
+    public get id(): string {
+        return `${this.schema}.${this.name}`;
```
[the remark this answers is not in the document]
wwlorey: After you create a function on the server you would have to do a nasty query to get the oid for the tree item 🤢 So I thought this was a cleaner alternative

src/postgres/tree/PostgresServerTreeItem.ts
wwlorey: Those naming rules apply to any postgres identifier, so tables, funcs, stored procs, etc. I think it would be safe to drop `identifierLabel` and just use "Name"

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +34,69 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+            schema = schemas[0];
+        } else {
+            const schemaQuickPicks = schemas.map(s => { return { label: s }; });
+            schema = (await ext.ui.showQuickPick(schemaQuickPicks, {
+                placeHolder: localize('selectSchema', 'Select schema for new function...')
+            })).label;
+        }
+
+        const cachedChildren: AzExtTreeItem[] = await this.getCachedChildren(context);
+        const existingFunctionsInSchema: Set<string> = new Set();
```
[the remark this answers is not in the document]
wwlorey: Yeah that makes sense - creating those sets each time a child is created could get excessive

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -18,6 +20,8 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
     public readonly parent: PostgresDatabaseTreeItem;
     public clientConfig: ClientConfig;
 
+    private _functionsAndSchemas: {}; // Function name to list of schemas
```
wwlorey, line 23: Is something like this what you had in mind @nturinski? It made sense to just have func names as the properties instead of `schemas: string[]`

nturinski: Well, what I had was {schemas: string[]} which was type defining an object where the key is the function name and the value would be the array of schemas that have that name.

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +36,59 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+            })).label;
+        }
+
+        const name: string = (await ext.ui.showInputBox({
+            prompt: localize('enterFunctionName', 'Enter function name'),
+            validateInput: value => this.validateFunctionName(value, schema)
+        })).trim();
+
+        const schemasContainingFunction: string[] = this._functionsAndSchemas[name] || [];
+        const isDuplicate: boolean = schemasContainingFunction.length > 1 || (!schemasContainingFunction.includes(schema) && schemasContainingFunction.length === 1);
```
nturinski, line 75: What edge case is this? `(!schemasContainingFunction.includes(schema) && schemasContainingFunction.length === 1);`

Can you have two functions with the same name and schema?

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +36,59 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
     }
 
     public async loadMoreChildrenImpl(): Promise<PostgresFunctionTreeItem[]> {
+        const rows: IPostgresProceduresQueryRow[] = await this.listFunctions();
+        this._functionsAndSchemas = {};
```
nturinski, line 40: Do you want to clear out `this._functionsAndSchemas` every time you loadMoreChildren? I figured we only need to clear the object when we do a refresh

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +36,59 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+            schema = schemas[0];
+        } else {
+            const schemaQuickPicks = schemas.map(s => { return { label: s }; });
+            schema = (await ext.ui.showQuickPick(schemaQuickPicks, {
+                placeHolder: localize('selectSchema', 'Select schema for new function...')
+            })).label;
+        }
+
+        const name: string = (await ext.ui.showInputBox({
+            prompt: localize('enterFunctionName', 'Enter function name'),
```
nturinski, line 70: nit: Provde function name

src/postgres/tree/PostgresFunctionsTreeItem.ts
nturinski: DON'T YOU THUMBS UP A COMMENT WHERE I ADMIT TO BE DUMB WILL

wwlorey: 😂😂 You know what I mean

src/postgres/tree/PostgresFunctionsTreeItem.ts
wwlorey: This was supposed to handle when one other schema had this function name already. But I just realized I can remove it if I just update `functionsAndSchemas` before I check

src/postgres/tree/PostgresFunctionsTreeItem.ts
wwlorey: If we didn't clear it, would we only populate new entries in `functionsAndSchemas`? I feel like we would still need to get all the rows from the server to see if anything changed

nturinski: I was thinking a refresh would be able to take care of that, but load children gets called when you expand/refresh anyway (I think?) so it's probably better to do it here anyway.

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +36,63 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+            validateInput: value => this.validateFunctionName(value, schema)
+        })).trim();
+
+        if (this._functionsAndSchemas[name]) {
+            this._functionsAndSchemas[name].push(schema);
+        } else {
+            this._functionsAndSchemas[name] = [schema];
+        }
+
+        const definition: string = defaultFunctionDefinition(schema, name);
```
nturinski, line 80: Call this before you add to the _functionsAndSchemas array.  That way if the query errors out, you won't have pushed in the schema.

src/postgres/tree/PostgresFunctionsTreeItem.ts
```diff
@@ -32,13 +36,63 @@ export class PostgresFunctionsTreeItem extends AzureParentTreeItem<ISubscription
...
+                placeHolder: localize('selectSchema', 'Select schema for new function')
+            })).label;
+        }
+
+        const name: string = (await ext.ui.showInputBox({
+            prompt: localize('provideFunctionName', 'Provide function name'),
+            validateInput: value => this.validateFunctionName(value, schema)
+        })).trim();
+
+        if (this._functionsAndSchemas[name]) {
```
nturinski, line 74: nit: We could probably just extract this to a function since it's called twice in this file.

wwlorey:
This PR will be in flux for a little while, I'll ping once it's ready for a re-review

wwlorey:
Closing in favor of the query feature (starting with https://github.com/microsoft/vscode-cosmosdb/pull/1444)

PR #1421, opened 2020-04-15
Add support for deleting Postgres functions

wwlorey:

### DIFF
--- a/src/postgres/commands/deletePostgresFunction.ts
+++ b/src/postgres/commands/deletePostgresFunction.ts
@@ -1,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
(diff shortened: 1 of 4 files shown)


src/postgres/commands/deletePostgresFunction.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { window } from 'vscode';
+import { DialogResponses, IActionContext } from "vscode-azureextensionui";
+import { ext } from "../../extensionVariables";
+import { localize } from "../../utils/localize";
+import { PostgresFunctionTreeItem } from "../tree/PostgresFunctionTreeItem";
+
+export async function deletePostgresFunction(context: IActionContext, treeItem?: PostgresFunctionTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = <PostgresFunctionTreeItem>await ext.tree.showTreeItemPicker(PostgresFunctionTreeItem.contextValue, context);
```
EricJizbaMSFT, line 14: Instead of `context`, do this:
```
{ ...context, suppressCreatePick: true }
```

To prevent someone from creating then instantly deleting

PR #1427, opened 2020-04-20
Add schema description for duplicate Postgres table names

neelip:


nturinski:
I think you should wait to merge this one https://github.com/microsoft/vscode-cosmosdb/pull/1420 since it's using the same duplicate resource logic

package-lock.json
```diff
@@ -4,6 +4,24 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
```
EricJizbaMSFT, line 6: Revert changes to this file

nturinski:
This logic right here:
https://github.com/microsoft/vscode-cosmosdb/pull/1420/files#diff-466895c380cb7ae7167922111578c77eR81-R88

Rather than creating two sets every time the extension lists all of the resources, it uses a hashmap/object to store the name of resource and the schemes associated with it.  I personally find the duplicate set logic kind of 😬

Changing the loadMore logic seems like something we could do in a separate PR, but I don't have strong enough opinions to block this PR for that 😄

nturinski:
Could we change this duplicate duplicate (haha) logic before we merge this PR? 

https://github.com/microsoft/vscode-cosmosdb/blob/master/src/postgres/tree/PostgresFunctionsTreeItem.ts#L56-L64

Including this PR and Will's (#1427), there are 3 separate times this gets called.  I'd like to change it mostly because it's a waste to create 2 new Sets for everytime a PostgresTreeItem loads or refreshes.  

It also seems likely that other Postgres resources will use this logic, so I'd rather just consolidate it and make it a function on the `PostgresDatabaseTreeItem`.

Could either of you create a smaller PR that implements [this logic]( https://github.com/microsoft/vscode-cosmosdb/pull/1420/files#diff-466895c380cb7ae7167922111578c77eR81-R88)?

It will probably be a while before Will's Functions PR gets merged since it's a pretty complex feature. so it makes sense to just add that specific part in now.

nturinski:
@neelip Could you use the new function that Will added to `PostgresDatabaseTreeItem` to add children rather than the duplicate set logic?

neelip:
Sorry Nathan I missed your msg from yesterday. Thanks Will :)

PR #1429, opened 2020-04-21
Add arguments to Postgres function label

wwlorey:
It's possible to have multiple functions with the same name in a schema, each with different arguments. This accounts for that but it does look a little funky with the description in parenthesis too:

Screen Shot 2020-04-21 at 2 12 53 PM

### DIFF
--- a/src/postgres/tree/PostgresFunctionTreeItem.ts
+++ b/src/postgres/tree/PostgresFunctionTreeItem.ts
@@ -16,6 +16,7 @@ export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext
     public readonly parent: PostgresFunctionsTreeItem;
     public readonly schema: string;
     public readonly name: string;
+    public readonly args: string;
     public readonly id: string;
     public readonly isDuplicate: boolean;
     public definition: string;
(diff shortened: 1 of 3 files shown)


src/postgres/tree/PostgresFunctionTreeItem.ts
```diff
@@ -24,13 +25,14 @@ export class PostgresFunctionTreeItem extends AzureTreeItem<ISubscriptionContext
...
         this.name = row.name;
+        this.args = row.args;
         this.id = String(row.oid);
         this.definition = row.definition;
         this.isDuplicate = isDuplicate;
     }
 
     public get label(): string {
-        return this.name;
+        return `${this.name}(${this.args})`;
```
EricJizbaMSFT, line 35: I'm skeptical that this is common enough to be worth it. Even if there are duplicates, there's probably not that many and I can just click the function to open it up and view the args

wwlorey: I agree this is a pretty unlikely case. Updating the label is probably overkill, but I think it'll be worthwhile to add the functions's arguments to the `PostgresFunctionTreeItem` id once https://github.com/microsoft/vscode-cosmosdb/pull/1420 is merged. Otherwise, loading functions with the same name & different args will lead to this error:

Screen Shot 2020-04-22 at 12 42 20 PM

EricJizbaMSFT: Well to @nturinski's [point](https://github.com/microsoft/vscode-cosmosdb/pull/1420#discussion_r409911922) - that's why oid is preferred as the id. I think it's the only reliable way to uniquely identify a function

wwlorey: True - I'll add a query to get the oid after creating the function tree item. It'll prevent any more issues like this one

Issue #1430, opened 2020-04-22
Update repo name


wwlorey:
We'll also need to update in-code GitHub links to match

PR #1432, opened 2020-04-22
Rebrand to 'Azure Databases'

wwlorey:
A few notes/questions on this:

* I didn't touch any cosmos content in the README other than the title and a few command names
* For commands that only pertain to cosmos, I left 'Cosmos DB' as the category
* Should 'Cosmos DB' verbiage be left in the mongo experience [here](https://github.com/microsoft/vscode-cosmosdb/blob/wwl/rebrand/src/CosmosDBExperiences.ts#L90)?
* Should the API name `vscode-cosmosdbgraph.api` change? 
* Should appended labels like [this](https://github.com/microsoft/vscode-cosmosdb/blob/wwl/rebrand/src/extension.ts#L138) and [this](https://github.com/microsoft/vscode-cosmosdb/blob/wwl/rebrand/src/docdb/registerDocDBCommands.ts#L69) be changed?
* We still need to update the repo name & GitHub links in code https://github.com/microsoft/vscode-cosmosdb/issues/1430
* We still need to update the extension's icon https://github.com/microsoft/vscode-cosmosdb/issues/1431

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1399

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 
-# Cosmos DB Support for VS Code (Preview)
+# Azure Databases for VS Code (Preview)
 
 [![Version](https://vsmarketplacebadge.apphb.com/version/ms-azuretools.vscode-cosmosdb.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb) [![Installs](https://vsmarketplacebadge.apphb.com/installs-short/ms-azuretools.vscode-cosmosdb.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb) [![Build Status](https://dev.azure.com/ms-azuretools/AzCode/_apis/build/status/vscode-cosmosdb)](https://dev.azure.com/ms-azuretools/AzCode/_build/latest?definitionId=7)
 
--- a/package.json
+++ b/package.json
@@ -75,13 +75,13 @@
         "onCommand:cosmosDB.copyConnectionString",
         "onCommand:cosmosDB.launchMongoShell",
         "onCommand:cosmosDB.openGraphExplorer",
-        "onCommand:postgreSQL.createPostgresServer",
-        "onCommand:postgreSQL.deletePostgresServer",
-        "onCommand:postgreSQL.createPostgresDatabase",
-        "onCommand:postgreSQL.deletePostgresDatabase",
-        "onCommand:postgreSQL.deletePostgresTable",
+        "onCommand:postgreSQL.createServer",
+        "onCommand:postgreSQL.deleteServer",
+        "onCommand:postgreSQL.createDatabase",
+        "onCommand:postgreSQL.deleteDatabase",
+        "onCommand:postgreSQL.deleteTable",
         "onCommand:postgreSQL.openFunction",
-        "onCommand:postgreSQL.deletePostgresFunction",
+        "onCommand:postgreSQL.deleteFunction",
         "onCommand:postgreSQL.enterCredentials",
         "onCommand:postgreSQL.configureFirewall"
     ],
@@ -875,7 +875,7 @@
                     ],
                     "description": "The field values to display as labels in the treeview for Cosmos DB and MongoDB documents, in priority order"
                 },
-                "azureDatabases.showSavePrompt": {
+                "cosmosDB.showSavePrompt": {
                     "type": "boolean",
                     "default": true,
                     "description": "Show warning dialog when uploading a document to the cloud."
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
         "onCommand:cosmosDB.attachDatabaseAccount",
         "onCommand:cosmosDB.attachEmulator",
         "onCommand:cosmosDB.createAccount",
+        "onCommand:cosmosDB.createPostgresServer",
         "onCommand:cosmosDB.deleteAccount",
         "onCommand:cosmosDB.deletePostgresServer",
         "onCommand:cosmosDB.detachDatabaseAccount",
@@ -169,6 +170,15 @@
                 "category": "Cosmos DB",
                 "command": "cosmosDB.createAccount",
                 "title": "Create Account...",
+                "icon": {
+                    "light": "resources/icons/light/add.svg",
+                    "dark": "resources/icons/dark/add.svg"
+                }
+            },
+            {
+                "category": "PostgreSQL",
+                "command": "cosmosDB.createPostgresServer",
+                "title": "Create PostgreSQL Server...",
                 "icon": {
                     "light": "resources/icons/light/add.svg",
                     "dark": "resources/icons/dark/add.svg"
@@ -452,6 +462,11 @@
                     "group": "navigation"
                 },
                 {
+                    "command": "cosmosDB.createPostgresServer",
+                    "when": "view == cosmosDBExplorer",
+                    "group": "navigation"
+                },
+                {
                     "command": "cosmosDB.attachDatabaseAccount",
                     "when": "view == cosmosDBExplorer",
                     "group": "navigation"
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -1,9 +1,9 @@
 {
-    "name": "vscode-cosmosdb",
+    "name": "vscode-azuredatabases",
```
EricJizbaMSFT, line 2: Don't change the name yet. That has lots of ramifications that we should discuss/handle separately

package.json
```diff
@@ -96,9 +97,9 @@
         "views": {
             "azure": [
                 {
-                    "id": "cosmosDBExplorer",
-                    "name": "Cosmos DB",
-                    "when": "config.cosmosDB.showExplorer == true"
+                    "id": "azureDatabasesExplorer",
+                    "name": "Azure Databases",
```
EricJizbaMSFT, line 101: Remove "Azure" from this name. It's already under the "Azure" viewsContainer and would look weird if they only have this extension installed since VS Code appends the viewContainer name to the view name:
Screen Shot 2020-04-22 at 4 06 39 PM

package.json
```diff
@@ -854,22 +855,22 @@
...
                     "description": "Show warning dialog when uploading a document to the cloud."
                 },
-                "cosmosDB.enableOutputTimestamps": {
+                "azureDatabases.enableOutputTimestamps": {
                     "type": "boolean",
                     "default": true,
                     "description": "Prepends each line displayed in the output channel with a timestamp."
                 },
-                "cosmosDB.emulator.mongoPort": {
+                "azureDatabases.emulator.mongoPort": {
```
EricJizbaMSFT, line 868: Changing setting names will break users with existing settings and changing command names will break users with existing keybindings. For settings/commands that are unique to Cosmos DB - I would really prefer not to change those

src/mongo/registerMongoCommands.ts
```diff
@@ -18,7 +18,7 @@ import { MongoCollectionTreeItem } from "./tree/MongoCollectionTreeItem";
 import { MongoDatabaseTreeItem } from "./tree/MongoDatabaseTreeItem";
 import { MongoDocumentTreeItem } from "./tree/MongoDocumentTreeItem";
 
-const connectedDBKey: string = 'ms-azuretools.vscode-cosmosdb.connectedDB';
+const connectedDBKey: string = 'ms-azuretools.vscode-azuredatabases.connectedDB';
```
EricJizbaMSFT, line 21: This means users will lose the cache of their connected database. Since the key is completely internal and not displayed to the user - I would keep it as-is. Plus the feature is cosmos specific anyways

src/vscode-azuredatabases.api.d.ts
```diff
@@ -3,39 +3,39 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-export interface CosmosDBExtensionApi {
+export interface AzureDatabasesExtensionApi {
```
EricJizbaMSFT, line 6: > Should the API name vscode-cosmosdbgraph.api change?

No - I would not change this yet (or anything in this file). It is an internal detail that has no affect on our branding as far as the user is concerned. We will have to thoughtfully refactor this API to also support Postgres and I would prefer to change this file at that point

src/EditorManager.ts
```diff
@@ -41,19 +41,19 @@ export interface ShowEditorDocumentOptions {
 
 type PersistedEditors = { [key: string]: string };
 
-export class CosmosEditorManager {
-    private fileMap: { [key: string]: ICosmosEditor } = {};
+export class EditorManager {
```
EricJizbaMSFT, line 44: Can you please revert stuff like this that's just an internal engineering change so that the diff is much smaller? Branding is about strings visible to the user (not internal engineering changes) and seemingly simple changes can cause nasty bugs only discoverable once you release, so I want to be extra careful.

Plus - if you submit a separate PR with _just_ the engineering name changes, I can approve that very quickly since I know VS Code did most of the changes automatically for us anyways (and CI should catch any problems)

README.md
```diff
@@ -1,5 +1,5 @@
```
EricJizbaMSFT, line 1: > Should 'Cosmos DB' verbiage be left in the mongo experience here?

That branding is cloned exactly from the portal AFAIK so I would keep it

src/EditorManager.ts
```diff
@@ -65,7 +65,7 @@ export class CosmosEditorManager {
         }
 
         const localFilename = fileName.replace(/[<>:"\/\\|?*]/g, "-");
-        const localDocPath = path.join(os.tmpdir(), 'vscode-cosmosdb-editor', localFilename);
+        const localDocPath = path.join(os.tmpdir(), 'vscode-azuredatabases-editor', localFilename);
```
EricJizbaMSFT, line 68: > Should appended labels like this and this be changed

That feature is still specific to Cosmos DB, so I don't see a reason to change it.

package.json
```diff
@@ -176,170 +177,170 @@
             },
             {
                 "category": "Cosmos DB",
```
EricJizbaMSFT, line 179: > For commands that only pertain to cosmos, I left 'Cosmos DB' as the category

💯👍

package.json
```diff
@@ -21,7 +21,8 @@
         "DocumentDB",
         "Graph",
         "Gremlin",
-        "MongoDB"
+        "MongoDB",
+        "PostgreSQL"
     ],
     "preview": true,
     "homepage": "https://github.com/microsoft/vscode-cosmosdb/blob/master/README.md",
```
EricJizbaMSFT, line 28: > We still need to update the repo name & GitHub links in code #1430

I don't know that I would say _need_ for something that's mostly engineering. We definitely want to be careful with this so we don't lose anything

package.json
```diff
@@ -36,52 +37,52 @@
...
-        "onCommand:cosmosDB.loadMore",
-        "onCommand:cosmosDB.refresh",
-        "onCommand:cosmosDB.openInPortal",
-        "onCommand:cosmosDB.copyConnectionString",
-        "onCommand:cosmosDB.launchMongoShell",
-        "onCommand:cosmosDB.openGraphExplorer",
-        "onCommand:cosmosDB.enterPostgresCredentials",
-        "onCommand:cosmosDB.configurePostgresFirewall"
+        "onView:azureDatabasesExplorer",
+        "onCommand:azureDatabases.attachDatabaseAccount",
```
EricJizbaMSFT, line 41: I'm nervous about changing all these command ids. It breaks people's custom keybindings since the command id gets hard-coded in their keybindings.json file (See "Open Keyboard Shortcuts (JSON)" in the command palette). For example, we default `cosmosDB.executeMongoCommand` to `cmd+shift+'` for Mac, but maybe they changed it to something else.

How about we keep `cosmosDB.` as the prefix for cosmos-specific commands, use `postgreSQL.` for the prefix for any postgres-specific commands, and use `azureDatabases.` for any commands that apply to both. This _kind of_ follows the same pattern we use for the command category. One note assuming we do that: "postgreSQL.deletePostgresFunction" feels redundant, so I think I'd prefer the pattern "postgreSQL.deleteFunction"

src/CosmosEditorManager.ts
```diff
@@ -45,7 +45,7 @@ export class CosmosEditorManager {
     private fileMap: { [key: string]: ICosmosEditor } = {};
     private ignoreSave: boolean = false;
 
-    private readonly showSavePromptKey: string = 'cosmosDB.showSavePrompt';
+    private readonly showSavePromptKey: string = 'azureDatabases.showSavePrompt';
```
EricJizbaMSFT, line 48: A big part of why I'm having the discussion in https://github.com/microsoft/vscode-cosmosdb/pull/1442 is because of this setting. If we only use the editor for Cosmos DB and not PostgreSQL, I feel like we can keep this setting as-is and avoid people seeing the prompt after they've already clicked "Don't show again"

You could argue that now is a good time to break people (since we're doing all the re-branding anyways) in case other database types ever use the editor manager, but I feel like it might be possible we switch to the virtual file system provider before another database type ever needs this setting

PR #1436, opened 2020-04-24
Test cases 'creating accounts' are executed in parallel

v-wuzhai:


test/nightly/azureResourceGraph.test.ts
```diff
@@ -5,12 +5,11 @@
...
 import * as assert from 'assert';
-import { CosmosDBManagementModels } from 'azure-arm-cosmosdb';
 import * as vscode from 'vscode';
 import { randomUtils } from '../../extension.bundle';
 import { longRunningTestsEnabled, testUserInput } from '../global.test';
 import { getConnectionString } from './getConnectionString';
-import { client, resourceGroupsToDelete, testAccount } from './global.resource.test';
+import { accountList, api, resourceGrouList, resourceGroupsToDelete, testAccount } from './global.resource.test';
 
 suite('Graph action', async function (this: Mocha.Suite): Promise<void> {
```
nturinski, line 14: Could you add `require(./azureResourceCreateAccount.test.ts)`?  I think that this should worked based off this example https://stackoverflow.com/a/32934285

I'm nervous if the tests ever run out of order or we have a name that comes before the account creating test, then the accounts won't have been created.

test/nightly/azureResourceGraph.test.ts
```diff
@@ -23,22 +22,12 @@ suite('Graph action', async function (this: Mocha.Suite): Promise<void> {
             this.skip();
         }
         this.timeout(2 * 60 * 1000);
-        resourceGroupName = randomUtils.getRandomHexString(12);
-        // Cosmos DB account must have lower case name
-        accountName = randomUtils.getRandomHexString(12).toLowerCase();
+        resourceGroupName = resourceGrouList[api.Graph] === undefined ? '' : resourceGrouList[api.Graph];
```
nturinski, line 25: If we can't find the the graph account, we should throw an error rather than returning `''`.  That's just going to result in weird test errors.  But hopefully with the `require` method I was talking about, this won't be an issue, but we should still check regardless.

test/nightly/global.resource.test.ts
```diff
@@ -13,6 +13,13 @@ import { longRunningTestsEnabled } from '../global.test';
 export let testAccount: TestAzureAccount;
 export let client: CosmosDBManagementClient;
 export const resourceGroupsToDelete: string[] = [];
+export const accountList: {} = {};
+export const resourceGrouList: {} = {};
```
nturinski, line 17: Small typo :) `resourceGroupList`

test/nightly/azureResourceCreateAccount.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+suite('Create account', async function (this: Mocha.Suite): Promise<void> {
+    this.timeout(20 * 60 * 1000);
+
+    suiteSetup(async function (this: Mocha.Context): Promise<void> {
+        if (!longRunningTestsEnabled) {
+            this.skip();
+        }
+        this.timeout(15 * 60 * 1000);
+        await Promise.all([delayCreateAccount(5, /graph/), delayCreateAccount(10, /MongoDB/), delayCreateAccount(15, /SQL/)]);
```
nturinski, line 21: How are these numbers being selected?  5, 10, 15?  Why are they all different?  Are those average times?

test/nightly/azureResourceCreateAccount.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import * as assert from 'assert';
+import { CosmosDBManagementModels } from 'azure-arm-cosmosdb';
+import * as vscode from 'vscode';
+import { randomUtils } from '../../extension.bundle';
+import { longRunningTestsEnabled, testUserInput } from '../global.test';
+import { accountList, api, client, resourceGrouList } from './global.resource.test';
+
+suite('Create account', async function (this: Mocha.Suite): Promise<void> {
+    this.timeout(20 * 60 * 1000);
```
nturinski, line 14: Do we need two `this.timeout`s in this file?

test/nightly/azureResourceGraph.test.ts
v-wuzhai: I add require(./azureResourceCreateAccount.test.ts) and this file will run twice.

Note: I will create account in the "global.resource.test.ts" file. I think this problem can be solved.

test/nightly/azureResourceCreateAccount.test.ts
v-wuzhai: We can't create several accounts in the same time, so wait 5s to create the next account.

test/nightly/global.resource.test.ts
```diff
@@ -46,3 +56,29 @@ async function deleteResourceGroups(): Promise<void> {
         }
     }));
 }
+
+async function createAccount(accountType: RegExp): Promise<void> {
+    // Cosmos DB account must have lower case name
+    const accountName: string = randomUtils.getRandomHexString(12).toLowerCase();
+    const resourceGroupName: string = randomUtils.getRandomHexString(12);
```
EricJizbaMSFT, line 63: @wwlorey discovered today that there were quite a few Cosmos DB accounts that weren't being cleaned up properly, and I wonder if it's related to this PR. You should always add the following line _right after_ generating a resource group name:
```
resourceGroupsToDelete.push(resourceGroupName);
```

NOTE: Once you add it here, you can remove that line from `suiteSetup` in the other three files

PR #1442, opened 2020-04-28
Rename CosmosEditorManager

wwlorey:
Since we'll be using it with Postgres

### DIFF
--- a/src/EditorManager.ts
+++ b/src/EditorManager.ts
@@ -23,7 +23,7 @@ import { PostgresFunctionTreeItem } from './postgres/tree/PostgresFunctionTreeIt
 import { nonNullValue } from './utils/nonNull';
 import * as vscodeUtils from './utils/vscodeUtils';
 
-export interface ICosmosEditor<T = {}> {
+export interface IEditor<T = {}> {
     label: string;
     id: string;
     getData(context: IActionContext): Promise<T>;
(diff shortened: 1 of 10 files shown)


EricJizbaMSFT:
When will this be used for Postgres?

wwlorey:
It's used for editing functions and queries

EricJizbaMSFT:
For queries you execute the query with that little code lens thing you added similar to mongo scrapbooks. You don't save the file to upload to the cloud

And I thought we discussed in the last standup that we weren't going to let people edit functions directly - they would just use a query

wwlorey:
I thought if a function tree item was selected we would use this code to edit the function. Would we use code lens in that case?

EricJizbaMSFT:
I thought the two options were:
1. Open the whole query and the user executes the query (with codelens) to update the function
1. Open just the function and the user saves the file to upload

I'd prefer not to mix the two for a variety of reasons:
- Queries can get complex fast, so I feel like people would want to save them locally without uploading to the cloud
- The save action is hidden - people don't know ahead of time that it'll update the function. However, the Codelens thing literally says "Execute Query" in the file. At one point, we discussed adding a comment to the top of the file saying something like "Save this file to execute the query and upload this function", but the codelens thing already serves that purpose which is nice
- It's simpler code-wise to have a consistent "Execute Query" pattern (isn't that part of why you liked option 1 better?)
- Users could open the query and edit it to do something completely unrelated to a function. That could make your life difficult if you expect to get a function back after they save the file

wwlorey:
Ok that makes sense - no reason to mix & match how we open queries/functions 👍

PR #1444, opened 2020-04-29
Open Postgres functions with vscodeUtils & codelens instead of CosmosEditorManager

wwlorey:
This is the first chunk of the query feature

Related: https://github.com/microsoft/vscode-cosmosdb/pull/1442#issuecomment-621355904


src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -15,7 +20,16 @@ import { deletePostgresTable } from "./deletePostgresTable";
 import { enterPostgresCredentials } from "./enterPostgresCredentials";
 import { openPostgresFunction } from "./openPostgresFunction";
 
+export const connectedDBKey: string = 'ms-azuretools.vscode-azuredatabases.connectedPostgresDB';
```
nturinski, line 23: I think this should be named `connectedPostgresKey` and go back and rename `connectedDBKey` in `registerMongoCommands.ts` to `connectedMongoKey`.

Having them both be called connectedDBKey is confusing when you're importing them into files.

src/postgres/commands/connectPostgresDatabase.ts
```diff
@@ -0,0 +1,29 @@
...
+import { setConnectedDatabase } from "../setConnectedDatabase";
+import { PostgresDatabaseTreeItem } from "../tree/PostgresDatabaseTreeItem";
+import { connectedDBKey } from "./registerPostgresCommands";
+
+export async function connectPostgresDatabase(context: IActionContext, treeItem?: PostgresDatabaseTreeItem): Promise<void> {
+    if (!treeItem) {
+        treeItem = <PostgresDatabaseTreeItem>await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+    }
+
+    const oldTreeItemId: string | undefined = ext.connectedMongoDB && ext.connectedMongoDB.fullId;
```
nturinski, line 17: Did you forget to change `connectedMongoDB` to `connectedPostgresDB ` or do we not support connecting both at the same time?

src/postgres/setConnectedDatabase.ts
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ext } from "../extensionVariables";
+import { PostgresDatabaseTreeItem } from "./tree/PostgresDatabaseTreeItem";
+
+export function setConnectedDatabase(treeItem: PostgresDatabaseTreeItem | undefined): void {
```
nturinski, line 9: I don't think you need to make a command for it since it's only called once by `connectPostgresDatabase` and all it really does is just check if `connectedPostgresDB` exists in the `extensionVariables`.  Also, it shares a function name of `postgresCodeLensProvider` which confused me 😕

src/postgres/services/PostgresCodeLensProvider.ts
```diff
@@ -0,0 +1,66 @@
...
+        return this._onDidChangeEmitter.event;
+    }
+
+    public setConnectedDatabase(database: string | undefined): void {
+        this._connectedDatabase = database;
+        this._connectedDatabaseInitialized = true;
+        this._onDidChangeEmitter.fire();
+    }
+
+    public provideCodeLenses(): ProviderResult<CodeLens[]> {
```
nturinski, line 25: Just out of curiosity, what did you use as a reference for this?  I just want to know since I don't have a lot of knowledge on how to providing code lenses.

src/postgres/commands/connectPostgresDatabase.ts
wwlorey: Yeah I should have changed that

src/postgres/services/PostgresCodeLensProvider.ts
wwlorey: I modeled it after the mongo implementation: https://github.com/microsoft/vscode-cosmosdb/blob/master/src/mongo/services/MongoCodeLensProvider.ts

src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -25,4 +39,28 @@ export function registerPostgresCommands(): void {
     registerCommand('postgreSQL.deleteTable', deletePostgresTable);
     registerCommand('postgreSQL.openFunction', openPostgresFunction, doubleClickDebounceDelay);
     registerCommand('postgreSQL.deleteFunction', deletePostgresFunction);
+    registerCommand('postgreSQL.connectDatabase', connectPostgresDatabase);
+}
+
+export async function loadPersistedPostgresDatabase(): Promise<void> {
```
nturinski, line 45: I don't think this gets used outside of this file so no need for the export 🙂

src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -25,4 +39,28 @@ export function registerPostgresCommands(): void {
...
+    // NOTE: We want to make sure this function never throws or returns a rejected promise because it gets awaited multiple times
+    await callWithTelemetryAndErrorHandling('postgreSQL.loadPersistedDatabase', async (context: IActionContext) => {
+        context.errorHandling.suppressDisplay = true;
+        context.telemetry.properties.isActivationEvent = 'true';
+
+        try {
+            const persistedTreeItemId: string | undefined = ext.context.globalState.get(connectedPostgresKey);
+            if (persistedTreeItemId) {
+                const persistedTreeItem: PostgresDatabaseTreeItem = <PostgresDatabaseTreeItem>await ext.tree.findTreeItem(persistedTreeItemId, context);
```
nturinski, line 54: nit: Type should be `PostgresDatabaseTreeItem | undefined`

PR #1445, opened 2020-04-30
Add Postgres to existing CosmosDB "Create Account" command

neelip:
**Draft PR**
I am looking for feedback on the overall design before changing the syntax or styling.


src/CosmosDBExperiences.ts
```diff
@@ -11,12 +11,14 @@ export enum API {
     MongoDB = 'MongoDB',
```
wwlorey, line 11: This is pretty minor but I thought we decided on updating this to 'Add firewall rule for *your* IP "x.x.x.x"' just in case the IP we displayed didn't mean much to the user
Screen Shot 2020-05-04 at 11 44 40 AM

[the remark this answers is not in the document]
neelip: This fix is part of a separate PR. The other PR contains all smaller fixes, thought I would include with that

neelip: Fixed in different PR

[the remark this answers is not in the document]
neelip: Fixed in different PR #1447

neelip:
Using Azure Database "server" instead "resource" seems to get confusing within the code given the use of DatabaseAccount for CosmosDB and server for Postgres. Another option is to use "server" for output and "resource" for the code?

package.json
```diff
@@ -169,18 +168,9 @@
...
-                    "dark": "resources/icons/dark/add.svg"
-                }
-            },
-            {
-                "category": "PostgreSQL",
-                "command": "postgreSQL.createServer",
-                "title": "Create PostgreSQL Server...",
+                "category": "Azure Databases",
+                "command": "azureDatabases.createResource",
+                "title": "Create Resource...",
```
wwlorey, line 173: Should be "Create Server..."

src/tree/SubscriptionTreeItem.ts
```diff
@@ -61,32 +60,33 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         LocationListStep.addStep(wizardContext, promptSteps);
 
         const wizard = new AzureWizard(wizardContext, {
             promptSteps,
-            executeSteps: [
-                new CosmosDBAccountCreateStep()
-            ],
-            title: 'Create new Cosmos DB account'
+            executeSteps: [],
+            title: 'Create new Azure Database Resource'
```
wwlorey, line 74: Does this wording sound ok?
```suggestion
            title: 'Create new Azure Database Server'
```

EricJizbaMSFT:
I think I agree with Will that consistently using Server would be the least confusing

neelip:
Does that mean renaming "DatabaseAccount" in code for CosmosDB or just output?

EricJizbaMSFT:
I would stick with DatabaseAccount for anything specific to Cosmos DB

src/postgres/commands/PostgresAccountWizard/PostgresServerCreateStep.ts
```diff
@@ -18,14 +18,14 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresWi
 
         const locationName = nonNullProp(nonNullProp(wizardContext, 'location'), 'name');
         const rgName: string = nonNullProp(nonNullProp(wizardContext, 'resourceGroup'), 'name');
-        const serverName = nonNullProp(wizardContext, 'newServerName');
+        const serverName = nonNullProp(wizardContext, 'serverName');
```
EricJizbaMSFT, line 21: I believe we've discussed this before, but please continue to use `newServerName` like it was before instead of just `serverName`

src/tree/AzureDBAPIStep.ts
```diff
@@ -0,0 +1,67 @@
...
+            const executeSteps: AzureWizardExecuteStep<IPostgresWizardContext>[] = [
+                new PostgresServerCreateStep(),
+                new PostgresServerSetCredentialsStep(),
+                new PostgresServerSetFirewallStep()
+            ];
+            // tslint:disable-next-line: no-unnecessary-local-variable
+            const wizardOptions: IWizardOptions<IPostgresWizardContext> = { promptSteps, executeSteps };
+            return wizardOptions;
+        }
+        const promptSteps: AzureWizardPromptStep<ICosmosDBWizardContext>[] = [
```
EricJizbaMSFT, line 54: I'm a big fan of else cases and would suggest you almost always use one in situations like this. For example, if you forgot the line `return wizardOptions;` above, today it would fall through and return the cosmos stuff instead. If you used an else case and forgot that line, you would get a build/lint error saying this method doesn't always return a value. The build error makes it much easier to prevent bugs before anyone even starts testing

It would also fix your tslint errors about shadowed variables. I don't think I would ever disable that rule - I would always refactor to fix it

src/tree/CosmosDBAccountWizard/CosmosDBAccountCreateStep.ts
```diff
@@ -31,19 +31,19 @@ export class CosmosDBAccountCreateStep extends AzureWizardExecuteStep<ICosmosDBW
...
         }
 
-        wizardContext.databaseAccount = await client.databaseAccounts.createOrUpdate(rgName, accountName, options);
+        wizardContext.databaseAccount = await client.databaseAccounts.createOrUpdate(rgName, serverName, options);
 
         // createOrUpdate always returns an empty object - so we have to get the DatabaseAccount separately
-        wizardContext.databaseAccount = await client.databaseAccounts.get(rgName, accountName);
-        ext.outputChannel.appendLog(`Successfully created Cosmos DB account "${accountName}".`);
+        wizardContext.databaseAccount = await client.databaseAccounts.get(rgName, serverName);
+        ext.outputChannel.appendLog(`Successfully created Cosmos DB server "${serverName}".`);
```
EricJizbaMSFT, line 46: Right so the Cosmos DB-specific messages (like in this file) should probably still use "account". I don't care about the internal code (variable names, etc.) - just the user-visible strings

src/tree/SubscriptionTreeItem.ts
```diff
@@ -61,32 +60,33 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+        context.showCreatingTreeItem(resourceName);
         await wizard.execute();
         // don't wait
-        vscode.window.showInformationMessage(`Successfully created account "${accountName}".`);
+        vscode.window.showInformationMessage(`Successfully created server "${resourceName}".`);
+        if (wizardContext.defaultExperience?.api === API.Postgres) {
+            return new PostgresServerTreeItem(this, nonNullProp(wizardContext, 'server'));
+
+        }
         return await this.initCosmosDBChild(client, nonNullProp(wizardContext, 'databaseAccount'));
```
EricJizbaMSFT, line 90: Again - use an else case

Another option is this:
```
return wizardContext.defaultExperience?.api === API.Postgres ?
    new PostgresServerTreeItem(this, nonNullProp(wizardContext, 'server')) :
    await this.initCosmosDBChild(client, nonNullProp(wizardContext, 'databaseAccount'));
```

src/postgres/commands/PostgresAccountWizard/PostgresServerCreateStep.ts
neelip: Do you mean just in this file or in IAzureDBWizardContext as well?

EricJizbaMSFT: In the wizard context, too. Basically everywhere

src/tree/CosmosDBAccountWizard/CosmosDBAccountCreateStep.ts
neelip: What about this message in SubscriptionTreeItem.ts `vscode.window.showInformationMessage(`Successfully created server "${resourceName}".`);`
This applies to both Cosmos and Postgres. Should we have separate messages for both or keep it as it is with "server" in the string?

EricJizbaMSFT: The correct thing would be to have separate messages when possible. This file is easy to change so I commented on it. That example is a little more annoying to change, so I don't care as much

src/tree/AzureDBAPIStep.ts
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+            promptSteps = [
+                new CosmosDBAccountNameStep()
+            ];
+            executeSteps = [
+                new CosmosDBAccountCreateStep()
+            ];
+            // tslint:disable-next-line: no-unnecessary-local-variable
+            const wizardOptions: IWizardOptions<ICosmosDBWizardContext> = { promptSteps, executeSteps };
+            return wizardOptions;
```
EricJizbaMSFT, line 63: 😝 I feel like you're a little too quick to disable tslint rules sometimes. It's saying the variable `wizardOptions` is unnecessary and you could just do this instead:
```
return { promptSteps, executeSteps };
```

I feel like that looks a lot cleaner

src/tree/AzureDBAPIStep.ts
```diff
@@ -0,0 +1,70 @@
...
+        const result: IAzureQuickPickItem<Experience> = await ext.ui.showQuickPick(picks, {
+            placeHolder: localize('selectDBServerMsg', 'Select an Azure Database Server.')
+        });
+
+        wizardContext.defaultExperience = result.data;
+    }
+
+    public async getSubWizard(wizardContext: IAzureDBWizardContext): Promise<IWizardOptions<IPostgresWizardContext | ICosmosDBWizardContext>> {
+        let promptSteps: AzureWizardPromptStep<IPostgresWizardContext | ICosmosDBWizardContext>[];
+        let executeSteps: AzureWizardExecuteStep<IPostgresWizardContext | ICosmosDBWizardContext>[];
```
EricJizbaMSFT, line 37: The way you defined these variables is a little odd to me. In general variables should be defined in the minimal scope that they're used. For example:

Option 1: They're defined  and used _outside_ the if/else cases
```
let promptSteps: AzureWizardPromptStep<IPostgresWizardContext | ICosmosDBWizardContext>[];
let executeSteps: AzureWizardExecuteStep<IPostgresWizardContext | ICosmosDBWizardContext>[];
if () {
    promptSteps = [];
    executeSteps = [];
} else {
    promptSteps = [];
    executeSteps = [];
}
return { promptSteps, executeSteps };
```

Option 2: They're defined  and used _inside_ the if/else cases
```
if () {
    const promptSteps: AzureWizardPromptStep<IPostgresWizardContext>[] = [];
    const executeSteps: AzureWizardExecuteStep<IPostgresWizardContext>[] = [];
    return { promptSteps, executeSteps };
} else {
    const promptSteps: AzureWizardPromptStep<ICosmosDBWizardContext>[] = [];
    const executeSteps: AzureWizardExecuteStep<ICosmosDBWizardContext>[] = [];
    return { promptSteps, executeSteps };
}
```

src/tree/AzureDBAPIStep.ts
neelip: I was having some problems with type casting in beginning, wizardOptions was the fix and 
I added that to this commit as well...don't really need it now 😅

src/tree/AzureDBAPIStep.ts
neelip: Agreed. This is carried from previous commits, missed to resolve it here.

PR #1447, opened 2020-05-04
Design fixes - icons, validation, columns

neelip:
- Added icons for Table and Column Names --> 
image 
image
- Fixes to validation messages --> Password and Server name 
- Added column tree item


src/postgres/tree/PostgresColumnTreeItem.ts
```diff
@@ -0,0 +1,34 @@
...
+    public get id(): string {
+        return String(this.columnName);
+    }
+
+    public get label(): string {
+        return this.columnName;
+    }
+
+    public get iconPath(): string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } {
+        return getThemeAgnosticIconPath('PostgresColumn.svg');
```
EricJizbaMSFT, line 31: Please post screenshots of the tree in both light and dark theme with your new icons

src/postgres/commands/PostgresAccountWizard/PostgresServerCredPWStep.ts
```diff
@@ -15,7 +15,10 @@ export class PostgresServerCredPWStep extends AzureWizardPromptStep<IPostgresWiz
         const user = nonNullProp(wizardContext, 'adminUser');
         wizardContext.adminPassword = (await ext.ui.showInputBox({
             placeHolder: localize('pwPlaceholder', 'Password'),
-            prompt: localize('enterPWPrompt', 'Enter administrator password for the server.'),
+            prompt: localize('enterPWPrompt', 'Enter administrator password for the server. ' +
```
EricJizbaMSFT, line 18: Don't add strings together like this inside `localize`. I know it makes it look a little cleaner, but I don't think the localization tools work unless it's one constant string

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -46,8 +46,12 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
     const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
 
     if (!availability.nameAvailable) {
-
-        return availability.message;
+        if (availability.reason === 'AlreadyExists') {
+            return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
+        } else if (!name.match(/^(?![-])[a-zA-Z0-9-]*(?<![-])$/)) {
```
EricJizbaMSFT, line 51: Put this check for special characters before you call `await client.checkNameAvailability.execute`. Remember these validate methods happen every keystroke, so we want to backload azure calls (which may take a while) as much as possible

src/postgres/tree/PostgresColumnTreeItem.ts
```diff
@@ -0,0 +1,34 @@
...
+    public readonly columnName: string;
+    public readonly parent: PostgresTableTreeItem;
+
+    constructor(parent: PostgresTableTreeItem, columnName: string) {
+        super(parent);
+        this.columnName = columnName;
+    }
+
+    public get id(): string {
+        return String(this.columnName);
```
EricJizbaMSFT, line 23: nit: `return this.columnName;` should suffice

src/postgres/commands/PostgresAccountWizard/PostgresServerCredPWStep.ts
```diff
@@ -15,7 +15,10 @@ export class PostgresServerCredPWStep extends AzureWizardPromptStep<IPostgresWiz
...
+                '**Note** Password must contain characters from three of the following categories: ' +
+                '"uppercase letters", "lowercase letters", "numbers (0-9)", and "non-alphanumeric characteries (!, $, etc.)"'
```
wwlorey, line 20: nit: I know "characteries" is technically correct here (I didn't know it was a word until today lol) but "characters" makes more sense at least to me
```suggestion
                '"uppercase letters", "lowercase letters", "numbers (0-9)", and "non-alphanumeric characters (!, $, etc.)"'
```

src/postgres/tree/PostgresColumnTreeItem.ts
neelip: image
image
Icons are not visible with Abyss and Dark theme. 
Can we modify these icons or have custom icons or do we have to specifically use icons from https://cds-icons.azurewebsites.net/ ?

EricJizbaMSFT: Well you can also try the VS Code icon set, but I didn't see any table stuff on a quick glance: https://github.com/Microsoft/vscode-icons

We definitely need something that works with default light and default dark, so maybe delay the icon part of this PR. Personally, I think the Azure ones you found look pretty ugly anyways haha. I might be making this up, but I thought Matt he said he'd help with icons - so maybe ask him?

src/postgres/commands/PostgresAccountWizard/PostgresServerCredPWStep.ts
neelip: Lol its a typo 😅 I didn't know characteries was word, thanks for catching it

resources/icons/theme-agnostic/PostgresColumn.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 9"><title>Asset 28</title><rect y="8" width="7" height="1"/><rect x="9" y="8" width="7" height="1"/><rect y="4" width="7" height="1"/><rect x="9" y="4" width="7" height="1"/><rect width="7" height="1"/><rect x="9" width="7" height="1"/></svg>
```
wwlorey, line 1: These unused icon files should be removed

src/postgres/commands/PostgresAccountWizard/PostgresServerCredPWStep.ts
```diff
@@ -14,8 +14,8 @@ export class PostgresServerCredPWStep extends AzureWizardPromptStep<IPostgresWiz
     public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
         const user = nonNullProp(wizardContext, 'adminUser');
         wizardContext.adminPassword = (await ext.ui.showInputBox({
-            placeHolder: localize('pwPlaceholder', 'Password'),
-            prompt: localize('enterPWPrompt', 'Enter administrator password for the server.'),
+            placeHolder: localize('pwPlaceholder', 'Administrator Password'),
+            prompt: localize('enterPWPrompt', 'Password must contain characters from three of the following categories: uppercase letters, lowercase letters, numbers (0-9), and non-alphanumeric characters (!, $, etc.).'),
```
wwlorey, line 18: nit: This string could be a constant since it's used again below

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -43,11 +42,18 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
...
+    const characterMatch = name.match(/^(?![-])[a-zA-Z0-9-]*(?<![-])$/);
+    const prefixSuffixMatch = name.match(/^[-]*.*[-]$/);
     const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
 
     if (!availability.nameAvailable) {
-
-        return availability.message;
+        if (availability.reason === 'AlreadyExists') {
+            return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
+        } else if (prefixSuffixMatch) {
```
wwlorey, line 52: This condition wasn't met for a server name starting with a hyphen
Screen Shot 2020-05-07 at 10 35 35 AM

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -43,11 +42,18 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
...
+    const characterMatch = name.match(/^(?![-])[a-zA-Z0-9-]*(?<![-])$/);
+    const prefixSuffixMatch = name.match(/^[-]*.*[-]*$/);
     const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
 
     if (!availability.nameAvailable) {
-
-        return availability.message;
+        if (availability.reason === 'AlreadyExists') {
+            return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
+        } else if (prefixSuffixMatch) {
```
wwlorey, line 52: There's still some weirdness here...
Screen Shot 2020-05-07 at 1 09 39 PM

What if we do this instead?
```suggestion
        } else if (name[0] === '-' || name[name.length - 1] === '-') {
```

I like to avoid regex in cases like this, it can get tricky lol

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -43,11 +42,18 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
...
+            return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
+        } else if (!characterMatch) {
```
wwlorey, line 54: If the hyphen check is fixed, capital letters aren't caught with this check
Screen Shot 2020-05-07 at 1 19 51 PM

We should be able to simplify the regex to this
```suggestion
        } else if (name.match(/[^a-z\d-]/)) {
```

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
EricJizbaMSFT: I agree a RegExp is probably overkill in this case, but I would suggest this instead:

```suggestion
        } else if (name.startsWith('-') || name.endsWith('-')) {
```

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
EricJizbaMSFT: I like where Will is going, but a few nits:

```suggestion
        } else if (/[^a-z0-9-]/.test(name)) {
```


I prefer `<regexp>.test(<string>)` instead of `<string>.match(<regexp>)` because it returns a boolean instead of `RegExpMatchArray | null`. 
I also prefer `0-9` instead of `\d` because it feels more read-able to me. No need to remember that "d" stands for "digit", that it has to be `\` and not `/`, and that it has to be `d` and not `D` (which is the opposite)

neelip: Thanks for catching, I missed that the uppercase letters were excluded for serverName

src/postgres/commands/PostgresAccountWizard/PostgresServerCredPWStep.ts
```diff
@@ -13,11 +13,12 @@ export class PostgresServerCredPWStep extends AzureWizardPromptStep<IPostgresWiz
 
     public async prompt(wizardContext: IPostgresWizardContext): Promise<void> {
         const user = nonNullProp(wizardContext, 'adminUser');
+        const pwConditionMsg = localize('passwordConditionMsg', 'Password must contain characters from three of the following categories - uppercase letters, lowercase letters, numbers (0-9), and non-alphanumeric characters (!, $, etc.).');
```
EricJizbaMSFT, line 16: nit: we usually don't pass string constants as parameters. You can just move the string to the class or file level so that it's in scope for both usages.

src/postgres/commands/PostgresAccountWizard/PostgresServerNameStep.ts
```diff
@@ -43,11 +42,17 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
...
     const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
+    const characterMatch = name.match(/^[a-z\d-]+$/);
     const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
 
     if (!availability.nameAvailable) {
-
-        return availability.message;
+        if (availability.reason === 'AlreadyExists') {
+            return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
+        } else if (!characterMatch) {
```
EricJizbaMSFT, line 51: Seems like Will and I both prefer to have the RegExp in the condition itself instead of a separate variable. To me, it makes it more read-able to have the RegExp and the message right next to each other. I also just prefer fewer characters when possible haha 🤷‍♂️ Any particular reason you prefer the variable?

neelip: Lol 😅 you mentioned in earlier suggestion to place the check before await to reduce delay https://github.com/microsoft/vscode-cosmosdb/pull/1447#discussion_r419774990

EricJizbaMSFT: Oh yes that's correct, but actually your existing code doesn't reduce delay since you await the call to check name availability before you return the character check message. You need to do something like this:

```
if (name.length < min || name.length > max) {
    return localize('serverNameLengthCheck', 'The name must be between {0} and {1} characters.', min, max);
} else if (!name.match(/^[a-z\d-]+$/)) {
    return localize('serverNameCharacterCheck', 'Server name must only contain lowercase letters, numbers, and hyphens.');
} else if (name.startsWith('-') || name.endsWith('-')) {
    return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
} else {
    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
    if (!availability.nameAvailable) {
        if (availability.reason === 'AlreadyExists') {
            return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
        } else {
            return availability.message;
        }
    }
}
```

It can't even reach the async code if it has invalid characters.

neelip: You're right, for some reason I was thinking I needed NameAvailability for the regex check

EricJizbaMSFT:
@wwlorey @neelip anything else blocking this PR from getting merged?

PR #1424, opened 2020-04-17
Create Graph Database

v-wuzhai:

PR #1448, opened 2020-05-04
Bump shared packages

EricJizbaMSFT:
We no longer need to set reporter

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,24 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
+        "@azure/cosmos": {
+            "version": "3.6.3",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -4,6 +4,24 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
+        "@azure/cosmos": {
```
EricJizbaMSFT, line 7: Turns out this was added in https://github.com/microsoft/vscode-cosmosdb/pull/1424, but package-lock.json was never updated

EricJizbaMSFT: Changed Dev Ops to use `npm ci` to prevent this problem. See this build as an example: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=18710&view=results

PR #1449, opened 2020-05-05
Add 'Create Query From Template' experience for Postgres functions

wwlorey:


src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryNameStep.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ext } from "../../../../extensionVariables";
+import { localize } from "../../../../utils/localize";
+import { PostgresServerTreeItem } from "../../../tree/PostgresServerTreeItem";
+import { IPostgresFunctionQueryWizardContext } from "./IPostgresFunctionQueryWizardContext";
+
+export class FunctionQueryNameStep extends AzureWizardPromptStep<IPostgresFunctionQueryWizardContext> {
+    public async prompt(wizardContext: IPostgresFunctionQueryWizardContext): Promise<void> {
+        wizardContext.name = (await ext.ui.showInputBox({
+            prompt: localize('provideFunctionName', 'Provide function name'),
+            validateInput: PostgresServerTreeItem.validateIdentifier
```
wwlorey, line 16: I think it's little weird to validate the function name when it can be changed to anything once the query is created. But this seemed better than having no validation

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryReturnTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            return { promptSteps: [new FunctionQueryCustomReturnTypeStep()] };
+        }
+
+        wizardContext.returnType = wizardContext.returnTypePick.label;
+        return undefined;
+    }
+}
+
+// A subset of return types available on pgAdmin
+const returnTypes: string[] = [
```
wwlorey, line 40: I paired down the full list of return types. Open to suggestions for showing more/less options
<details>
  <summary>Full list</summary>

```
"char"
"char"[]
abstime
abstime[]
aclitem
aclitem[]
anyarray
anyelement
anyenum
anynonarray
anyrange
bigint
bigint[]
bit
bit varying
bit varying[]
bit[]
boolean
boolean[]
box
box[]
bytea
bytea[]
character
character
character varying
character[]
cid
cid[]
cidr
cidr[]
circle
circle[]
cstring
cstring[]
date
date[]
daterange
daterange[]
double precision
double precision[]
fdw_handler
gtsvector
gtsvector[]
index_am_handler
inet
inet[]
int2vector
int2vector[]
int4range
int4range[]
int8range
int8range[]
integer
integer[]
internal
interval
interval[]
json
json[]
jsonb
jsonb[]
line
line[]
lseg
lseg[]
macaddr
macaddr8
macaddr8[]
macaddr[]
money
money[]
name
name[]
numeric
numeric[]
numrange
numrange[]
oid
oid[]
oidvector
oidvector[]
opaque
path
path[]
pg_aggregate
pg_am
pg_amop
pg_amproc
pg_attrdef
pg_attribute
pg_auth_members
pg_authid
pg_available_extension_versions
pg_available_extensions
pg_buffercache
pg_cast
pg_class
pg_collation
pg_config
pg_constraint
pg_conversion
pg_cursors
pg_database
pg_db_role_setting
pg_ddl_command
pg_default_acl
pg_depend
pg_dependencies
pg_description
pg_enum
pg_event_trigger
pg_extension
pg_file_settings
pg_foreign_data_wrapper
pg_foreign_server
pg_foreign_table
pg_group
pg_hba_file_rules
pg_index
pg_indexes
pg_inherits
pg_init_privs
pg_language
pg_largeobject
pg_largeobject_metadata
pg_locks
pg_lsn
pg_lsn[]
pg_matviews
pg_namespace
pg_ndistinct
pg_node_tree
pg_opclass
pg_operator
pg_opfamily
pg_partitioned_table
pg_pltemplate
pg_policies
pg_policy
pg_prepared_statements
pg_prepared_xacts
pg_proc
pg_publication
pg_publication_rel
pg_publication_tables
pg_range
pg_replication_origin
pg_replication_origin_status
pg_replication_slots
pg_rewrite
pg_roles
pg_rules
pg_seclabel
pg_seclabels
pg_sequence
pg_sequences
pg_settings
pg_shadow
pg_shdepend
pg_shdescription
pg_shseclabel
pg_stat_activity
pg_stat_all_indexes
pg_stat_all_tables
pg_stat_archiver
pg_stat_bgwriter
pg_stat_database
pg_stat_database_conflicts
pg_stat_progress_vacuum
pg_stat_replication
pg_stat_ssl
pg_stat_statements
pg_stat_subscription
pg_stat_sys_indexes
pg_stat_sys_tables
pg_stat_user_functions
pg_stat_user_indexes
pg_stat_user_tables
pg_stat_wal_receiver
pg_stat_xact_all_tables
pg_stat_xact_sys_tables
pg_stat_xact_user_functions
pg_stat_xact_user_tables
pg_statio_all_indexes
pg_statio_all_sequences
pg_statio_all_tables
pg_statio_sys_indexes
pg_statio_sys_sequences
pg_statio_sys_tables
pg_statio_user_indexes
pg_statio_user_sequences
pg_statio_user_tables
pg_statistic
pg_statistic_ext
pg_stats
pg_subscription
pg_subscription_rel
pg_tables
pg_tablespace
pg_timezone_abbrevs
pg_timezone_names
pg_transform
pg_trigger
pg_ts_config
pg_ts_config_map
pg_ts_dict
pg_ts_parser
pg_ts_template
pg_type
pg_user
pg_user_mapping
pg_user_mappings
pg_views
point
point[]
polygon
polygon[]
real
real[]
record
record[]
refcursor
refcursor[]
regclass
regclass[]
regconfig
regconfig[]
regdictionary
regdictionary[]
regnamespace
regnamespace[]
regoper
regoper[]
regoperator
regoperator[]
regproc
regproc[]
regprocedure
regprocedure[]
regrole
regrole[]
regtype
regtype[]
reltime
reltime[]
smallint
smallint[]
smgr
tab
tab[]
tabl
tabl[]
test.tab
text
text[]
tid
tid[]
time with time zone
time with time zone[]
time without time zone
time without time zone[]
timestamp with time zone
timestamp with time zone[]
timestamp without time zone
timestamp without time zone[]
tinterval
tinterval[]
tsm_handler
tsquery
tsquery[]
tsrange
tsrange[]
tstzrange
tstzrange[]
tsvector
tsvector[]
txid_snapshot
txid_snapshot[]
uuid
uuid[]
varying[]</li>
void
xid
xid[]
xml
xml[]
```
</details>

package.json
```diff
@@ -789,6 +795,10 @@
                 {
                     "command": "postgreSQL.connectDatabase",
                     "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
+                },
+                {
+                    "command": "postgreSQL.createQuery",
```
EricJizbaMSFT, line 800: As a user, if my main goal is to create a function, I don't think I would know to right click at the database level and click "Create Query from template...". We don't have plans for a single other template (right?), so I'm not sure a generic "Create Query From Template..." command is the best idea at the moment.

I see this going one of two ways:
1. We figure out enough templates to make a generic "Create Query From Template..." command worth it
1. We refactor the wizard to be function-specific.

The first one sounds pretty complicated, so I prefer the second option. Why were we opposed to a simple "Create function..." command on the "Functions" node? We might have to mess with the messaging since the user is actually creating a query they have to execute, but overall that entry point feels much simpler and more discoverable

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryNameStep.ts
EricJizbaMSFT: I agree some validation here is better than no validation

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryNameStep.ts
```diff
@@ -0,0 +1,23 @@
...
+        })).trim();
+    }
```
EricJizbaMSFT, line 18: Can we move `validateIdentifier` to this file? I only see it used here, so I'm not entirely sure why it's in `PostgresServerTreeItem.ts`

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryReturnTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            placeHolder: localize('selectReturnType', 'Select return type')
+        }));
+    }
+
+    public shouldPrompt(): boolean {
+        return true;
+    }
+
+    public async getSubWizard(wizardContext: IPostgresFunctionQueryWizardContext): Promise<IWizardOptions<IPostgresFunctionQueryWizardContext> | undefined> {
+        if (wizardContext.returnTypePick === customReturnTypePick) {
```
EricJizbaMSFT, line 30: So it's a little weird to me to have `returnTypePick` on the wizard context. I would leverage `IAzureQuickPickItem`, specifically the `data` field.

The picks could be defined like this:
```
const returnTypeQuickPicks: IAzureQuickPickItem<string | undefined>[] = returnTypes.map(r => { return { label: r, data: r }; });
returnTypeQuickPicks.push({ label: localize('enterCustomReturnType', '$(pencil) Enter custom return type...'), data: undefined });
```

You add just `data` to the wizardContext (not the pick) and then here you just do this:
```
return wizardContext.returnType ? undefined : { promptSteps: [new FunctionQueryCustomReturnTypeStep()] };
```

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/FunctionQueryReturnTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+        const returnTypeQuickPicks: QuickPickItem[] = returnTypes.map(r => { return { label: r }; });
+        returnTypeQuickPicks.push(customReturnTypePick);
+
+        wizardContext.returnTypePick = (await ext.ui.showQuickPick(returnTypeQuickPicks, {
+            placeHolder: localize('selectReturnType', 'Select return type')
+        }));
+    }
+
+    public shouldPrompt(): boolean {
+        return true;
```
EricJizbaMSFT, line 26: In general, `shouldPrompt` should be relative to the thing you're prompting about. So in this case:
```
public shouldPrompt(wizardContext: IPostgresFunctionQueryWizardContext): boolean {
    return !wizardContext.returnType;
}
```

The idea is that if someone sets the returnType before the wizard starts (aka some sort of smart default), it won't prompt for this step.

src/postgres/commands/PostgresQueryWizard/FunctionQuerySteps/IPostgresFunctionQueryWizardContext.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { QuickPickItem } from "vscode";
+import { IPostgresQueryWizardContext } from "../IPostgresQueryWizardContext";
+
+export interface IPostgresFunctionQueryWizardContext extends IPostgresQueryWizardContext {
+    name: string;
```
EricJizbaMSFT, line 10: wizard context properties should almost always be optional with `?` since they don't exist until after the prompt step. That means you'll have to use a decent amount of `nonNullProp` whenever you want to access it, but it does make it easer to find bugs in telemetry if we have any steps out of order

package.json
wwlorey: If I remember right, we decided to go with the query feature so it could encompass stored procedures, tables, & anything else we might want to do via a query. And we wanted to avoid "Create Function..." just because it's really creating a query that would be executed to create a function. But I agree that creating a function isn't very discoverable under "Create Query From Template..." command.

I'd vote for refactoring the wizard to be function specific. For this release, we won't have any query options other than functions. And I would kinda like having separate commands for "Create Function Query From Template...", "Create Stored Procedure Query From Template...", etc. If that's something we want to do

wwlorey: Also, "Create Function Query From Template..." sounds kinda cumbersome if we go with that option. What do you think of just "Create Function Query..." ?

EricJizbaMSFT: Well...there's two query features here:
1. Execute query
1. Create query template

IMO 1 still has to be generic enough to cover "anything else we might want to do via a query", but 2 is less important in that regard.

I'm fine with either name for the function query thing - easy to change anyways

src/postgres/commands/createPostgresFunctionQuery.ts
```diff
@@ -0,0 +1,30 @@
...
+import { nonNullProp } from "../../utils/nonNull";
+import * as vscodeUtil from '../../utils/vscodeUtils';
+import { PostgresDatabaseTreeItem } from "../tree/PostgresDatabaseTreeItem";
+import { connectPostgresDatabase } from "./connectPostgresDatabase";
+import { FunctionQueryCreateStep } from "./FunctionQueryWizard/FunctionQueryCreateStep";
+import { FunctionQueryNameStep } from "./FunctionQueryWizard/FunctionQueryNameStep";
+import { FunctionQueryReturnTypeStep } from "./FunctionQueryWizard/FunctionQueryReturnTypeStep";
+import { IPostgresFunctionQueryWizardContext } from "./FunctionQueryWizard/IPostgresFunctionQueryWizardContext";
+
+export async function createPostgresFunctionQuery(wizardContext: IPostgresFunctionQueryWizardContext, treeItem?: PostgresDatabaseTreeItem): Promise<void> {
```
EricJizbaMSFT, line 16: You should do `context: IActionContext` as the first parameter and create the wizardContext below like this:
```
const wizardContext: IPostgresFunctionQueryWizardContext = context;
```

It might seem silly, but it has better type validation. For example, [here](https://github.com/microsoft/vscode-azurefunctions/blob/9d9a73014c5edb17db523ba907ab017ff005dabb/src/commands/createFunction/createFunction.ts#L45)'s an example in the functions repo where several properties like `projectPath` are required on the wizardContext (aka they have to be set before the wizard starts). If I accidentally misspell projectPath like this:
```
const wizardContext: IFunctionWizardContext = Object.assign(context, { projectOOPSPath ... });
```

It would show this error:
> Property 'projectPath' is missing in type 'IActionContext & { projectOOPSPath: string; workspacePath: string; workspaceFolder: WorkspaceFolder | undefined; version: FuncVersion; language: ProjectLanguage; functionName: string | undefined; }' but required in type 'IFunctionWizardContext'.ts(2741)

NOTE: An alternative fix is to turn on `strictFunctionTypes` in tsconfig, but that's a lot more work affecting the whole repo

src/postgres/commands/createPostgresFunctionQuery.ts
```diff
@@ -0,0 +1,30 @@
...
+        executeSteps: [new FunctionQueryCreateStep()],
+        title: 'Create PostgreSQL Function Query'
+    });
+
+    await wizard.prompt();
+    await wizard.execute();
+    await vscodeUtil.showNewFile(nonNullProp(wizardContext, 'query'), 'query', '.sql');
+
+    if (treeItem) {
+        await connectPostgresDatabase(wizardContext, treeItem);
```
EricJizbaMSFT, line 28: Isn't the entrypoint on the "Functions" tree item now and not the database tree item?

wwlorey: Yeah I forgot to change that

PR #1451, opened 2020-05-05
Add 'Connected' description to Postgres database

wwlorey:
Screen Shot 2020-05-05 at 4 09 42 PM

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -36,6 +36,10 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         return this.databaseName;
     }
 
+    public get description(): string {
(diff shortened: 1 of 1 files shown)


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -36,6 +36,10 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         return this.databaseName;
     }
 
+    public get description(): string {
+        return ext.connectedPostgresDB && ext.connectedPostgresDB.fullId === this.fullId ? 'Connected' : '';
```
EricJizbaMSFT, line 40: Does the node get refreshed when people connect/disconnect a database?

wwlorey: Yeah both the old node and the new connected node get refreshed

EricJizbaMSFT, line 40:
```suggestion
        return ext.connectedPostgresDB?.fullId === this.fullId ? localize('connected', 'Connected') : '';
```

PR #1453, opened 2020-05-07
Add support for executing Postgres queries

wwlorey:
I added 'returned fields' to the query output. Here's an example of querying a table followed by creating a function:

Screen Shot 2020-05-07 at 1 58 29 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,8 @@
         "onCommand:postgreSQL.deleteFunction",
         "onCommand:postgreSQL.enterCredentials",
         "onCommand:postgreSQL.configureFirewall",
-        "onCommand:postgreSQL.connectDatabase"
+        "onCommand:postgreSQL.connectDatabase",
+        "onCommand:postgreSQL.executeQuery"
     ],
     "contributes": {
         "viewsContainers": {
@@ -402,6 +403,11 @@
                 "category": "PostgreSQL",
                 "command": "postgreSQL.connectDatabase",
                 "title": "Connect to Database..."
+            },
+            {
+                "category": "PostgreSQL",
+                "command": "postgreSQL.executeQuery",
+                "title": "Execute Query..."
             }
         ],
         "menus": {
@@ -788,6 +794,10 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
+                    "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
+                },
+                {
+                    "command": "postgreSQL.executeQuery",
                     "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
                 }
             ],
--- a/package.json
+++ b/package.json
@@ -795,10 +795,6 @@
                 {
                     "command": "postgreSQL.connectDatabase",
                     "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
-                },
-                {
-                    "command": "postgreSQL.executeQuery",
-                    "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
                 }
             ],
             "explorer/context": [
@@ -835,6 +831,10 @@
                 {
                     "command": "cosmosDB.executeMongoCommand",
                     "when": "editorLangId == 'mongo'"
+                },
+                {
+                    "command": "postgreSQL.executeQuery",
+                    "when": "editorLangId == 'sql'"
                 }
             ]
         },
--- a/src/postgres/commands/executePostgresQuery.ts
+++ b/src/postgres/commands/executePostgresQuery.ts
@@ -3,12 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Client, QueryResult } from 'pg';
+import { Client, ClientConfig, QueryResult } from 'pg';
 import * as vscode from 'vscode';
-import { AzExtTreeItem, IActionContext } from 'vscode-azureextensionui';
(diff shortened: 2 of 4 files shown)


src/postgres/commands/executePostgresQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+            treeItem = ext.connectedPostgresDB;
+        } else {
+            treeItem = <PostgresDatabaseTreeItem>await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+        }
+    }
+
+    if (!treeItem.clientConfig) {
+        let children: AzExtTreeItem[] = await treeItem.loadAllChildren(context);
+
+        while (children.length === 1) {
```
EricJizbaMSFT, line 30: All of this tree stuff is just to make sure they have the firewall and credentials set up, correct? We can make this more efficient (and more readable IMO) if we avoid the tree to do this. Instead of the references looking like this:

> executePostgresQuery.ts --> PostgresDatabaseTreeItem.ts --> cred/firewall stuff

Hopefully you can refactor so that it looks more like this:

> executePostgresQuery.ts --> cred/firewall stuff (in a util file maybe?) <-- PostgresDatabaseTreeItem.ts

I assume the util file would have functions like `isFirewallSetup`, `validateFirewall`, etc.

src/postgres/commands/executePostgresQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+        throw new Error(localize('openQueryBeforeExecuting', 'Open a PostgreSQL query before executing.'));
+    }
+
+    const client: Client = new Client(treeItem.clientConfig);
+    await client.connect();
+    const queryResult: QueryResult = await client.query(query);
+
+    let resultString: string = localize('executedQuery', 'Successfully executed "{0}" query.', queryResult.command);
+    const fieldNames: string[] = queryResult.fields.map(field => field.name);
+    const fieldsString: string = queryResult.fields.length ? `[${fieldNames.join(', ')}]` : 'none';
```
EricJizbaMSFT, line 57: Can you explain exactly what field names is? Is it just the column names and not the actual values?

package.json
```diff
@@ -789,6 +795,10 @@
                 {
                     "command": "postgreSQL.connectDatabase",
                     "when": "view == azureDatabasesExplorer && viewItem == postgresDatabase"
+                },
+                {
+                    "command": "postgreSQL.executeQuery",
```
EricJizbaMSFT, line 800: With this entry point, you assume the active editor is the query they want to execute, but it might not be the right query or even a query at all. I would prefer to remove this entry point - users should only be able to execute a query if they have a ".sql" file open. When in doubt, I would model this after mongo scrapbook behavior

src/postgres/commands/executePostgresQuery.ts
wwlorey: Having the firewall and credentials set up was part of it but the main reason to load children here was to populate `clientConfig` so we could execute a query. I'll think more on how to populate that

src/postgres/commands/executePostgresQuery.ts
wwlorey: Yeah my original example was kinda bad - the fields are column names and you can have values associated with them. I just pushed a change that shows values as well. This may be closer to what we want (each entry in this example is just "data")

Screen Shot 2020-05-08 at 2 24 20 PM

EricJizbaMSFT: Still would prefer a csv (separate PR?) but better than displaying nothing

src/postgres/commands/executePostgresQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as vscode from 'vscode';
+import { IActionContext, IParsedError, parseError } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../utils/localize';
+import { firewallNotConfiguredErrorType, invalidCredentialsErrorType, PostgresDatabaseTreeItem } from '../tree/PostgresDatabaseTreeItem';
+import { configurePostgresFirewall } from './configurePostgresFirewall';
+import { enterPostgresCredentials } from './enterPostgresCredentials';
+import { loadPersistedPostgresDatabase } from './registerPostgresCommands';
+
+export async function executePostgresQuery(context: IActionContext, treeItem?: PostgresDatabaseTreeItem): Promise<void> {
```
EricJizbaMSFT, line 16: We got rid of the tree item entry point, right? So we can remove the `treeItem` parameter?

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -115,6 +100,30 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            const host: string = nonNullProp(this.parent.server, 'fullyQualifiedDomainName');
+            const clientConfig: ClientConfig = { user: username, password, ssl, host, port: 5432, database: this.databaseName };
+
+            // Ensure the client config is valid before returning
+            const client: Client = new Client(clientConfig);
+            await client.connect();
+
+            return clientConfig;
+        } else {
+            const invalidCredentialsError: Error & { code: string } = Object.assign(new Error(localize('mustEnterCredentials', 'Must enter credentials to connect to server.')), { code: invalidCredentialsErrorType });
```
EricJizbaMSFT, line 123: Lol this is a little funky. What about:
```
throw {
    message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.')),
    code: invalidCredentialsErrorType
};
```

Issue #1446, opened 2020-04-30
Credentials details are not deleted with Postgres account deletion


neelip:
Repro Steps:
1. Delete a Postgres account 
2. Create a new Postgres account on azure portal with the same name as the deleted account
3. Refresh tree view

Expected:
"Enter Credentials..." requirement should show up

Actual:
Enter Credentials is skipped and directly goes to "Configure Firewall..." requirement

PR #1454, opened 2020-05-07
Delete credentials when deleting a Postgres server

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1446


src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -109,6 +109,7 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
         const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
         const deletingMessage: string = `Deleting server "${this.name}"...`;
         await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: deletingMessage }, async () => {
+            await this.deletePostgresCredentials();
```
EricJizbaMSFT, line 112: Mmm I feel like this should be after `deleteMethod`? Like if the delete fails, we probably don't need to delete the credentials

PR #1457, opened 2020-05-12
Fix connecting to databases from the explorer context menu

wwlorey:
This fixes the "treeItem.refresh is not a function" error.

Also it made sense to me to open the file you connect to db from

### DIFF
--- a/src/postgres/commands/connectPostgresDatabase.ts
+++ b/src/postgres/commands/connectPostgresDatabase.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Uri, window, workspace } from 'vscode';
+import { Uri, window } from 'vscode';
 import { AzureTreeItem, IActionContext } from "vscode-azureextensionui";
 import { ext } from "../../extensionVariables";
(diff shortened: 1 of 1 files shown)


src/postgres/commands/connectPostgresDatabase.ts
```diff
@@ -3,13 +3,18 @@
...
 import { ext } from "../../extensionVariables";
 import { PostgresDatabaseTreeItem } from "../tree/PostgresDatabaseTreeItem";
 import { connectedPostgresKey } from "./registerPostgresCommands";
 
-export async function connectPostgresDatabase(context: IActionContext, treeItem?: PostgresDatabaseTreeItem): Promise<void> {
-    if (!treeItem) {
+export async function connectPostgresDatabase(context: IActionContext, treeItem?: Uri | PostgresDatabaseTreeItem): Promise<void> {
+    if (!treeItem || treeItem instanceof Uri) {
+        if (treeItem) {
+            window.showTextDocument(await workspace.openTextDocument(treeItem));
```
neelip, line 15: nit: found this showTextDocument version. Can we use just the Uri instead of await on openTextDocument?
`export function showTextDocument(uri: Uri, options?: TextDocumentShowOptions): Thenable<TextEditor>;`

wwlorey: Nice find 👍

Issue #1455, opened 2020-05-08
Make sure "Execute Query" codelens isn't annoying


EricJizbaMSFT:
In https://github.com/microsoft/vscode-cosmosdb/pull/1444, we added an "Execute Query" codelens to all "sql" files, but I think there's a risk of that codelens showing up in sql files unrelated to our PostgreSQL database support and annoying/confusing users. We might want to investigate having our own unique file type that "extends" sql so that _only_ the right people get the codelens but we maintain existing support for sql colorization (I'm _pretty sure_ that's possible, but not 100%)

wwlorey:
So I found that it is possible to have sql syntax coloring for other file extensions using this setting:

```json
"files.associations": {
    "*.psql": "sql"
}
```
<img width="263" alt="Screen Shot 2020-05-15 at 1 04 36 PM" src="https://user-images.githubusercontent.com/22795803/82091676-a2514d80-96ac-11ea-8ec8-7bdb83502a55.png">

But unfortunately there's some weird behavior with the codelens. Because this aliases `psql` files to `sql`, the codelens doesn't apply once this setting is set. I'll keep looking for another way to take care of this

PR #1468, opened 2020-05-15
Prevent codelens from appearing in all '.sql' files

wwlorey:
This changes the default query file name to "pg-query" and applies the codelens only to files with names that contain "pg-query" and end with ".sql" 
("pg-query" seemed like a more Postgres specific option than just "query")

I also changed the file name format for function queries opened via the tree view. It looked too busy to show the server name, schema, and "pg-query"

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1455

### DIFF
--- a/src/postgres/commands/registerPostgresCommands.ts
+++ b/src/postgres/commands/registerPostgresCommands.ts
@@ -5,7 +5,7 @@
 
 import { languages } from "vscode";
 import { callWithTelemetryAndErrorHandling, IActionContext, registerCommand } from "vscode-azureextensionui";
-import { doubleClickDebounceDelay } from "../../constants";
+import { connectedPostgresKey, doubleClickDebounceDelay, postgresBaseFileName, sqlFileExtension } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { PostgresCodeLensProvider } from "../services/PostgresCodeLensProvider";
 import { PostgresDatabaseTreeItem } from "../tree/PostgresDatabaseTreeItem";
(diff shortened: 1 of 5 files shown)


src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -23,11 +23,12 @@ import { executePostgresQuery } from "./executePostgresQuery";
 import { openPostgresFunction } from "./openPostgresFunction";
 
 export const connectedPostgresKey: string = 'ms-azuretools.vscode-azuredatabases.connectedPostgresDB';
-const postgresLanguageId: string = 'sql';
+export const postgresFileExtension: string = '.sql';
+export const postgresBaseFileName: string = 'pg-query';
```
EricJizbaMSFT, line 27: This stuff should go in `constants.ts`

src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -23,11 +23,12 @@ import { executePostgresQuery } from "./executePostgresQuery";
 import { openPostgresFunction } from "./openPostgresFunction";
 
 export const connectedPostgresKey: string = 'ms-azuretools.vscode-azuredatabases.connectedPostgresDB';
-const postgresLanguageId: string = 'sql';
+export const postgresFileExtension: string = '.sql';
```
EricJizbaMSFT, line 26: nit: `sqlFileExtension`

Makes it more clear to me that `postgresBaseFileName` is what you actually use to identify this as postgres instead of just sql

PR #1469, opened 2020-05-15
copyConnectionString command for Postgres Database

neelip:
Format: "postgres://user:password@host:port/dbName"

### DIFF
--- a/src/postgres/commands/copyConnectionString.ts
+++ b/src/postgres/commands/copyConnectionString.ts
@@ -1,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ClientConfig } from 'pg';
+import * as vscode from 'vscode';
+import { IActionContext, IParsedError, parseError } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { firewallNotConfiguredErrorType, invalidCredentialsErrorType, PostgresDatabaseTreeItem } from '../tree/PostgresDatabaseTreeItem';
+import { configurePostgresFirewall } from './configurePostgresFirewall';
+import { enterPostgresCredentials } from './enterPostgresCredentials';
+
+export async function copyConnectionString(context: IActionContext, node: PostgresDatabaseTreeItem): Promise<void> {
+    const message = 'The connection string has been copied to the clipboard';
+    if (!node) {
+        node = <PostgresDatabaseTreeItem>await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+    }
+
+    let clientConfig: ClientConfig | undefined;
+    while (!clientConfig) {
+        try {
+            clientConfig = await node.getClientConfig();
+        } catch (error) {
+            const parsedError: IParsedError = parseError(error);
+
+            if (parsedError.errorType === invalidCredentialsErrorType) {
+                await enterPostgresCredentials(context, node.parent);
+            } else if (parsedError.errorType === firewallNotConfiguredErrorType) {
+                await configurePostgresFirewall(context, node.parent);
+            } else {
+                throw error;
+            }
+        }
+    }
+
+    await vscode.env.clipboard.writeText(node.connectionString);
+    vscode.window.showInformationMessage(message);
+}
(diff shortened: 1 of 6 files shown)


src/postgres/commands/copyConnectionString.ts
```diff
@@ -0,0 +1,39 @@
...
+import { enterPostgresCredentials } from './enterPostgresCredentials';
+
+export async function copyConnectionString(context: IActionContext, node: PostgresDatabaseTreeItem): Promise<void> {
+    const message = 'The connection string has been copied to the clipboard';
+    if (!node) {
+        node = <PostgresDatabaseTreeItem>await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+    }
+
+    let clientConfig: ClientConfig | undefined;
+    while (!clientConfig) {
```
EricJizbaMSFT, line 21: You copied this code from a different file. Please put it in a place where the code can be shared

src/postgres/commands/copyConnectionString.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ClientConfig } from 'pg';
+import * as vscode from 'vscode';
+import { IActionContext, IParsedError, parseError } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { firewallNotConfiguredErrorType, invalidCredentialsErrorType, PostgresDatabaseTreeItem } from '../tree/PostgresDatabaseTreeItem';
+import { configurePostgresFirewall } from './configurePostgresFirewall';
+import { enterPostgresCredentials } from './enterPostgresCredentials';
+
+export async function copyConnectionString(context: IActionContext, node: PostgresDatabaseTreeItem): Promise<void> {
+    const message = 'The connection string has been copied to the clipboard';
```
EricJizbaMSFT, line 15: nit: Use localize and move this message down by where it's used

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -23,6 +23,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
     public static contextValue: string = "postgresDatabase";
     public readonly contextValue: string = PostgresDatabaseTreeItem.contextValue;
     public readonly childTypeLabel: string = "Resource Type";
+    public connectionString: string;
```
EricJizbaMSFT, line 26: The type `string` here isn't accurate, because this will be undefined until people call `getClientConfig`. It's also not clear that `getClientConfig` would populate the connection string. One possible fix is to remove `connectionString` from this class and have `copyConnectionString.ts` use the clientConfig to construct the connection string in it's own file

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -113,6 +114,11 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
             const host: string = nonNullProp(this.parent.server, 'fullyQualifiedDomainName');
             const clientConfig: ClientConfig = { user: username, password, ssl, host, port: 5432, database: this.databaseName };
 
+            // Create connection string
+            const credentials = username + ':' + password;
+            const serverDetails = host + ':5432';
```
EricJizbaMSFT, line 119: Don't hard-code 5432 multiple times. Reference `clientConfig.port` instead

src/postgres/commands/copyConnectionString.ts
```diff
@@ -0,0 +1,28 @@
...
+export async function copyConnectionString(context: IActionContext, node: PostgresDatabaseTreeItem): Promise<void> {
+    if (!node) {
+        node = <PostgresDatabaseTreeItem>await ext.tree.showTreeItemPicker(PostgresDatabaseTreeItem.contextValue, context);
+    }
+
+    const clientConfig: ClientConfig = await checkAuthentication(context, node);
+
+    const credentials = nonNullProp(clientConfig, 'user') + ':' + nonNullProp(clientConfig, 'password');
+    const serverDetails = nonNullProp(clientConfig, 'host') + ':' + String(nonNullProp(clientConfig, 'port'));
+    const connectionString = "postgres://" + credentials + '@' + serverDetails + '/\'' + node.databaseName + '\'';
```
EricJizbaMSFT, line 24: I prefer string template literals instead of adding a bunch of string together. Feels more readable to me. What do you think of this:

```typescript
const user: string = nonNullProp(clientConfig, 'user');
const password: string = nonNullProp(clientConfig, 'user');
const host: string = nonNullProp(clientConfig, 'host');
const port: string = String(nonNullProp(clientConfig, 'port'));
const connectionString = `postgres://${user}:${password}@${host}:${port}/'${node.databaseName}'`;
```

neelip: I agree, this is definitely better

Issue #1434, opened 2020-04-23
Fail to delete a SQL document


v-xinda:
**OS:** Mac
**Build Version:** [20200422.7](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=18192&view=artifacts&type=publishedArtifacts)

**Repro Steps:**
1. Create a SQL account.
2. Create a Collection with a partition key.
3. Create a document and leave blank for partition key.
4. Delete the document.

**Expect:**
Delete the document successfully without any error.

**Actual:**
Fail to delete the document and pop up an error.
![image](https://user-images.githubusercontent.com/57662131/80090718-83bad380-8592-11ea-87b0-0e8a27b4117b.png)
 Details in output:
	
2:57:23 PM: Error: Entity with the specified id does not exist in the system., 
	RequestStartTime: 2020-04-23T06:57:23.1250635Z, RequestEndTime: 2020-04-23T06:57:23.1250635Z,  Number of regions attempted:1
	ResponseTime: 2020-04-23T06:57:23.1250635Z, StoreResult: StorePhysicalAddress: rntbd://bluprdapp02-docdb-1.documents.azure.com:14471/apps/b3102e89-1a41-41aa-b073-4499e437059e/services/f29eff1a-a796-4f9f-aa41-7f1b28cdd08d/partitions/7fa2dfbb-71bc-4cd5-a2ed-1cae9d50deef/replicas/132315858713877073p/, LSN: 9, GlobalCommittedLsn: 9, PartitionKeyRangeId: 0, IsValid: True, StatusCode: 404, SubStatusCode: 0, RequestCharge: 1.24, ItemLSN: -1, SessionToken: 9, UsingLocalLSN: False, TransportException: null, ResourceType: Document, OperationType: Delete
	, Microsoft.Azure.Documents.Common/2.10.0
```

neelip:
We are able to repro this issue, we will get to fixing it in the next update

PR #1483, opened 2020-05-20
Use empty string for emptyPartitionKeyValue

wwlorey:
Using an empty object as the partition key when one wasn't provided seemed to be causing the issue.

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1434

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -73,8 +73,6 @@ export const defaultStoredProcedure =
     if (!isAccepted) throw new Error('The query was not accepted by the server.');
 };` ;
 
-export const emptyPartitionKeyValue: string = '';
-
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -73,7 +73,7 @@ export const defaultStoredProcedure =
     if (!isAccepted) throw new Error('The query was not accepted by the server.');
 };` ;
 
-export const emptyPartitionKeyValue = {};
+export const emptyPartitionKeyValue: string = '';
```
EricJizbaMSFT, line 76: Seemed excessive to have a constant for this in the first place, but seems even more excessive if it's just an empty string

wwlorey:
> Can you go through that discussion and make extra sure this change won't cause any regressions?

I went through the [test cases](https://github.com/microsoft/vscode-cosmosdb/pull/597#discussion_r185594896) Prashanth had tried and the portal errors out now for a partition key of `{}` with no quotes. Also, a partition key of `undefined`, which was also discussed in that thread, isn't allowed on the portal. And using `undefined` in the extension for the empty partition key also causes an error.

So an empty string for `emptyPartitionKeyValue` looks like the best option

Issue #866, opened 2018-10-12
Give better error message when local mongo server not running


StephenWeatherford:
failed to connect to server [127.0.0.1:27017] on first connect [MongoError: connect ECONNREFUSED 127.0.0.1:27017]

PR #1492, opened 2020-05-22
Add better error when local Mongo server not running

wwlorey:
I repurposed the existing better error message because it was applicable. But I'm not a big fan of having a non-clickable link in the tree item. Should I remove the link? Or maybe throw up a notification instead?

Here's what it looks like now:
Screen Shot 2020-05-21 at 4 57 48 PM

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/866

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -85,8 +85,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
             await node.deleteTreeItem(actionContext);
         });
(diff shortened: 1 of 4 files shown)


nturinski:
Wait one sec

nturinski:
Uhh... if you were waiting on me, I'm good with this now 👍

wwlorey:
Ok thanks @nturinski 👍 Eric & I decided to add one more thing so I'm working on that rn

Issue #1343, opened 2020-02-29
Error message is unclear : Retryable writes are not supported


softchris:
Repro steps:

1. Create a MongoDB Database by right clicking on CosmosDB acount
2. Shows `Retryable writes are not supported`

Comment: I don't understand why I have to do this. Will the extension not work properly if I don't? 
Why can't this be done for me?
From a UX perspective, it looks weird that I get an error the first thing that happens after creating a Database, makes me think something is broken in general

Action: cosmosDB.createMongoDatabase
Error type: 2
Error Message: Retryable writes are not supported. Please disable retryable writes by specifying "retrywrites=false" in the connection string or an equivalent driver specific config.

Version: 0.12.1
OS: darwin
Product: Visual Studio Code
Product Version: 1.42.1
Language: en

<details>
<summary>Call Stack</summary>

```
Connection.<anonymous> mongodb/pool.js:466:61
Connection.emit events.js:200:13
processMessage mongodb/connection.js:364:10
TLSSocket.<anonymous> mongodb/connection.js:533:15
TLSSocket.emit events.js:200:13
addChunk _stream_readable.js:294:12
readableAddChunk _stream_readable.js:275:11
TLSSocket.Readable.push _stream_readable.js:210:10
TLSWrap.onStreamRead stream_base_commons.js:166:17
```

</details>

DarkLite1:
I received the same error message today from the extension for our Azure Cosmos MongoDB. For us it happened after right clicking the database and selecting `Create new collection`. The thing is the error popped up and it seemed like the collection was not created. After restart it appears the collection was created none the less. Really strange bug....

DarkLite1:
The same error is thrown when trying to create a new document:

> Connection to the database is successful
> (node:14120) UnhandledPromiseRejectionWarning: MongoError: Retryable writes are not supported. Please disable retryable writes by specifying "retrywrites=false" in the connection string or an equivalent driver specific config.
>     at MessageStream.messageHandler (T:\Test\Brecht\Node\acadamind\mean-course\node_modules\mongodb\lib\cmap\connection.js:261:20)
>     at MessageStream.emit (events.js:321:20)
>     at processIncomingData (T:\Test\Brecht\Node\acadamind\mean-course\node_modules\mongodb\lib\cmap\message_stream.js:144:12)
[...]

PR #1493, opened 2020-05-22
Add query param retrywrite=false

nturinski:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1343

Technically this issue only reproduces with Azure Cosmos MongoDB v3.6, which our extension does not create by default (it creates v3.2, we should probably look into that 😂)

But I didn't notice any adverse effects appending the query param to older MongoDB accounts, so I assume that this is easier solution because I didn't see any obvious indicators if the account is 3.2 or 3.6.  (CosmosAccountType: Non-Production for 3.6, but that property didn't exist on the 3.2 accounts)

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -109,7 +109,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             // for any Mongo connectionString, append this query param because the Cosmos Mongo API v3.6 doesn't support retrywrites
             // but the newer node.js drivers started breaking this
             const searchParam: string = 'retrywrites';
-            if (!connectionString.searchParams.get(searchParam)) {
+            if (!connectionString.searchParams.has(searchParam)) {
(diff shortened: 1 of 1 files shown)


nturinski:
I guess I should mention that connectionStrings look like this
`mongodb://naturins-mongo:oaOySZTlAfWTwy4Bg2aotQD9qzhp5KTR8vOXq7xlzDyXLOCBrQlJJkT0Tk7IKbOr28QHQhvIOVvYSfHNBvHd9Q==@naturins-mongo.documents.azure.com:10255/?ssl=true&replicaSet=globaldb` which is why I can just append the query param.  But if we want to double check and validate the strings, I can do that.

EricJizbaMSFT:
On one hand I'm less concerned because this connection string is always coming from Azure (aka it's not the attached account scenario where it's the wild wild west and the user can enter anything), but I'm still a little concerned they could change the format of the connection string in the future.

I'm also a little lost on if/when retryablewrites will every be supported. Don't want to set it to false if it's actually supported

Issue #1504, opened 2020-05-22
Fail to create and delete databases at the same time


v-mengwe:
**OS:** Mac/Linux
**Build Version:** 20200522.22

**Repro step:**

1. Delete a database under a PostgreSQL server
2. Create a new database with a different name while deleting a database
3. Check the result 

**Expect:**
Succeed to delete and create databases without any error.

**Actual:**
Pop up a prompt "long running operation failed with error: '55006: databases "data1" is being accessed by other users' ".
![image](https://user-images.githubusercontent.com/62053665/82653726-39d91200-9c52-11ea-8039-5928aaf4c344.png)

**More info:**
It is unstable to reproduce this issue in Windows.

PR #1526, opened 2020-05-26
Cleanup Postgres connections

wwlorey:
Not ending connections once we were done with them was causing access issues. 

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1504

### DIFF
--- a/src/postgres/commands/executePostgresQuery.ts
+++ b/src/postgres/commands/executePostgresQuery.ts
@@ -37,6 +37,7 @@ export async function executePostgresQuery(context: IActionContext): Promise<voi
     await client.connect();
     const query: string | undefined = activeEditor.document.getText();
     const queryResult: QueryResult = await client.query(query);
+    await client.end();
     ext.outputChannel.appendLine(localize('executedQuery', 'Successfully executed "{0}" query.', queryResult.command));
 
     if (queryResult.rowCount) {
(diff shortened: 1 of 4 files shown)


src/postgres/commands/executePostgresQuery.ts
```diff
@@ -37,6 +37,7 @@ export async function executePostgresQuery(context: IActionContext): Promise<voi
     await client.connect();
     const query: string | undefined = activeEditor.document.getText();
     const queryResult: QueryResult = await client.query(query);
+    await client.end();
```
EricJizbaMSFT, line 40: you should use try/finally for all of these

src/postgres/commands/executePostgresQuery.ts
```diff
@@ -34,27 +34,31 @@ export async function executePostgresQuery(context: IActionContext): Promise<voi
...
+                }
+                csvData += `${fieldValues.join(',')}${EOL}`;
             }
-            csvData += `${fieldValues.join(',')}${EOL}`;
-        }
 
-        await vscodeUtil.showNewFile(csvData, outputFileName, '.csv');
+            await vscodeUtil.showNewFile(csvData, outputFileName, '.csv');
+        }
+    } finally {
```
EricJizbaMSFT, line 61: It's always a good idea to minimize what you put in a `try` block. In this case, it means the `client.end()` will happen sooner. If you're using a `catch` block, it means you won't be catching irrelevant errors

Can you shuffle the order of the lines so that only the `client.connect` and `client.query` calls are wrapped

Issue #1497, opened 2020-05-22
It takes about 1 min to pop up the confirmation dialog when configuring firewall


v-ruizh:
**OS:** All
**Build Version:** 20200522.22

**Repro Steps:**
1. Create a PostgreSQL server and skip the step "Add firewall rule...".
2. Drill down this PostgreSQL->Database postgres.
3. Click "Configure firewall to connect to...".
![image](https://user-images.githubusercontent.com/45614674/82658705-187c2400-9c5a-11ea-8ad1-ea2528cb1b5f.png)

**Expect:**
The confirmation dialog pops up in several seconds.
![image](https://user-images.githubusercontent.com/45614674/82658693-10bc7f80-9c5a-11ea-9faf-13b7aea5461c.png)

**Actual:**
It takes about 1 min to pop up the confirmation dialog.

PR #1529, opened 2020-05-26
Add progress notification when getting public IP

wwlorey:
Getting the public IP sometimes takes forever so I added a progress notification.

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1497


src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -19,15 +19,22 @@ export async function configurePostgresFirewall(context: IActionContext, treeIte
...
     }
 
-    const ip: string = await publicIp.v4();
+    const options: vscode.ProgressOptions = {
+        location: vscode.ProgressLocation.Notification,
+        title: localize('gettingPublicIp', 'Getting public IP...')
+    };
+
+    const ip: string = await vscode.window.withProgress(options, async () => {
+        return await publicIp.v4();
```
EricJizbaMSFT, line 28: Can you check for other cases where we call this? I feel like there's at least one more

Issue #1495, opened 2020-05-22
There are 4 errors for the created "Function Query"


v-ruizh:
**OS:** All
**Build Version:** 20200522.22

**Repro Steps:**
1. Drill down PostgreSQL->One database->Right click Functions node and select "Create Function Query"->Provide a function name->Select return type.
2. Install the SQL Server extension when prompt to install it.
3. Check the "PROBLEMS"

**Expect:**
There is no error in the "PROBLEMS".

**Actual:**
There are 4 errors in the "PROBLEMS".
![image](https://user-images.githubusercontent.com/45614674/82656663-b79f1c80-9c56-11ea-85e6-c855cc79a978.png)

**More info:**
Execute query successfully with these errors.
![image](https://user-images.githubusercontent.com/45614674/82656908-15cbff80-9c57-11ea-884a-650bac394518.png)

wwlorey:
For now, let's use the `psql` file extension and let people download a separate extension for coloring support

PR #1533, opened 2020-05-26
Use 'psql' file extension for Postgres

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1495

### DIFF
--- a/package.json
+++ b/package.json
@@ -443,7 +443,7 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
-                    "when": "resourceLangId==sql"
+                    "when": "resourceLangId==psql"
                 }
             ],
             "editor/title": [
@@ -815,7 +815,7 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
-                    "when": "resourceLangId == sql"
+                    "when": "resourceLangId == psql"
                 }
             ],
             "commandPalette": [
@@ -845,7 +845,7 @@
                 },
                 {
                     "command": "postgreSQL.executeQuery",
-                    "when": "editorLangId == 'sql'"
+                    "when": "editorLangId == 'psql'"
                 }
             ]
         },
--- a/package.json
+++ b/package.json
@@ -118,6 +118,15 @@
                 ],
                 "firstLine": "^#!/.*\\bmongo*\\b",
                 "configuration": "./grammar/configuration.json"
+            },
+            {
+                "id": "postgres",
+                "aliases": [
+                    "PostgreSQL"
+                ],
+                "extensions": [
+                    ".psql"
+                ]
             }
         ],
         "grammars": [
@@ -443,7 +452,7 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
-                    "when": "resourceLangId==psql"
+                    "when": "resourceLangId==postgres"
                 }
             ],
             "editor/title": [
@@ -815,7 +824,7 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
-                    "when": "resourceLangId == psql"
+                    "when": "resourceLangId == postgres"
                 }
             ],
             "commandPalette": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -815,7 +815,7 @@
                 },
                 {
                     "command": "postgreSQL.connectDatabase",
-                    "when": "resourceLangId == sql"
+                    "when": "resourceLangId == psql"
```
EricJizbaMSFT, line 818: Any idea what the difference between resourceLangId and editorLangId is?

src/postgres/commands/registerPostgresCommands.ts
```diff
@@ -24,7 +24,7 @@ import { openPostgresFunction } from "./openPostgresFunction";
 
 export function registerPostgresCommands(): void {
     ext.postgresCodeLensProvider = new PostgresCodeLensProvider();
-    ext.context.subscriptions.push(languages.registerCodeLensProvider({ pattern: `{**/,}*${postgresBaseFileName}*${sqlFileExtension}` }, ext.postgresCodeLensProvider));
+    ext.context.subscriptions.push(languages.registerCodeLensProvider({ pattern: `{**/,}*${postgresBaseFileName}*${postgresFileExtension}` }, ext.postgresCodeLensProvider));
```
EricJizbaMSFT, line 27: I feel like it would be fine to use the old logic based solely on the file extension, since it's now unique to postgres

wwlorey:
So I found a kinda fun issue here. I forgot to define the `psql` language id when I created this PR, so as written commands like [this](https://github.com/microsoft/vscode-cosmosdb/pull/1533/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R445) won't show up where they're expected to.

I defined the postgres language id in our package.json but it broke the [existing postgres extension](https://marketplace.visualstudio.com/items?itemName=ckolkman.vscode-postgres) we talked about in standup:

Screen Shot 2020-05-26 at 4 25 06 PM

And there isn't syntax coloring even with that other postgres extension installed after that error was thrown.

A possible fix for this would be to not define the postgres language in our extension and get rid of the context menu & command palette entry points that rely on the `psql` language id, just so we avoid breaking another extension. Does anyone know of a better solution?

neelip:
> So I found a kinda fun issue here. I forgot to define the `psql` language id when I created this PR, so as written commands like [this](https://github.com/microsoft/vscode-cosmosdb/pull/1533/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R445) won't show up where they're expected to.
> 
> I defined the postgres language id in our package.json but it broke the [existing postgres extension](https://marketplace.visualstudio.com/items?itemName=ckolkman.vscode-postgres) we talked about in standup:
> 
> Screen Shot 2020-05-26 at 4 25 06 PM
> 
> And there isn't syntax coloring even with that other postgres extension installed after that error was thrown.
> 
> A possible fix for this would be to not define the postgres language in our extension and get rid of the context menu & command palette entry points that rely on the `psql` language id, just so we avoid breaking another extension. Does anyone know of a better solution?

I was getting the same error before. I had to downgrade the version of PostgreSQL extension to 0.3.0 to get rid of this problem. I think its a problem with specific version of the PostgreSQL extension.

wwlorey:
@neelip was that fix for [Microsoft's postgres extension](https://marketplace.visualstudio.com/items?itemName=ms-ossdata.vscode-postgresql)? I didn't see a version 0.3.0 for [Chris Kolkman's extension](https://marketplace.visualstudio.com/items?itemName=ckolkman.vscode-postgres).

I do think I have a workaround for this though. If we set our postgres language ID to the same used in the other postgres extension, we don't break the syntax coloring. The "Couldn't start client..." error still pops up but would happen sometimes even when we didn't define a postgres language ID.

package.json
wwlorey: I'm struggling to find docs on this. But as far as I can tell `resourceLangId` is used for context menus and `editorLangId` is used for command palette. I copied those properties from the existing mongo commands

Issue #1502, opened 2020-05-22
[Suggestion] Add the log info when deleting a database


v-ruizh:
**OS:** All
**Build Version:** 20200522.22

**Repro Steps:**
Delete a database from any account.

**Actual:**
There is no log info when deleting a database.

**Suggestion:**
Add the log info when deleting a database.

**More info:**
1. This issue also reproduces when deleting collection, document....
2. There is notification message when deleting an account but no log in output.

PR #1534, opened 2020-05-27
Success notifications and log output for postgres operations

neelip:
Relates to #1502

### DIFF
--- a/src/postgres/commands/createPostgresDatabase.ts
+++ b/src/postgres/commands/createPostgresDatabase.ts
@@ -14,8 +14,8 @@ export async function createPostgresDatabase(context: IActionContext, node?: Pos
     if (!node) {
         node = <PostgresServerTreeItem>await ext.tree.showTreeItemPicker(PostgresServerTreeItem.contextValue, context);
     }
-    const newDatabase = <PostgresDatabaseTreeItem>await node.createChild(context);
-    const createMessage = localize('createPostgresDatabaseMsg', 'Successfully created database "{0}".', newDatabase.databaseName);
+    const newDatabase: PostgresDatabaseTreeItem = await node.createChild(context);
+    const createMessage: string = localize('createPostgresDatabaseMsg', 'Successfully created database "{0}".', newDatabase.databaseName);
     vscode.window.showInformationMessage(createMessage);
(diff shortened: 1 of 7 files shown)


src/postgres/commands/createPostgresDatabase.ts
```diff
@@ -3,13 +3,19 @@
...
+import { localize } from '../../utils/localize';
+import { PostgresDatabaseTreeItem } from '../tree/PostgresDatabaseTreeItem';
 import { PostgresServerTreeItem } from '../tree/PostgresServerTreeItem';
 
 export async function createPostgresDatabase(context: IActionContext, node?: PostgresServerTreeItem): Promise<void> {
     if (!node) {
         node = <PostgresServerTreeItem>await ext.tree.showTreeItemPicker(PostgresServerTreeItem.contextValue, context);
     }
-    await node.createChild(context);
+    const newDatabase = <PostgresDatabaseTreeItem>await node.createChild(context);
```
wwlorey, line 17: nit: I'm not sure if the tslint rule requiring types will be turned on soon but it's nice to have types on variables

src/postgres/commands/deletePostgresDatabase.ts
```diff
@@ -17,4 +18,7 @@ export async function deletePostgresDatabase(context: IActionContext, node?: Pos
     if (result === DialogResponses.deleteResponse) {
         await node.deleteTreeItem(context);
     }
+    const deleteMessage = localize('deletePostgresDatabaseMsg', 'Successfully deleted database "{0}".', node.databaseName);
```
wwlorey, line 21: Would it be better to put this logging in `deleteTreeItemImpl`? @EricJizbaMSFT may have thoughts on this

EricJizbaMSFT: I actually prefer it here for two reasons:
1. Since the tree item files are shared in so many places, I'd like to keep as much code out of there as possible
1. It's better to have the confirmation dialog outside of `deleteTreeItemImpl` so that we don't get the "Deleting..." thing in the tree while we wait for the user to confirm deletion. And it feels consistent to put the completed notification here if that's where the confirmation dialog is

PR #1538, opened 2020-05-27
Prep for Release 0.13.0

neelip:
- package.json and package-lock.json updates (version and name)
- updates to CHANGELOG.md

PR #1545, opened 2020-05-28
Update PostreSQL icons

EricJizbaMSFT:
Okay the PostgreSQL icons _reallllllyy_ bothered me during testing. I know these aren't perfect, but at least they make _some_ sense in my mind instead of being completely inaccurate like the old ones. Can we please include this in the release? 😄 Since we're waiting for the "Prep for release" [PR](https://github.com/microsoft/vscode-cosmosdb/pull/1538) anyways

||Before|After|
|---|---|---|
|Dark|Screen Shot 2020-05-28 at 11 31 06 AM|Screen Shot 2020-05-28 at 11 29 48 AM|
|Light|Screen Shot 2020-05-28 at 11 31 19 AM|Screen Shot 2020-05-28 at 11 29 59 AM|


jinglouMSFT:
The icon for 'abc' in dark theme isn't accessible. It's toooo dark.

EricJizbaMSFT:
@jinglouMSFT I already said "I know these aren't perfect"

jinglouMSFT:
@EricJizbaMSFT  the old set of the icons was just ugly, but no issues. The new set introduced an accessibility issue.  And I believe there are a lot dark theme users out there.  Can the 'f' icon/svg be configured to be theme sensitive?

EricJizbaMSFT:
They're not ugly - we use these icons elsewhere and I'm totally fine with it. The problem is that we were using an icon for a "document collection" when it's actually a "function". That _is_ an issue because it's an icon for the wrong resource type

> Can the 'f' icon/svg be configured to be theme sensitive?

Yes, but I'll let the actual icon designers do that for the next release. I originally got this icon from the portal and we use it in the Functions extension

PR #1546, opened 2020-05-28
Update tests for 'azureDatabases.createServer' (not including Postgres)

wwlorey:
Forgot to update the nightly tests when we merged 'cosmosDB.createAccount' into 'azureDatabases.createServer'


wwlorey:
Windows always hits the timeout on those nightly tests... I'm gonna queue another normal build so this can be merged

PR #1547, opened 2020-05-28
Update function icon

EricJizbaMSFT:
I still think the other color is shippable - we've been using it for years in the Functions extension (See [here](https://raw.githubusercontent.com/microsoft/vscode-azurefunctions/master/resources/readme/debug2.png)) and we originally got it from the portal. But the boss is probably writing my Connect so don't want to make him too mad 😬

Screen Shot 2020-05-28 at 3 24 55 PM
Screen Shot 2020-05-28 at 3 25 06 PM

Up to Will and Neelima if they actually want this change


neelip:
I like the individual function and table icons but the functions and tables are not that great. However, much better than the last set of icons.

jinglouMSFT:
Smart move, @EricJizbaMSFT!  Now go fix your Function extension :-)

PR #1548, opened 2020-05-28
Fix CHANGELOG formatting in VS Code

wwlorey:
Markdown formatting in VS Code was funky (peep the "Fixed" subheading). Shoutout to @neelip for finding this problem

Old -> New
MicrosoftTeams-imageScreen Shot 2020-05-28 at 3 35 58 PM


nturinski:
Nice catch, that _is_ funky! 🕺

Issue #934, opened 2018-11-09
Unrecognized child type


PrashanthCorp:
input: 
`db.coll.find({name: /"\\\\b"})`

Error: Unrecognized node type encountered. Could not parse / as part of /<missing '('>"\\\\b"})

ANOTHER REPRO:
db.ingestion_queue.find( { "date_time": {"$gte": ISODate2("2018-08-09T00:00:00.000Z")} } )

Error Message: Error near line 0, column 45: 'Unrecognized node type encountered. Could not parse ISODate2 as part of ISODate2("2018-08-09T00:00:00.000Z")'. Please check syntax.

StephenWeatherford:
We just need a friendlier message like "Unrecognized function name" or even "Unable to parse 'ISODate2', or "Unable to parse syntax" or whatever, whatever's easiest, just  so it doesn't sound like an internal error.

PR #1556, opened 2020-06-03
Update error when executing mongo command

wwlorey:
I'm a little hesitant to remove the "Unrecognized node type encountered" part because it's part of the error message and does accurately describe what the problem is. I shuffled around the error message to hopefully make it more readable.

| old | new |
| - | - |
| Screen Shot 2020-06-03 at 11 00 30 AM | Screen Shot 2020-06-03 at 10 54 24 AM |

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/934


neelip:
Is there a way to emphasize "Unable to parse syntax" as the error title/label?

wwlorey:
As far as I know we can only provide an error message :( And md styling isn't supported

neelip:
No worries. This changes makes sense tho, makes it easier to understand the error.

Issue #647, opened 2018-05-30
Incorrect line reported for syntax error if "Execute" CodeLens is used


StephenWeatherford:
1) Put this in a scrapbook:

db.createUser({
                "user": "buddhi",
                "pwd": "123",
                "roles": ()
                }
            )
       
2) Put cursor anywhere in the command and right-click -> "Execute MongoDB Command"
3) As expected, you get this error:

Error near line 3, column 25, text '('. Please check syntax.

4) Now click "Execute" at the top of the file
!) The line and column numbers are reported incorrectly:

Error near line 0, column 51, text '('. Please check syntax.

StephenWeatherford:
Still repro?

StephenWeatherford:
Current behavior:  
1) problems pane:  Behavior is correct
2) Execute Mongo Command: The line number is off by one e.g. if the bad line is line 8, the error message says line 7:
  
<img width="196" alt="image" src="https://user-images.githubusercontent.com/6913354/46114079-f8a30d00-c1a5-11e8-96d7-258414b07b10.png">
<img width="420" alt="image" src="https://user-images.githubusercontent.com/6913354/46114095-022c7500-c1a6-11e8-88d6-d09d4e3515ef.png">

3) Clicking Execute: You get completely incorrect line and column:
<img width="478" alt="image" src="https://user-images.githubusercontent.com/6913354/46114160-461f7a00-c1a6-11e8-8f32-dbfc95ab464d.png">

PrashanthCorp:
Regarding 3: The method `getAllCommandsFromText` removes line endings and spaces. We'll have to reconstruct the exact line and character indices before pushing the position to codelens. 
2 : Verified there's an off-by-one error [here](https://github.com/Microsoft/vscode-cosmosdb/blob/a6e8aa43a91b1600f9e5c140df1e6e222b0635d1/src/mongo/MongoScrapbook.ts#L106)

PR #1557, opened 2020-06-03
Only use 'executeCommandFromActiveEditor' when executing mongo commands

wwlorey:
Calling `executeCommandFromText` wouldn't preserve any whitespace so the line and column info for errors was wrong. And the `executeMongoCommand` rules in package.json should guarantee that there's always an active text editor when the command is called.

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/647

This also fixes the off by one error noted in the same issue: https://github.com/microsoft/vscode-cosmosdb/issues/647#issuecomment-424898870

### DIFF
--- a/src/mongo/services/MongoCodeLensProvider.ts
+++ b/src/mongo/services/MongoCodeLensProvider.ts
@@ -61,7 +61,8 @@ export class MongoCodeLensProvider implements vscode.CodeLensProvider {
                     lenses.push(<vscode.CodeLens>{
                         command: {
                             title: "Execute",
-                            command: 'cosmosDB.executeMongoCommand'
+                            command: 'cosmosDB.executeMongoCommand',
+                            arguments: [cmd.range.start]
                         },
                         range: cmd.range
                     });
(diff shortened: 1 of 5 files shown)


src/mongo/services/MongoCodeLensProvider.ts
```diff
@@ -61,8 +61,7 @@ export class MongoCodeLensProvider implements vscode.CodeLensProvider {
                     lenses.push(<vscode.CodeLens>{
                         command: {
                             title: "Execute",
-                            command: 'cosmosDB.executeMongoCommand',
-                            arguments: [cmd.text]
```
EricJizbaMSFT, line 65: `executeCommandFromActiveEditor` decides which command to run based on the cursor location, but that doesn't always correspond to the command being passed here in the codelens. You need to pass along _something_ here so that you know which command to execute. Can you pass along the position of the codelens? If so, that seems easy enough to do - it would replace the position used [here](https://github.com/microsoft/vscode-cosmosdb/blob/master/src/mongo/MongoScrapbook.ts#L52)

PR #1561, opened 2020-06-04
Move shared Postgres code to base class

wwlorey:
Setting up for stored procedures support

### DIFF
--- a/src/postgres/tree/PostgresResourcesTreeItemBase.ts
+++ b/src/postgres/tree/PostgresResourcesTreeItemBase.ts
@@ -7,6 +7,7 @@ import { ClientConfig } from "pg";
 import { AzureParentTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
 import { PostgresDatabaseTreeItem } from "./PostgresDatabaseTreeItem";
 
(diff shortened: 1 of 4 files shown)


src/postgres/tree/PostgresResourcesTreeItemBase.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ClientConfig } from "pg";
+import { AzureParentTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
+import { PostgresDatabaseTreeItem } from "./PostgresDatabaseTreeItem";
+
+export abstract class PostgresResourcesTreeItemBase extends AzureParentTreeItem<ISubscriptionContext> {
```
EricJizbaMSFT, line 10: `PostgresResourcesTreeItemBase` is a pretty generic name, but I don't have a better name for you. Maybe just add a comment in the file describing when and how this is meant to be used

PR #1562, opened 2020-06-05
Add tree view for Postgres 🐘 stored procedures ⚙️

wwlorey:
This only shows stored procs for Postgres versions 11+. And the duplicate name logic is the same as tables & functions.

Screen Shot 2020-06-05 at 12 20 07 PM


src/postgres/tree/PostgresStoredProceduresTreeItem.ts
```diff
@@ -0,0 +1,61 @@
...
+        return getThemedIconPath('list-unordered.svg');
+    }
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async loadMoreChildrenImpl(): Promise<PostgresStoredProcedureTreeItem[]> {
+        // Adapted from https://aka.ms/AA83fg8
+        const storedProceduresQuery: string = `select n.nspname as schema,
```
EricJizbaMSFT, line 36: Looks like only one or two lines of this query are changed from the functions query. Would prefer to share this instead of copying. I think it makes the most sense if it's a part of `getPostgresProcedureQueryRows`. Also, I would move the definition of `IPostgresProceduresQueryRow` into `getPostgresProcedureQueryRows.ts`, since the structure of the interface is completely dependent on how you authored this query. Just makes it easier to understand if you can see the query and the interface side-by-side.

PR #1568, opened 2020-06-11
Refactor postgres query wizard 🐘🧙‍♂️

wwlorey:
This just moves code around and creates a generic `IPostgresQueryWizardContext`. It should make it easier to implement more postgres query wizards. Defs going to use it for stored procedures and maybe tables.


src/postgres/commands/PostgresQueryWizards/FunctionQueryWizard/FunctionQueryCreateStep.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizardExecuteStep } from "vscode-azureextensionui";
```
EricJizbaMSFT, line 6: Okay let me explain how I would structure this, because it's different than what you've done. Here's [an example](https://github.com/microsoft/vscode-azurefunctions/tree/master/src/commands/createFunction) that explains my general philosophy:
- I prefer if immediate subfolders of the "commands" folder are named after the command itself. (And obviously the command would be in that folder - next to the wizard steps)
- I prefer not to use redundant words in folder names. I'm okay with it in function/class names because we see those without the context of the full path, but you never see a folder without the context of the folders around it
- I prefer to start folders with lowercase letters. Looks like all other folders in this repo start with a lowercase letter (except `PostgresAccountWizard` which I think you created haha)

All that being said, here's how I might structure this:
Screen Shot 2020-06-11 at 11 53 15 AM

I'm not saying you have to do it exactly like that, but hopefully you can apply some of my philosophy above. This case is a little weird because you have multiple commands sharing some of the same wizard. In my link above, it's one command using one wizard.

PR #1570, opened 2020-06-12
Fix mongo shell tests

wwlorey:
I'm not sure why these tests started to fail now... But to fix them I increased the mocha and mongo shell timeouts and made sure a directory that mongo needs was created.


src/vsCodeConfig/settings.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
```
EricJizbaMSFT, line 5: The whole "vsCodeConfig" file structure is kind of unique to the functions extension - partially because I have a lot more logic related to vs code settings. Also because I have an analagous "funcConfig" folder. I feel like this would be simpler as a "settingUtils.ts" file in the "utils" folder (In fact that's what I did [here](https://github.com/microsoft/vscode-cosmosdb/pull/1572/files#diff-1da0c03369229a5be1f98e829bf67baeR1) 😂)

test/mongoShell.test.ts
```diff
@@ -28,13 +29,18 @@ if (!isWindows) {
...
 
-suite("MongoShell", function (this: Mocha.Suite): void {
-    function testIfSupported(title: string, fn?: Mocha.Func | Mocha.AsyncFunc): void {
-        if (testsSupported) {
-            test(title, fn);
-        } else {
-            test(title);
-        }
+suite("MongoShell", async function (this: Mocha.Suite): Promise<void> {
+    // https://aka.ms/AA8o5r3
```
EricJizbaMSFT, line 33: Lol did you create an aka.ms link just for this? If so, pretty overkill for a comment in test code...

test/mongoShell.test.ts
```diff
@@ -51,6 +57,9 @@ suite("MongoShell", function (this: Mocha.Suite): void {
     } else if (!fse.existsSync(mongodPath)) {
         console.log(`Couldn't find mongo.exe at ${mongoPath} - skipping MongoShell tests`);
         testsSupported = false;
+    } else {
+        // Prevents code 100 error: https://aka.ms/AA8o5qm
+        await fse.ensureDir('D:\\data\\db\\');
```
EricJizbaMSFT, line 62: Weird to me that this is "D", but above the paths are "C"

test/mongoShell.test.ts
wwlorey: Lol my b, I got into the habit of doing that 😅

test/mongoShell.test.ts
wwlorey: Mongo looked for that path on the D drive for whatever reason. Since the general idea from the stack overflow link was right I included it

PR #1572, opened 2020-06-12
Replace editors with virtual file system ☁️📂

EricJizbaMSFT:
At least for this PR, I tried to keep the experience the same as much as possible (confirmation dialogs, file names, output, etc.). I noticed room for improvement in some of those areas, but I'd rather do that separately. This PR is solely meant to get our extensions onto the same editor experience from the UI package instead of each implementing their own (finally 😅).

The only thing that I _did_ change was how we handle fields like "_etag" for DocDB documents. Previously, we had to update the editor after a save to make sure it had the latest etag. That was a bit weird/complex, so I decided to hide those fields, handle them under the covers, and remove that extra "update" to the editor.
Before:
Screen Shot 2020-06-12 at 1 21 29 PM

After:
Screen Shot 2020-06-12 at 1 21 13 PM


src/mongo/MongoScrapbook.ts
```diff
@@ -105,14 +106,30 @@ async function executeCommand(context: IActionContext, command: MongoCommand): P
 
         // we don't handle chained commands so we can only handle "find" if isn't chained
         if (command.name === 'find' && !command.chained) {
-            await ext.editorManager.showDocument(context, new MongoFindResultEditor(database, command), 'cosmos-result.json', { showInNextColumn: true });
+            const db = await database.connectToDb();
+            const collectionName: string = nonNullProp(command, 'collection');
+            const collection: Collection = db.collection(collectionName);
+            // NOTE: Intentionally creating a _new_ tree item rather than searching for a cached node in the tree because
+            // the executed 'find' command could have a filter or projection that is not handled by a cached tree node
```
wwlorey, line 113: Would it be possible to update the existing collection tree item instead of creating a new one? Or is this easier?

EricJizbaMSFT: We _could_ update the existing collection tree item, but we don't _want_ to. The find command could have a filter or a projection in it's arguments, meaning the resulting documents could be completely different than what's in the tree. For example if the collection tree item has these documents:
```json
[
  {
    "_id": "apples",
    "qty": 5
  },
  {
    "_id": "avocados",
    "qty": "fourteen"
  },
  {
    "_id": "bananas",
    "qty": 26
  },
  {
    "_id": "oranges",
    "qty": {
      "in stock": 8,
      "ordered": 12
    }
  }
]
```

I can run a find command like this:
```
db.fruit.find( { qty: { $gt: 4 } } )
```

resulting in this:
```json
[
  {
    "_id": "apples",
    "qty": 5
  },
  {
    "_id": "bananas",
    "qty": 26
  }
]
```

We want the tree to still show all 4 documents, not just those 2. This piece is actually very similar to how it was before: https://github.com/microsoft/vscode-cosmosdb/pull/1572/files#diff-0cdcd82c7e289824f6c83cc3f04581d2L37

src/DatabasesFileSystem.ts
```diff
@@ -0,0 +1,125 @@
...
+        return { type: FileType.File, ctime: node.cTime, mtime: node.mTime, size };
+    }
+
+    public async readFileImpl(context: IActionContext, node: IEditableTreeItem): Promise<Uint8Array> {
+        return Buffer.from(await node.getFileContent(context));
+    }
+
+    public async writeFileImpl(context: IActionContext, node: IEditableTreeItem, content: Uint8Array, _originalUri: Uri): Promise<void> {
+        const showSavePromptKey: string = 'showSavePrompt';
+        const prefix: string = 'cosmosDB';
```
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I agree. I want the scheme for the file system to be generic so it can work for other resource types in the future, but I don't see enough reason (yet) to break backwards compatibility for people that have this setting. Added a note in the code to that effect

PR #1582, opened 2020-06-23
Return a new collection node if one doesn't exist

wwlorey:
From https://github.com/microsoft/vscode-azuretools/pull/749#issuecomment-648331257

### DIFF
--- a/src/DatabasesFileSystem.ts
+++ b/src/DatabasesFileSystem.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Collection, Db } from "mongodb";
-import { basename } from 'path';
+import { basename, dirname } from 'path';
 import { FileStat, FileType, MessageItem, Uri, workspace } from "vscode";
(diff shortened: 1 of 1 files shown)


src/DatabasesFileSystem.ts
```diff
@@ -84,7 +87,15 @@ export class DatabasesFileSystem extends AzExtTreeFileSystem<IEditableTreeItem>
     }
 
     protected async findItem(context: IActionContext, query: AzExtItemQuery): Promise<IEditableTreeItem | undefined> {
-        const node: IEditableTreeItem | undefined = await super.findItem(context, query);
+        let node: IEditableTreeItem | undefined = await super.findItem(context, query);
+        if (!node && ext.connectedMongoDB) {
```
EricJizbaMSFT, line 91: Hmm this `if` case does nothing to confirm the query is actually for a collection tree item. Can we fix that? One idea I had:
```
if (!node) {
   const parentId: string = // get parentId from query.id
   const parentNode: IEditableTreeItem | undefined = ext.tree.findTreeItem(parentId);
   if (parentNode instanceof MongoDatabaseTreeItem) {
        // create mongo collection tree item
   }
}
```

Also feels a bit more robust than using `ext.connectedMongoDB`

PR #1584, opened 2020-06-24
Display mongo command results in a single file

wwlorey:
Instead of creating a "result-x.json" file every time you execute a command, this creates one "Scrapbook-results.txt" file and appends each result to that. The results of `find` and `findOne` will still display in their own files.

I opted to use ".txt" instead of ".json" because the appended results won't always be valid json. 

After running "Execute All":
Screen Shot 2020-06-23 at 6 03 28 PM

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/607
Depends on https://github.com/microsoft/vscode-azuretools/pull/758

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -29,7 +29,7 @@ import { IMongoDocument, MongoDocumentTreeItem } from './tree/MongoDocumentTreeI
 const EJSON = require("mongodb-extended-json");
 
 const notInScrapbookMessage = "You must have a MongoDB scrapbook (*.mongo) open to run a MongoDB command.";
-const resultMap: Map<string, ReadOnlyContent> = new Map<string, ReadOnlyContent>();
+let defaultReadOnlyContent: ReadOnlyContent | undefined;
 
 export function getAllErrorsFromTextDocument(document: vscode.TextDocument): vscode.Diagnostic[] {
     const commands = getAllCommandsFromTextDocument(document);
@@ -137,9 +137,7 @@ async function executeCommand(context: IActionContext, command: MongoCommand, re
                 await ext.fileSystem.showTextDocument(docNode, { viewColumn: vscode.ViewColumn.Beside });
             } else {
                 const showOptions: vscode.TextDocumentShowOptions = { viewColumn: vscode.ViewColumn.Beside };
-                const label: string = 'Scrapbook-results';
-                const fullId: string = `${database.fullId}/${label}`;
-                readOnlyContent = readOnlyContent || resultMap.get(fullId);
+                readOnlyContent = readOnlyContent || defaultReadOnlyContent;
                 result += `${EOL}${EOL}`;
 
                 if (readOnlyContent) {
@@ -150,8 +148,9 @@ async function executeCommand(context: IActionContext, command: MongoCommand, re
                         await readOnlyContent.show(showOptions);
                     }
                 } else {
-                    readOnlyContent = await openReadOnlyContent({ label, fullId }, result, '.txt', showOptions);
-                    resultMap.set(fullId, readOnlyContent);
+                    const label: string = 'Scrapbook-results';
+                    const fullId: string = `${database.fullId}/${label}`;
+                    defaultReadOnlyContent = await openReadOnlyContent({ label, fullId }, result, '.txt', showOptions);
                 }
 
                 await refreshTreeAfterCommand(database, command, context);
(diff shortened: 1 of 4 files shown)


src/utils/vscodeUtils.ts
```diff
@@ -34,18 +35,36 @@ export async function showNewFile(data: string, fileName: string, fileExtension:
     const fullFileName: string | undefined = await getUniqueFileName(folderPath, fileName, fileExtension);
     uri = vscode.Uri.file(path.join(folderPath, fullFileName)).with({ scheme: 'untitled' });
     const textDocument = await vscode.workspace.openTextDocument(uri);
-    const editor = await vscode.window.showTextDocument(textDocument, column ? column > vscode.ViewColumn.Three ? vscode.ViewColumn.One : column : undefined, true);
+    const editor = await showTextDocument(textDocument, column);
     await writeToEditor(editor, data);
 }
 
-export async function writeToEditor(editor: vscode.TextEditor, data: string): Promise<void> {
+export async function createOrAppendToFile(data: string, fileName: string, fileExtension: string, column?: vscode.ViewColumn): Promise<void> {
```
EricJizbaMSFT, line 42: If you want to fix the save stuff, you could refactor [openReadonlyContent](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/openReadOnlyContent.ts) to support the `createOrAppend` scenario and use that

src/mongo/MongoScrapbook.ts
```diff
@@ -70,6 +73,8 @@ export function getAllCommandsFromTextDocument(document: vscode.TextDocument): M
 }
 
 async function executeCommands(context: IActionContext, commands: MongoCommand[]): Promise<void> {
+    clearResults = true;
+    isExecutingAll = true;
```
EricJizbaMSFT, line 77: What if you call `openReadOnlyContent` here and pass in `ReadOnlyContent` as an optional parameter to `executeCommand`. That way you know you always open a fresh read-only editor before you start an "Execute All" and I think that would allow you to get rid of all your file-scoped variables above

src/mongo/MongoScrapbook.ts
```diff
@@ -131,8 +136,24 @@ async function executeCommand(context: IActionContext, command: MongoCommand): P
                 const docNode = new MongoDocumentTreeItem(colNode, document);
                 await ext.fileSystem.showTextDocument(docNode, { viewColumn: vscode.ViewColumn.Beside });
             } else {
-                const viewColumn = vscode.window.activeTextEditor?.viewColumn;
-                await vscodeUtil.showNewFile(result, 'result', '.json', viewColumn ? viewColumn + 1 : undefined);
+                const showOptions: vscode.TextDocumentShowOptions = { viewColumn: vscode.ViewColumn.Beside };
+                const label: string = 'Scrapbook-results';
+                const fullId: string = `${database.fullId}/${label}`;
+                readOnlyContent = readOnlyContent || resultMap.get(fullId);
```
EricJizbaMSFT, line 142: Why do you need the `resultMap`? I thought it was only used for "Execute All", but that should be handled by passing in `readOnlyContent`

wwlorey: With the new "Execute All" implementation, `resultMap` is only used for appending each time plain old `executeCommand` is called (without passing in a value for `readOnlyContent`).

But it's overkill since that map would only ever have one item in it. I'll update it

src/mongo/MongoScrapbook.ts
```diff
@@ -131,8 +136,23 @@ async function executeCommand(context: IActionContext, command: MongoCommand): P
...
-                await vscodeUtil.showNewFile(result, 'result', '.json', viewColumn ? viewColumn + 1 : undefined);
+                const showOptions: vscode.TextDocumentShowOptions = { viewColumn: vscode.ViewColumn.Beside };
+                readOnlyContent = readOnlyContent || defaultReadOnlyContent;
+                result += `${EOL}${EOL}`;
+
+                if (readOnlyContent) {
+                    await readOnlyContent.append(result);
+
+                    // If the read only content isn't visible anymore, re-show it. Appending to a hidden document does not show it
+                    if (!(await readOnlyContent.isVisible())) {
```
EricJizbaMSFT, line 147: Should be able to get rid of this, too, since `openReadOnlyContent` was called at the beginning of "Execute All". If users click away from the output _while_ the commands are running, I feel like that's on them and we would just be annoying if we re-show it

wwlorey: The main reason I added that check is because if the output is closed, it won't be re-opened on its own when appending. So the results wouldn't be displayed in the case where `executeCommand` is called without `readOnlyContent`.

Should I not re-show in the "Execute All" case?

wwlorey: Per offline discussion: I misinterpreted part of the original issue, we don't want to append content to the output if a single command is run

PR #1585, opened 2020-06-24
Clean up error reporting in MongoScrapbook.ts

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/958

### DIFF
--- a/src/mongo/MongoScrapbook.ts
+++ b/src/mongo/MongoScrapbook.ts
@@ -257,14 +257,16 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
                     let ejsonParsed = {};
                     try {
                         ejsonParsed = EJSON.parse(escapeHandled);
-                    } catch (err) { //EJSON parse failed due to a wrong flag, etc.
-                        this.addErrorToCommand(parseError(err), ctx);
+                    } catch (error) { //EJSON parse failed due to a wrong flag, etc.
+                        const parsedError: IParsedError = parseError(error);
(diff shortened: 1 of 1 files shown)


src/mongo/MongoScrapbook.ts
```diff
@@ -257,14 +257,16 @@ class FindMongoCommandsVisitor extends MongoVisitor<MongoCommand[]> {
                     let ejsonParsed = {};
                     try {
                         ejsonParsed = EJSON.parse(escapeHandled);
-                    } catch (err) { //EJSON parse failed due to a wrong flag, etc.
-                        this.addErrorToCommand(parseError(err), ctx);
+                    } catch (error) { //EJSON parse failed due to a wrong flag, etc.
```
EricJizbaMSFT, line 260: So what exactly did you decide to do to clean these up? The issue is not clear what a "proper fix" is

I think the big thing that annoys me about this code is how often there are catch blocks and the fact that it's returning `{}` in a bunch of places. Normally we let errors bubble up (so that you don't have to return a bogus blank object) and handle them in a shared place

wwlorey: I just focused on [this comment](https://github.com/microsoft/vscode-cosmosdb/pull/959#discussion_r235570052). I can refactor to address that concern though

EricJizbaMSFT: Yeah I'm 95% sure that's what we meant by a "proper fix". That being said, I'm not sure if it's worth your time - I'll leave it up to you to decide. Approving your changes as-is, but would like to re-review if you decide to fix the other stuff

Issue #1241, opened 2019-11-15
An error occurs when attaching MongoDB Emulator


v-ruizh:
**OS:** Windows
**Build Version:** [20191114.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=12450&view=results)
**Emulator Version:** 2.7.1.0

**Repro Steps:**
1. Install Cosmos DB Emulator and start it.
2. Right click on 'Attached Database Accounts' -> Attach Emulator -> MongoDB.
3. Check the result.

**Expect:**
Successfully attached.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/45614674/68944884-f8226700-07e9-11ea-8184-2feafc0e4ea5.png)

**More info:**
This issue does not reproduce when attaching SQL Emulator.

nturinski:
Did you make sure you enabled the emulator for using MongoDB?  I was able to get it to work:

![image](https://user-images.githubusercontent.com/5290572/68982094-7d8e2300-07ba-11ea-82f3-8fbb7804059f.png)

Follow the instructions here: https://github.com/microsoft/vscode-cosmosdb/wiki/Debugging-Connection-Errors specifically 2.  That will show you how to open it with Mongo DB enabled.

v-dantch:
Thanks for your sharing. Verified with the latest build [20191208.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=12927) on Windows. This issue doesn't reproduce after enabling the emulator.

v-dantch:
Reopen this issue due to it reproduces on the latest build [20200415.13](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=17851&view=artifacts&type=publishedArtifacts).
![image](https://user-images.githubusercontent.com/34729092/79422898-45507200-7ff0-11ea-84ad-bf531a6c384e.png)

PR #1588, opened 2020-06-25
Fix self signed certificate error when connecting to Mongo emulator

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1241

### DIFF
--- a/main.js
+++ b/main.js
@@ -16,10 +16,10 @@ let perfStats = {
 
 Object.defineProperty(exports, "__esModule", { value: true });
 
-const extension = require('./dist/extension.bundle');
+const extension = require('./out/src/extension');
 
 async function activate(ctx) {
-    return await extension.activateInternal(ctx, perfStats);
+    return await extension.activateInternal(ctx, perfStats, true /* ignoreBundle */);
 }
 
 async function deactivate(ctx) {
--- a/src/mongo/connectToMongoClient.ts
+++ b/src/mongo/connectToMongoClient.ts
@@ -15,7 +15,7 @@ export async function connectToMongoClient(connectionString: string, appName: st
         useNewUrlParser: true
     };
 
-    if (isLocalEmulatorConnectionString(connectionString)) {
+    if (isCosmosEmulatorConnectionString(connectionString)) {
         // Prevents self signed certificate error for emulator https://github.com/microsoft/vscode-cosmosdb/issues/1241#issuecomment-614446198
         options.tlsAllowInvalidCertificates = true;
     }
@@ -43,7 +43,6 @@ export class MongoConnectError extends Error {
     }
 }
 
-export function isLocalEmulatorConnectionString(connectionString: string): boolean {
-    const localEmulatorRegex: RegExp = new RegExp(`(localhost|127\.0\.0\.1):${encodeURIComponent(emulatorPassword)}`, 'i');
-    return localEmulatorRegex.test(connectionString);
+export function isCosmosEmulatorConnectionString(connectionString: string): boolean {
+    return connectionString.includes(encodeURIComponent(emulatorPassword));
 }
--- a/src/mongo/connectToMongoClient.ts
+++ b/src/mongo/connectToMongoClient.ts
@@ -15,6 +15,11 @@ export async function connectToMongoClient(connectionString: string, appName: st
         useNewUrlParser: true
     };
 
+    if (!!connectionString.match(/^mongodb:\/\/(localhost|127\.0\.0\.1)/)) {
+        // Prevents self signed certificate error https://github.com/microsoft/vscode-cosmosdb/issues/1241#issuecomment-614446198
+        options.tlsAllowInvalidCertificates = true;
+    }
+
     try {
         return await MongoClient.connect(connectionString, options);
     } catch (err) {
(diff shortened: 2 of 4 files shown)


src/mongo/connectToMongoClient.ts
```diff
@@ -15,6 +15,11 @@ export async function connectToMongoClient(connectionString: string, appName: st
         useNewUrlParser: true
     };
 
+    if (!!connectionString.match(/^mongodb:\/\/(localhost|127\.0\.0\.1)/)) {
+        // Prevents self signed certificate error https://github.com/microsoft/vscode-cosmosdb/issues/1241#issuecomment-614446198
+        options.tlsAllowInvalidCertificates = true;
```
EricJizbaMSFT, line 20: I'm probably fine with this, but I'd like to see a more detailed explanation when you're loosening security-related settings. Basically, convince me that you're not increasing risk for our users

wwlorey: Mongo recommends that valid certificates should definitely be used in production environments ([source](https://docs.mongodb.com/manual/tutorial/configure-ssl/)). But since the emulator is likely being used for development, it seems fine to allow invalid certificates here. 

Would it be better to throw up a warning before setting `tlsAllowInvalidCertificates`? Just so this isn't happening entirely under the covers

EricJizbaMSFT: Well why do invalid certificate errors happen? I feel like I've never run into this error and I've used a local mongo instance - can you repro?

wwlorey: Yeah I was able to repro along with Nathan & CTI. It looks like certificate behavior changed in mongo v4.0: https://docs.mongodb.com/manual/release-notes/4.0/#x-509-authentication-certificate-restrictions. So it makes sense that we're just now seeing this type of error

src/mongo/connectToMongoClient.ts
```diff
@@ -15,6 +15,11 @@ export async function connectToMongoClient(connectionString: string, appName: st
         useNewUrlParser: true
     };
 
+    if (!!connectionString.match(/^mongodb:\/\/(localhost|127\.0\.0\.1)/)) {
```
EricJizbaMSFT, line 18: nit: I prefer `.test` over `.match` and I almost always include the `i` flag to be safe
```suggestion
    if (/^mongodb:\/\/(localhost|127\.0\.0\.1)/i.test(connectionString)) {
```

EricJizbaMSFT, line 18: [this](https://github.com/microsoft/vscode-cosmosdb/blob/master/test/mongoConnectionStrings.test.ts#L50) seems to imply that "localhost" isn't always the first thing after "mongodb://"

Might be worth creating an `isLocalhost` function and adding unit tests against the same connection strings in those tests. Never hurts to be careful when dealing with security stuff

src/mongo/connectToMongoClient.ts
EricJizbaMSFT: Are you reproducing with the Cosmos DB emulator for mongo or just plain old mongodb? If it only repros in the Cosmos DB case, we might not want to do this for everyone

wwlorey: I've only ever seen the error when using the emulator. Are you suggesting we only set this flag if the connection string is for the Cosmos DB emulator?

EricJizbaMSFT: Yeah

main.js
```diff
@@ -16,10 +16,10 @@ let perfStats = {
 
 Object.defineProperty(exports, "__esModule", { value: true });
 
-const extension = require('./out/src/extension');
+const extension = require('./dist/extension.bundle');
```
EricJizbaMSFT, line 19: revert these changes

src/mongo/connectToMongoClient.ts
```diff
@@ -37,3 +42,8 @@ export class MongoConnectError extends Error {
         super(`Unable to connect to local Mongo DB instance. Make sure it is started correctly. See ${Links.LocalConnectionDebuggingTips} for tips.`);
     }
 }
+
+export function isLocalEmulatorConnectionString(connectionString: string): boolean {
+    const localEmulatorRegex: RegExp = new RegExp(`(localhost|127\.0\.0\.1):${encodeURIComponent(emulatorPassword)}`, 'i');
```
EricJizbaMSFT, line 47: What do you think - could we just check `return connectionString.includes(encodeURIComponent(emulatorPassword));`? Maybe check for the un-encoded password, too (is that possible?)

Seems virtually impossible that exact string would show up in other connection strings.

src/mongo/connectToMongoClient.ts
```diff
@@ -37,3 +42,8 @@ export class MongoConnectError extends Error {
         super(`Unable to connect to local Mongo DB instance. Make sure it is started correctly. See ${Links.LocalConnectionDebuggingTips} for tips.`);
     }
 }
+
+export function isLocalEmulatorConnectionString(connectionString: string): boolean {
```
EricJizbaMSFT, line 46: Can you rename this to `isCosmosEmulatorConnectionString`

src/mongo/connectToMongoClient.ts
wwlorey: Yeah that makes sense. It seems unlikely that `emulatorPassword` would be used elsewhere.

I tried attaching the emulator with the un-encoded password but the extension threw an error:

Screen Shot 2020-07-09 at 12 33 21 PM

So I'll just verify the connection string includes the encoded password.

PR #1589, opened 2020-06-26
No Attached Accounts Message

### LINKED ISSUE #793: [Suggestion]Display 'No result/ Attach New Database Account' when there are no attached database accounts
Repro Steps:
1. Make sure there is no attached database account.
2. F1 -> Type 'Cosmos DB: Detach', then check the result.

Actual:
There is no any warning to prompt the customer there is no any attached database account.

Suggestion:
Display 'No result/ Attach New Database Account'.
suggestion

neelip:
Fixes #793


src/extension.ts
```diff
@@ -100,18 +100,25 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
-            if (!node) {
-                node = await ext.tree.showTreeItemPicker<AzureTreeItem>(cosmosDBTopLevelContextValues.map((val: string) => val += AttachedAccountSuffix), actionContext);
-            }
-            if (node instanceof MongoAccountTreeItem) {
-                if (ext.connectedMongoDB && node.fullId === ext.connectedMongoDB.parent.fullId) {
-                    setConnectedNode(undefined, codeLensProvider);
-                    await node.refresh();
+        registerCommand('cosmosDB.detachDatabaseAccount', async (actionContext: IActionContext & ITreeItemPickerContext, node?: AzureTreeItem) => {
+            const numChildren: number = ext.attachedAccountsNode.loadAllChildren.length;
+            if (numChildren < 2) {
```
wwlorey, line 105: This logic works if the only child is the "Attach Database Account..." node. But if you have a single attached account you won't be able to detach it.

Maybe instead you could check the context value of the child node here before showing that info message?

src/extension.ts
```diff
@@ -100,18 +100,25 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
         registerCommand('azureDatabases.refresh', async (_actionContext: IActionContext, node?: AzExtTreeItem) => await ext.tree.refresh(node));
-        registerCommand('cosmosDB.detachDatabaseAccount', async (actionContext: IActionContext, node?: AzureTreeItem) => {
-            if (!node) {
-                node = await ext.tree.showTreeItemPicker<AzureTreeItem>(cosmosDBTopLevelContextValues.map((val: string) => val += AttachedAccountSuffix), actionContext);
-            }
-            if (node instanceof MongoAccountTreeItem) {
-                if (ext.connectedMongoDB && node.fullId === ext.connectedMongoDB.parent.fullId) {
-                    setConnectedNode(undefined, codeLensProvider);
-                    await node.refresh();
+        registerCommand('cosmosDB.detachDatabaseAccount', async (actionContext: IActionContext & ITreeItemPickerContext, node?: AzureTreeItem) => {
```
wwlorey, line 103: nit: I would rename `actionContext` to `context` since the type is different now

src/extension.ts
```diff
@@ -100,18 +100,25 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+            const children = await ext.attachedAccountsNode.loadAllChildren(actionContext);
+            if (!("connectionString" in children[0])) {
```
wwlorey, line 105: Could you check the `contextValue` of `children[0]` here instead? While this check works, I think it's more robust to explicitly check for an expected context value

src/extension.ts
```diff
@@ -100,18 +100,25 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+                const message = localize('noAttachedAccounts', 'There are no Attached Accounts.');
+                vscode.window.showInformationMessage(message);
+                actionContext.suppressCreatePick = true;
```
wwlorey, line 108: We should be able to remove this line since the tree item picker isn't being shown

src/extension.ts
```diff
@@ -100,18 +100,24 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
         registerCommand('azureDatabases.refresh', async (_actionContext: IActionContext, node?: AzExtTreeItem) => await ext.tree.refresh(node));
-        registerCommand('cosmosDB.detachDatabaseAccount', async (actionContext: IActionContext, node?: AzureTreeItem) => {
-            if (!node) {
-                node = await ext.tree.showTreeItemPicker<AzureTreeItem>(cosmosDBTopLevelContextValues.map((val: string) => val += AttachedAccountSuffix), actionContext);
-            }
-            if (node instanceof MongoAccountTreeItem) {
-                if (ext.connectedMongoDB && node.fullId === ext.connectedMongoDB.parent.fullId) {
-                    setConnectedNode(undefined, codeLensProvider);
-                    await node.refresh();
+        registerCommand('cosmosDB.detachDatabaseAccount', async (_context: IActionContext & ITreeItemPickerContext, node?: AzureTreeItem) => {
```
wwlorey, line 103: My bad - I didn't realize `context` was going to be a shadowed variable here. 

We typically reserve names beginning with `_` for unused variables. Can you rename `_context` to not include the underscore? (using `actionContext` is fine). Sorry for the snip-snap

neelip: No worries. I did noticed that `_` was used rarely

Issue #1073, opened 2019-03-20
Question: how to change Mongodb aggregate limit


sublimeye:
Question:
1. How to make `limit` and `aggregate` with limit return more than 20 elements?
2. I saw `maxDocs` property mentioned in a couple of issues, but I don't understand where do I add it or maybe it's just a built-in internal `package.json` property that is controlled by the extension? I didn't find any mention of it in the documentation. I tried to set `maxDocs` as a global property in `settings.json` - it says that the property is unknown.

Current Behavior:
- `.find()` query returns as many objects as collection contains (at least up to 50)
- `.find().limit(N)` query returns `N` (if `N` <= 20), for `N` larger than 20 returns 20 only
- `.aggregate()` - returns 20 if `$limit` is not set
- `.aggregate([ {$limit: N} ])` – similar behavior to regular `find().limit()`. If `N` less than 20 returns N, otherwise return 20.

wwlorey:
Hi @sublimeye, we just released [v0.14.0](https://github.com/microsoft/vscode-cosmosdb/releases/tag/v0.14.0) which added a setting called "azureDatabases.batchSize" that you can use to configure the aggregate limit.

PR #1598, opened 2020-06-30
Add a setting to configure the Mongo shell batch size

wwlorey:
I got the default batch size of 20 from the mongo default (https://docs.mongodb.com/manual/tutorial/configure-mongo-shell/index.html#change-the-mongo-shell-batch-size)

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1073

### DIFF
--- a/package.json
+++ b/package.json
@@ -960,9 +960,9 @@
                     "default": true,
                     "description": "Show warning dialog when uploading a document to the cloud."
                 },
-                "cosmosDB.batchSize": {
+                "azureDatabases.batchSize": {
                     "type": "number",
-                    "description": "The batch size to be used when querying Cosmos DB resources.",
+                    "description": "The batch size to be used when querying Azure Database resources.",
                     "default": 50
                 },
                 "azureDatabases.enableOutputTimestamps": {
--- a/package.json
+++ b/package.json
@@ -935,6 +935,11 @@
                     "description": "The duration allowed (in seconds) for the Mongo shell to execute a command. Default value is 30 seconds.",
                     "default": 30
                 },
+                "mongo.shell.batchSize": {
+                    "type": "number",
+                    "description": "The maximum number of documents to be returned from a query executed in the Mongo shell.",
+                    "default": 20
+                },
                 "azureDatabases.showExplorer": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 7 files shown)


EricJizbaMSFT:
Uh I thought the issue was about adding a setting for this: https://github.com/microsoft/vscode-cosmosdb/blob/7cbe733979db5d49840cd803256aade75a73e862/src/constants.ts#L44

wwlorey:
> Uh I thought the issue was about adding a setting for this:

Oops I didn't realize that existed! I'll modify the setting to reflect that.

And I'll still have to update `MongoShell` to include that batch size setting

src/utils/workspacUtils.ts
```diff
@@ -4,9 +4,16 @@
...
+import { ext } from '../extensionVariables';
+import { nonNullValue } from './nonNull';
 
 // tslint:disable-next-line: export-name
 export function getRootPath(): string | undefined {
     // if this is a multi-root workspace, return undefined
     return vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1 ? vscode.workspace.workspaceFolders[0].uri.fsPath : undefined;
 }
+
+export function getBatchSizeSetting(): number {
```
wwlorey, line 16: I would have put this in constants.ts but languageService.ts can't take a dependency on vscode (languageService.ts consumes something from constants.ts)

package.json
```diff
@@ -960,6 +960,11 @@
                     "default": true,
                     "description": "Show warning dialog when uploading a document to the cloud."
                 },
+                "cosmosDB.batchSize": {
```
EricJizbaMSFT, line 963: Do we want this to be specific to Cosmos? I could see us using this batchSize for other database types in the future.

src/mongo/tree/MongoCollectionTreeItem.ts
```diff
@@ -126,11 +127,11 @@ export class MongoCollectionTreeItem extends AzureParentTreeItem<IMongoTreeRoot>
 
     public async loadMoreChildrenImpl(clearCache: boolean): Promise<AzExtTreeItem[]> {
         if (clearCache || this._cursor === undefined) {
-            this._cursor = this.collection.find(this._query).batchSize(defaultBatchSize);
+            this._batchSize = getBatchSizeSetting();
```
EricJizbaMSFT, line 130: I would put this kind of thing in `refreshImpl`

package.json
wwlorey: I thought that too but didn't want to be misleading since it only applies to Cosmos atm. 

With that said it's probably worth making it a generic setting to future proof it

PR #1600, opened 2020-07-01
Log info for deleting cosmos account

### LINKED ISSUE #1544: There is no log "Successfully deleted server..."
**OS:** All
**Build Version:** 20200527.16

**Repro Steps:**

1. Delete a server account.
2. Create a new server account while deleting the above server account.
3. Check output log.

**Expect:**
There is a log "Successfully deleted server...", such as:
image

**Actual:**
There is no log "Successfully deleted server...".
image

neelip:
Fixes #1544

### DIFF
--- a/src/commands/deleteCosmosDBAccount.ts
+++ b/src/commands/deleteCosmosDBAccount.ts
@@ -8,6 +8,7 @@ import * as vscode from 'vscode';
 import { AzureTreeItem, createAzureClient, DialogResponses, UserCancelledError } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { azureUtils } from '../utils/azureUtils';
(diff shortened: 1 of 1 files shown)


src/commands/deleteCosmosDBAccount.ts
```diff
@@ -21,7 +22,9 @@ export async function deleteCosmosDBAccount(node: AzureTreeItem): Promise<void>
             await client.databaseAccounts.deleteMethod(resourceGroup, accountName);
         });
         // don't wait
```
wwlorey, line 24: nit: it makes sense to keep this "don't wait" by the `showInformationMessage` call if we even need to keep it at all

neelip: might not need it actually

Issue #1209, opened 2019-09-30
Can't auth with username containing @


SkeLLLa:
Repro steps:
1. Try to login to mongo with connections string like `mongodb://username@example.com:password@localhost/` or `mongodb://username%40example.com:password@localhost/`
2. Get error

Action: cosmosDB.attachDatabaseAccount
Error type: Error
Error Message: Username containing an unescaped at-sign

Version: 0.11.0
OS: linux
Product: VSCodium
Product Version: 1.38.1
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:253:234753
l extension.bundle.js:253:240574
QueryReqWrap.callback extension.bundle.js:253:242137
QueryReqWrap.onresolve [as oncomplete] dns.js:199:10
```

</details>

nturinski:
Hi @SkeLLLA

Thanks for reporting.  I will investigate what we can do to automate this for users, but the current workaround is to escape @ signs with a \ (backslash.  For example:

`mongodb://username\@example.com:password\@localhost`

SkeLLLa:
I didn't remember if I tried escaping, so I can't be sure that it works or not.
However handling this case should be pretty simple: 
1) Find the latest index of `@`.
2) Everything before that index (username and password) should be escaped, and after - that the host part.

PR #1604, opened 2020-07-06
Encode mongo connection strings as needed

wwlorey:
I didn't want to always encode connection strings just in case they're already encoded. So this only encodes if an encoding-related error is thrown first.

Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1209

### DIFF
--- a/src/mongo/mongoConnectionStrings.ts
+++ b/src/mongo/mongoConnectionStrings.ts
@@ -49,24 +49,20 @@ export async function parseMongoConnectionString(connectionString: string): Prom
     let host: string;
     let port: string;
 
+    let mongoClient: MongoClient;
     try {
-        // Attempt to connect to determine if the username/password need to be encoded
-        await connectToMongoClient(connectionString, appendExtensionUserAgent());
+        mongoClient = await connectToMongoClient(connectionString, appendExtensionUserAgent());
     } catch (error) {
         const parsedError: IParsedError = parseError(error);
         if (parsedError.message.match(/unescaped/i)) {
-            const matches: RegExpMatchArray | null = connectionString.match(/^(.*):\/\/(.*):(.*)@(.*)/);
-            if (matches) {
-                const prefix: string = matches[1];
-                const username: string = matches[2];
-                const password: string = matches[3];
-                const hostAndQuery: string = matches[4];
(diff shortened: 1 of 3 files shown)


src/mongo/mongoConnectionStrings.ts
```diff
@@ -49,6 +49,23 @@ export async function parseMongoConnectionString(connectionString: string): Prom
     let host: string;
     let port: string;
 
+    try {
+        // Attempt to connect to determine if the username/password need to be encoded
+        await connectToMongoClient(connectionString, appendExtensionUserAgent());
```
EricJizbaMSFT, line 54: Can you combine this with line 69? No need to connect twice for the happy-path.

```
let mongoClient: MongoClient;
try {
    mongoClient = await connectToMongoClient(connectionString, appendExtensionUserAgent());
} catch (error) {
    // unescaped stuff
    mongoClient = await connectToMongoClient(connectionString, appendExtensionUserAgent());
}
```

src/mongo/mongoConnectionStrings.ts
```diff
@@ -49,6 +49,23 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
+    } catch (error) {
+        const parsedError: IParsedError = parseError(error);
+        if (parsedError.message.match(/unescaped/i)) {
+            const matches: RegExpMatchArray | null = connectionString.match(/^(.*):\/\/(.*):(.*)@(.*)/);
```
EricJizbaMSFT, line 58: How about pulling some of this stuff into it's own function and adding some unit tests? Other functions in this file already have it.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -49,6 +49,23 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
+            if (matches) {
+                const prefix: string = matches[1];
+                const username: string = matches[2];
+                const password: string = matches[3];
+                const hostAndQuery: string = matches[4];
+                connectionString = `${prefix}://${encodeURIComponent(username)}:${encodeURIComponent(password)}@${hostAndQuery}`;
```
EricJizbaMSFT, line 64: Nathan's recommendation in the issue was to escape with a slash, but you're using `encodeURIComponent` which is different. Do they both work?

wwlorey: I was not able to get escaping with a slash to work

PR #1607, opened 2020-07-08
Suppress "Create..." option for delete commands tree item picker

### LINKED ISSUE #1441: Display the option "+Create new Account…" after selecting "Cosmos DB: Delete Account…" from Command palette
**OS:** Win10
**Build Version:** [20200427.6](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=18396&view=results)

**Repro Steps:**
1. F1 --> Type and select "Cosmos DB: Delete Account...".
2. Prompt to select subscription.
3. Prompt to select Account to delete.

**Expect:**
Only display the existing accounts.

**Actual:**
Display the option "+Create new Account…".
image

**More Info:**
This issue also reproduces on all delete command from Command Palette.

neelip:
Fixes #1441

### DIFF
--- a/src/docdb/registerDocDBCommands.ts
+++ b/src/docdb/registerDocDBCommands.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { commands } from "vscode";
-import { IActionContext, registerCommand } from "vscode-azureextensionui";
+import { IActionContext, ITreeItemPickerContext, registerCommand } from "vscode-azureextensionui";
 import { doubleClickDebounceDelay } from "../constants";
 import { ext } from "../extensionVariables";
 import { DocDBAccountTreeItem } from "./tree/DocDBAccountTreeItem";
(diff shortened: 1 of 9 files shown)


src/docdb/registerDocDBCommands.ts
```diff
@@ -46,12 +46,16 @@ export function registerDocDBCommands(): void {
 
     });
     registerCommand('cosmosDB.deleteDocDBDatabase', async (context: IActionContext, node?: DocDBDatabaseTreeItem) => {
+        const suppressCreateContext: ITreeItemPickerContext = context;
```
wwlorey, line 49: It seems a little odd to declare a new variable for the context here. What if you change the `context` parameter type to `ITreeItemPickerContext` (which extends `IActionContext`). Then you can just set `context.suppressCreatePick = true`

EricJizbaMSFT: Related to [this comment](https://github.com/microsoft/vscode-cosmosdb/pull/1449#discussion_r422372415), Neelima's current method actually has better type validation. I would be totally fine with Will's suggestion if we turned on `strictFunctionTypes` in tsconfig. Wonder how hard that would be?

neelip: Should we hold off on turning on strictFunctionTypes for this release? I am not sure what other parts of code will be affected by it

wwlorey: I don't think we have time to mess with tsconfig before the next release

Issue #1619, opened 2020-07-13
The document is still updated with the changes after clicking "X" or "Don't upload" button in the notification


v-ruizh:
**OS:** All
**Build Version:** 20200711.5

**Repro Steps:**
1. Open one document and make some changes for it.
2. Press "Ctrl+S", pop up a notification.
![image](https://user-images.githubusercontent.com/45614674/87283871-55f98100-c528-11ea-813c-cdefa38bcfdd.png)
3. Click "X" or "Don't upload" button in the above notification.
4. Close and reopen the document, then check the result.

**Expect:**
The document is not updated.

**Actual:**
The document is updated with the changes.

wwlorey:
The document isn't updated on the server unless "Upload" or "Always upload" is selected. If you refresh without uploading, the unsaved change will be gone.

PR #1622, opened 2020-07-13
Clear un-uploaded changes to Mongo & DocDB documents before opening

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1619

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -150,7 +150,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
             }
 
             // Clear un-uploaded local changes to the document before opening https://github.com/microsoft/vscode-cosmosdb/issues/1619
-            await node.refresh();
+            ext.fileSystem.fireChangedEvent(node);
             await ext.fileSystem.showTextDocument(node);
             // tslint:disable-next-line:align
         }, doubleClickDebounceDelay);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -149,6 +149,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
                 node = await ext.tree.showTreeItemPicker<MongoDocumentTreeItem | DocDBDocumentTreeItem>([MongoDocumentTreeItem.contextValue, DocDBDocumentTreeItem.contextValue], actionContext);
             }
 
+            // Clear un-uploaded local changes to the document before opening https://github.com/microsoft/vscode-cosmosdb/issues/1619
+            await node.refresh();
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -149,6 +149,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
                 node = await ext.tree.showTreeItemPicker<MongoDocumentTreeItem | DocDBDocumentTreeItem>([MongoDocumentTreeItem.contextValue, DocDBDocumentTreeItem.contextValue], actionContext);
             }
 
+            // Clear un-uploaded local changes to the document before opening https://github.com/microsoft/vscode-cosmosdb/issues/1619
+            await node.refresh();
```
EricJizbaMSFT, line 153: Our extension isn't caching the local changes, so you don't need to refresh the whole node. Pretty sure VS Code is doing the caching, in which case this should work:
```
ext.fileSystem.fireChangedEvent(node);
```

We call `fireChangedEvent` in `refreshImpl`, which I'm pretty sure is the only reason `node.refresh()` worked in the first place.

src/extension.ts
```diff
@@ -149,6 +149,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
                 node = await ext.tree.showTreeItemPicker<MongoDocumentTreeItem | DocDBDocumentTreeItem>([MongoDocumentTreeItem.contextValue, DocDBDocumentTreeItem.contextValue], actionContext);
             }
 
+            // Clear un-uploaded local changes to the document before opening https://github.com/microsoft/vscode-cosmosdb/issues/1619
+            ext.fileSystem.fireChangedEvent(node);
```
EricJizbaMSFT, line 153: Actually any chance you could do this in [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/AzExtTreeFileSystem.ts#L40)? If you don't want to do it for this release candidate that's fine, but I think other extensions will have the same problem, so I want it in the shared package eventually

wwlorey: Sounds good. I logged a new issue for it: https://github.com/microsoft/vscode-cosmosdb/issues/1624

EricJizbaMSFT: Okay well this one will take longer to discuss during triage than it'll take to fix. So if you don't want to do it within like a day - I'll just do it myself haha. Let me know

wwlorey: Lol gotcha, I can take care of it right after we ship tomorrow

PR #1636, opened 2020-07-28
Parse Postgres connection string function

neelip:


src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { parse } from "pg-connection-string";
+import { ParsedConnectionString } from "../ParsedConnectionString";
+import { nonNullProp } from "../utils/nonNull";
+
+export async function parsePostgresConnectionString(connectionString: string): Promise<ParsedPostgresConnectionString> {
```
EricJizbaMSFT, line 10: What's the purpose of adding this function/file? I prefer to see how something is used when it's added

neelip: This is to parse the postgres connection string for attached accounts support. The latest PR for postgres attached accounts includes this commit.

nturinski:
Does this still need review?  It seems like this PR https://github.com/microsoft/vscode-cosmosdb/pull/1640 has code that is more up to date, so I assume that we can close this one?

neelip:
Sounds good! This PR doesn't contain the latest code.

PR #1640, opened 2020-07-30
Attached Accounts support for Postgres

neelip:


neelip:
This version works for connection strings with/without the database name and will display all databases under the account. Do we want to keep it this way or restrict displaying the databases based on the database name included in the connection string?

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -95,21 +96,41 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            await client.databases.deleteMethod(azureUtils.getResourceGroupFromId(this.fullId), this.parent.name, this.databaseName);
+        } else {
+            const username_connString = nonNullProp(this.parent.connectionString, 'username');
+            const password_connString = nonNullProp(this.parent.connectionString, 'password');
+            const host = nonNullProp(this.parent.connectionString, 'hostName');
+            const config = { user: username_connString, password: password_connString, host, port: 5432 };
+            dropdb(config, this.databaseName, (err) => {
+                if (err) {
+                    console.error(err);
+                    process.exit(-1);
```
wwlorey, line 110: Is there a reason you did this instead of throwing an error like we normally do?

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -95,21 +96,41 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
     public async deleteTreeItemImpl(): Promise<void> {
-        const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
-        await client.databases.deleteMethod(azureUtils.getResourceGroupFromId(this.fullId), this.parent.name, this.databaseName);
+        if (this.parent.server) {
+            const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
+            await client.databases.deleteMethod(azureUtils.getResourceGroupFromId(this.fullId), this.parent.name, this.databaseName);
+        } else {
+            const username_connString = nonNullProp(this.parent.connectionString, 'username');
+            const password_connString = nonNullProp(this.parent.connectionString, 'password');
+            const host = nonNullProp(this.parent.connectionString, 'hostName');
```
wwlorey, line 105: Why do you have to call `nonNullProp` one these connection string properties?

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -120,6 +141,18 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            const clientConfig: ClientConfig = { user: username_connString, password: password_connString, ssl, host, port: 5432, database: this.databaseName };
+            const client: Client = new Client(clientConfig);
+            try {
+                await client.connect();
+                return clientConfig;
+            } finally {
+                await client.end();
+            }
```
wwlorey, line 155: Is it possible to move this somewhere where it can be shared with the same code above?

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -68,16 +83,39 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+        if (this.connectionString) {
+            const username_connString = nonNullProp(this.connectionString, 'username');
+            const password_connString = nonNullProp(this.connectionString, 'password');
+            const host = nonNullProp(this.connectionString, 'hostName');
+            const ssl: ConnectionOptions = {
+                // Always provide the certificate since it is accepted even when SSL is disabled
+                // Certificate source: https://aka.ms/AA7wnvl
+                ca: BaltimoreCyberTrustRoot
+            };
+            const config: ClientConfig = { user: username_connString, password: password_connString, ssl, host, port: 5432, database: 'postgres' };
```
wwlorey, line 95: Can this be handled in `getClientConfig`? It looks to be the same code

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -71,6 +74,16 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
         return MONGO_CONNECTION_EXPECTED;
     }
 
+    public static validatePostgresConnectionString(value: string): string | undefined {
```
wwlorey, line 77: Does the "pg-connection-string" package have anything that can do this for us?

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -299,6 +319,10 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
 
             label = label || `${id} (${getExperienceFromApi(api).shortName})`;
             treeItem = new MongoAccountTreeItem(this, id, label, connectionString, isEmulator);
+            // tslint:disable-next-line: possible-timing-attack
```
wwlorey, line 322: Lol this sounds scary I've never seen it before. Why do we need to disable the rule here?

wwlorey:
> This version works for connection strings with/without the database name and will display all databases under the account. Do we want to keep it this way or restrict displaying the databases based on the database name included in the connection string?

I think it would be best to support connection strings with/without the database name. Both versions seem useful

src/postgres/tree/PostgresDatabaseTreeItem.ts
neelip: Thought callback option was the only option available with this function. I checked the library, actually callback is not necessary as promise API is available as well.

src/postgres/tree/PostgresDatabaseTreeItem.ts
neelip: Actually nonNullProp is not necessary here, I think it is moved code which was not updated.

src/tree/AttachedAccountsTreeItem.ts
neelip: Lol I know, the lint rule has to do with passwords in general more related to security issues. I am not sure why it shows up here tho, even the earlier version of this code disabled it. The same comment is present again few lines above for mongo check.

src/tree/AttachedAccountsTreeItem.ts
neelip: The pg-connection-string library only has parse function. I feel like it is unnecessary here for quick initial check.

neelip:
There is now two attached accounts formats for postgres based on connection string. 
- If the connection string contained database, then only that specified database will be displayed all the time. 
- The other case is without the database specified, where the whole account is displayed. 

The only problem with this format is that the "Create Database..." command wouldn't work when connection string is specific to a database. Right now, we are getting a `Database [new database name] does not exist` after the responses to the "Create Database..." prompt. Any suggestions on how to handle it for this release.

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,40 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ConnectionOptions, parse } from "pg-connection-string";
+import { ParsedConnectionString } from "../ParsedConnectionString";
+import { nonNullProp } from "../utils/nonNull";
+
+export async function parsePostgresConnectionString(connectionString: string): Promise<ParsedPostgresConnectionString> {
+    connectionString = connectionString ? connectionString.trim() : '';
```
EricJizbaMSFT, line 11: Can `connectionString` be `undefined`? If yes - you should update the type to reflect that. If no - I feel like you could just do `connectionString = connectionString.trim();`

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,40 @@
...
+    const config: ConnectionOptions = parse(connectionString);
+    return new ParsedPostgresConnectionString(connectionString, config, accountConnection);
+}
+
+export class ParsedPostgresConnectionString extends ParsedConnectionString {
+    public readonly hostName: string;
+    public readonly port: string;
+    public readonly username: string;
+    public readonly password: string;
+    public readonly accountConnection: boolean;
```
EricJizbaMSFT, line 30: You don't need an `accountConnection` property. `ParsedConnectionString` already states that `databaseName` is what should be used to determine if it's an account connection or not. One reason that's preferred is because it helps the compiler automatically figure out when `databaseName` is defined and means you don't have to use `nonNullProp` as much. For example, this would throw a typescript error because `string | undefined` can't be assigned to `string`:
```
if (!connectionString.accountConnection) {
    const example: string = connectionString.databaseName;
}
```

However, this would have no error:
```
if (connectionString.databaseName) {
    const example: string = connectionString.databaseName;
}
```

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -71,6 +74,16 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
     }
 
+    public static validatePostgresConnectionString(value: string): string | undefined {
+        value = value ? value.trim() : '';
+
+        if (value && value.match(/^postgres:\/\//)) {
+            return undefined;
+        }
+
+        return POSTGRES_CONNECTION_EXPECTED;
```
EricJizbaMSFT, line 84: You should only make a constant for something like `POSTGRES_CONNECTION_EXPECTED` if you're going to use it in more than one place

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,40 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ConnectionOptions, parse } from "pg-connection-string";
+import { ParsedConnectionString } from "../ParsedConnectionString";
+import { nonNullProp } from "../utils/nonNull";
+
+export async function parsePostgresConnectionString(connectionString: string): Promise<ParsedPostgresConnectionString> {
+    connectionString = connectionString ? connectionString.trim() : '';
+    let accountConnection: boolean = false;
+    if (!connectionString.match(/^postgres:\/\/[^\/\s]+\/[^\/\s]+$/)) {
```
EricJizbaMSFT, line 13: Why do you need to add "/postgres" to the connection string?

EricJizbaMSFT, line 13: If you're going to have your own parsing logic, you need to add unit tests similar to mongoConnectionString tests. This kind of complex regexp/chatAt/index stuff is too easy to mess up an edge case. If you only rely on "pg-connection-string" (the ideal case) - we don't need our own unit tests.

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -95,58 +95,20 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
-                ca: BaltimoreCyberTrustRoot
-            };
-
-            const host: string = nonNullProp(this.parent.server, 'fullyQualifiedDomainName');
-            const clientConfig: ClientConfig = { user: username, password, ssl, host, port: 5432, database: this.databaseName };
-
-            // Ensure the client config is valid before returning
-            const client: Client = new Client(clientConfig);
+        if (this.parent.server) {
+            const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
```
EricJizbaMSFT, line 99: Ideally the attached and non-attached databases use the same code for deleting. The more they share code, the less testing we have to do to ensure they both work. For example, I don't think the Mongo tree items call any Azure-specific methods for Databases/collections/documents

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -95,58 +95,20 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
+            const host = this.parent.connectionString.hostName;
+            const config = { user: username_connString, password: password_connString, host, port: 5432 };
             try {
-                await client.connect();
-                return clientConfig;
-            } finally {
-                await client.end();
+                await dropdb(config, this.databaseName);
+            } catch (error) {
+                throw error;
```
EricJizbaMSFT, line 109: Haha just remove the try/catch if it's just gonna throw the error that would've been thrown anyways

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -32,15 +36,20 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     public readonly contextValue: string = PostgresServerTreeItem.contextValue;
     public readonly childTypeLabel: string = "Database";
     public readonly server: Server;
+    public readonly connectionString: ParsedPostgresConnectionString;
     public resourceGroup: string;
 
     private _serverId: string;
 
-    constructor(parent: AzureParentTreeItem, server: Server) {
+    constructor(parent: AzureParentTreeItem, server?: Server, connectionString?: ParsedPostgresConnectionString) {
```
EricJizbaMSFT, line 44: Similar to another comment, but the connectionString should be the preferred way of doing things because it works for attached and non attached. So you should make `connectionString` a required parameter and move it before `server`. Also you should use `server` as little as possible. Similar to what MongoAccountTreeItem does

neelip: For CosmosDB accounts, CosmosDBManagmentClient provides `listConnectionsStrings` for Mongo accounts. However, for Postgres accounts we don't have access to postgres connection strings through the PostgreSQLManagmentClient, we only have access to Server or Database. However, we can create the connection string and add it to the connectionString field but it won't be initialized in the constructor but in the loadMore... function of ...ServerTreeItem. Would that be ok?

package.json
```diff
@@ -1038,7 +1050,9 @@
         "ms-rest": "^2.2.1",
         "ms-rest-azure": "^2.3.1",
         "pg": "^7.18.1",
+        "pg-connection-string": "^2.3.0",
         "pg-structure": "^5.10.7",
+        "pgtools": "^0.3.0",
```
EricJizbaMSFT, line 1055: Still need to remove pgtools

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -32,15 +35,20 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     public readonly contextValue: string = PostgresServerTreeItem.contextValue;
     public readonly childTypeLabel: string = "Database";
     public readonly server: Server;
+    public connectionString: ParsedPostgresConnectionString;
     public resourceGroup: string;
```
EricJizbaMSFT, line 39: resourceGroup should be `string | undefined` now

src/postgres/commands/checkAuthentication.ts
```diff
@@ -20,7 +20,7 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
 
             if (parsedError.errorType === invalidCredentialsErrorType) {
                 await enterPostgresCredentials(context, treeItem.parent);
-            } else if (parsedError.errorType === firewallNotConfiguredErrorType) {
+            } else if (treeItem.parent.resourceGroup && parsedError.errorType === firewallNotConfiguredErrorType) {
```
EricJizbaMSFT, line 23: It's not automatically obvious why `configurePostgresFirewall` needs a resource group, but `enterPostgresCredentials` doesn't. Either change the code or add a comment (in the code, not here) why you're doing what you're doing

src/postgres/getClientConfig.ts
```diff
@@ -11,35 +11,38 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
...
-        }
+        host = nonNullProp(treeItem.connectionString, 'hostName');
     } else {
         throw {
             message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.'),
             code: invalidCredentialsErrorType
         };
     }
+    clientConfig = { user: username, password: password, ssl, host, port: 5432, database: databaseName };
+    client = new Client(clientConfig);
```
EricJizbaMSFT, line 36: Please keep this comment in the code:
```
// Ensure the client config is valid before returning
```

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ParsedConnectionString } from "../ParsedConnectionString";
+import { nonNullProp } from "../utils/nonNull";
+
+export async function parsePostgresConnectionString(connectionString: string): Promise<ParsedPostgresConnectionString> {
+    const config: ConnectionOptions = parse(connectionString.trim());
+    return new ParsedPostgresConnectionString(connectionString, config);
+}
+
+export async function createPostgresConnectionString(host: string, username?: string | undefined, password?: string | undefined): Promise<ParsedPostgresConnectionString> {
+    const port = '5432';
```
EricJizbaMSFT, line 16: You hard-code 5432 in several places. Make a `const` for it. If you use it in several files, stick it in the `constants.ts` file

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -53,49 +51,22 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
...
-
-            if (this.parent.supportsStoredProcedures()) {
-                children.push(new PostgresStoredProceduresTreeItem(this, clientConfig));
-            }
-
-            return children;
-        } catch (error) {
-            const parsedError: IParsedError = parseError(error);
-
-            if (parsedError.errorType === invalidCredentialsErrorType) {
```
EricJizbaMSFT, line 71: How come all of this stuff is gone? This kind of stuff is worth keeping, even if it only applies for the Azure case

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -52,11 +60,18 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     }
 
     public get name(): string {
-        return nonNullProp(this.server, 'name');
+        if (this._serverId) {
+            return this.connectionString.name;
```
EricJizbaMSFT, line 64: I don't understand why one uses name, but the other fullId? Can you give some real-life examples maybe?

src/postgres/tree/PostgresDatabaseTreeItem.ts
neelip: I just checked that it was gone. I was working with enabling authentication for attached accounts  and testing somethings, might have not undone the changes here.

src/postgres/tree/PostgresServerTreeItem.ts
neelip: For subscription accounts, 
`connectionString.fullId = "test-configure-firewall-log.postgres.database.azure.com:5432"`
`this._serverId = "/subscriptions/70593b55-43fd-4d86-9326-b33ac00cfdf1/resourceGroups/myResourceGroup/providers/Microsoft.DBforPostgreSQL/servers/test-azure-postgresql"`

We are using `serverId` for setting credentials and configuring firewall. I felt that the serverId provides a little bit more precision when storing credentials and firewall, when compared to using just the `connectionString.fullId`.

src/postgres/getClientConfig.ts
```diff
@@ -5,41 +5,47 @@
...
+        username = treeItem.connectionString.username;
+        password = treeItem.connectionString.password;
+    }
+    let host: string;
+    const ssl: ConnectionOptions = {
+        // Always provide the certificate since it is accepted even when SSL is disabled
+        // Certificate source: https://aka.ms/AA7wnvl
+        ca: BaltimoreCyberTrustRoot
+    };
+    let clientConfig: ClientConfig;
```
EricJizbaMSFT, line 26: nit: No need to define host, clientConfig, and client up here with `let`. Just define them as a `const` below

src/postgres/tree/PostgresServerTreeItem.ts
EricJizbaMSFT: Okay I see what's happening here. A few suggestions:
- Please rename `_serverId` to `_azureId`

IMO, this makes it more obvious where the id is coming from and why it doesn't exist in some cases.

- Remove `ParsedPostgresConnectionString.name`
- Remove `PostgresServerTreeItem.name`
- Add `PostgresServerTreeItem.azureName`

`azureName` should be an optional param to the constructor, instead of something you try to parse from the connection string. Since `azureName` will have the type `string | undefined`, this will have better type-safety for Azure-specific scenarios. For example, it will force you to call `nonNullProp` inside of `deleteTreeItemImpl`, which would throw an error we can easily watch out for in telemetry. As another example, label would look like this:
```
public get label(): string {
    return this.azureName || this.connectionString.fullId;
}
```

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,41 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ConnectionOptions, parse } from "pg-connection-string";
+import { postgresDefaultPort } from "../constants";
+import { ParsedConnectionString } from "../ParsedConnectionString";
+import { nonNullProp } from "../utils/nonNull";
+
+export async function parsePostgresConnectionString(connectionString: string): Promise<ParsedPostgresConnectionString> {
```
EricJizbaMSFT, line 11: Neither `parsePostgresConnectionString` nor `createPostgresConnectionString` need to be async

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -41,7 +39,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
     }
 
     public get id(): string {
-        return this.databaseName;
+        return 'database\/' + this.databaseName;
```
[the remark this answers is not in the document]
neelip: #### Attached Accounts #1 (Without Database in connection string): 

`"postgres://nepothar@test-configure-firewall-log:123456@test-configure-firewall-log.postgres.database.azure.com:5432"`
[...ServerTreeItem].fullId = `/cosmosDBAttachedAccounts/test-configure-firewall-log.postgres.database.azure.com:5432`
**[...DatabaseTreeItem].fullId** = `/cosmosDBAttachedAccounts/test-configure-firewall-log.postgres.database.azure.com:5432/test`

#### Attached Account #2 (With Database in the connection string):

`"postgres://nepothar@test-configure-firewall-log:123456@test-configure-firewall-log.postgres.database.azure.com:5432/test"`
**[...ServerTreeItem].fullId** = `/cosmosDBAttachedAccounts/test-configure-firewall-log.postgres.database.azure.com:5432/test`
[...DatabaseTreeItem].fullId = `/cosmosDBAttachedAccounts/test-configure-firewall-log.postgres.database.azure.com:5432/test/test`

The `[...DatabaseTreeItem].fullId` of one account and `[...ServerTreeItem].fullId` of another attached account are the same, which is causing an "Element ID already exists..." error. One account is for attaching all the databases and another one to attach just that specific database. 

Having `/database/[databaseName]` differentiates both the scenarios.

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -94,8 +92,8 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
-        const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
-        await client.databases.deleteMethod(azureUtils.getResourceGroupFromId(this.fullId), this.parent.name, this.databaseName);
+        const config = await getClientConfig(this.parent, 'postgres');
+        const pool = new Pool(config);
```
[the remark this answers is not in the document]
neelip: There doesn't seem to be any difference for this use case. I was having some problems earlier with client.query, I realize it was because I was not closing the connection. runPostgresQuery works for this scenario as well.

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class ParsedPostgresConnectionString extends ParsedConnectionString {
+    public readonly hostName: string;
+    public readonly username: string;
+    public readonly password: string;
+    public readonly name: string;
+    public readonly port: string;
+
+    constructor(connectionString: string, config: ConnectionOptions) {
+        super(connectionString, config.database?.replace(/[']+/g, ''));
```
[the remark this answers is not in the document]
neelip: Connection strings contain quotes around the database name to preserve if full database name (in case it contains extra leading or trailing whitespace). And these quotes are read in as part of the string by parse function from pg-connection-string.

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -32,8 +31,8 @@ export async function configurePostgresFirewall(context: IActionContext, treeIte
 export async function setFirewallRule(treeItem: PostgresServerTreeItem, ip: string): Promise<void> {
 
     const client: PostgreSQLManagementClient = createAzureClient(treeItem.root, PostgreSQLManagementClient);
-    const resourceGroup: string = azureUtils.getResourceGroupFromId(treeItem.id);
-    const serverName: string = nonNullProp(treeItem.server, 'name');
+    const resourceGroup: string = nonNullProp(treeItem, 'resourceGroup');
+    const serverName: string = nonNullProp(treeItem, 'label');
```
EricJizbaMSFT, line 35: You should only use `label` for stuff displayed to the user (like the message above is fine). For this, you should make `_azureName` public and use that. There's a few other instances like this throughout the PR

src/postgres/postgresConnectionStrings.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Could it use double quotes instead of single quotes? Also, is it possible to have a database name with a single quote in the name?

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -96,24 +118,26 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
         return new PostgresDatabaseTreeItem(this, databaseName);
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
         const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
-        const deletingMessage: string = `Deleting server "${this.name}"...`;
+        const deletingMessage: string = `Deleting server "${this.label}"...`;
         await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: deletingMessage }, async () => {
-            await client.servers.deleteMethod(this.resourceGroup, this.name);
+            await client.servers.deleteMethod(nonNullProp(this, 'resourceGroup'), this.label);
```
EricJizbaMSFT, line 140: Use azureName instead of label

src/postgres/commands/configurePostgresFirewall.ts
neelip: azureName can only be used in cases of firewall or credentials related code. Label is necessary for other cases as the naming would differ between attached and azure accounts.

neelip: I can add the name() function back to the ServerTreeItem instead of using azureName directly in few places and label in others.

src/postgres/postgresConnectionStrings.ts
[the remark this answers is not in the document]
neelip: I haven't found any standard online for using double or single quotes. Why would you prefer double quotes? That's a good question. Actually database names can have pretty much any character including  single and double quotes.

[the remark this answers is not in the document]
EricJizbaMSFT: _I_ don't prefer double quotes. I'm saying the _user_ could use double quotes and then your code would have problems.

Also, if the database name can have quotes in it, that will also cause problems with your current 'replace' logic.

src/postgres/commands/configurePostgresFirewall.ts
neelip: Changed label to azureName for non UI use cases.

src/postgres/postgresConnectionStrings.ts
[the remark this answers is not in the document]
neelip: That's true. Fixed this with replacing double quotes only and even the copy connection string outputs database name with double quotes around it.

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -68,15 +75,30 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+            const { username, password } = await this.getCredentials();
+            if (username && password) {
+                await this.updateConnectionString(username, password);
+            }
+        }
+        let config;
+        let query: string;
+        if (this.connectionString.databaseName) {
+            config = await getClientConfig(this, this.connectionString.databaseName);
+            query = `SELECT datname FROM pg_catalog.pg_database WHERE datistemplate = false and datname = '` + nonNullProp(this.connectionString, 'databaseName') + `';`;
```
EricJizbaMSFT, line 88: I'd like you to make a function like "wrapArgInQuotes" (maybe put it in "runPostgresQuery.ts"?) so that our quoting logic is centralized. It'll ensure we're consistent, but also make it easier to change the behavior for all queries at once. Quoted identifiers are case sensitive and unquoted identifiers are case insensitive - that doesn't feel like something where one option will satisfy all users. In other words, "wrapArgInQuotes" may be a one-liner in this PR, but I don't expect it to stay that way

src/postgres/tree/PostgresDatabaseTreeItem.ts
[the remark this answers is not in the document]
neelip: Sure, in that case we cannot attach account and database connection strings from the same server at the same time. One option is to only have one behavior which is to allow single database connections or just account connections?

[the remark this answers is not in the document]
EricJizbaMSFT: > in that case we cannot attach account and database connection strings from the same server at the same time

Yes I'm fine with this. I'm pretty sure mongo has the same limitation and I don't think any users have complained yet

src/postgres/commands/copyConnectionString.ts
```diff
@@ -23,7 +23,7 @@ export async function copyConnectionString(context: IActionContext, node: Postgr
     const password: string = nonNullProp(clientConfig, 'password');
     const host: string = nonNullProp(clientConfig, 'host');
     const port: string = String(nonNullProp(clientConfig, 'port'));
-    const connectionString = `postgres://${user}:${password}@${host}:${port}/'${node.databaseName}'`;
+    const connectionString = `postgres://${user}:${password}@${host}:${port}/"${node.databaseName}"`;
```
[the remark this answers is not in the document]
neelip: "To write an empty value, or a value containing spaces, surround it with single quotes" https://www.postgresql.org/docs/10/libpq-connect.html
But given databaseName can contain single quotes, double quotes seems to be a more reasonable option. I do agree it is not the best way to approach this. We can try something such as encodeURIComponent and fill in the whitespaces to contain the database name string. What do you think about this approach?

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -68,15 +75,30 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
-        const client: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
-        const listOfDatabases: DatabaseListResult = await client.databases.listByServer(this.resourceGroup, this.name);
+        if (!(this.connectionString.username && this.connectionString.password)) {
+            const { username, password } = await this.getCredentials();
+            if (username && password) {
+                await this.updateConnectionString(username, password);
```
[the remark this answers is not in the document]
neelip: That's true, this will cause an error before allowing the user to enter the credentials for Azure scenario. 
I like the first option better for now, given there is no credentials support for attached scenario. It is large piece of code to move to the ServerTreeItem, I would hold off on that for now.

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -68,16 +76,37 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+                    return database.name && !['azure_maintenance', 'azure_sys'].includes(database.name) ? new PostgresDatabaseTreeItem(this, database.name) : undefined;
+                },
+                (database) => database.name
+            );
+        } else {
+            const config = await getClientConfig(this, postgresDefaultDatabase);
+            const query = `SELECT datname FROM pg_catalog.pg_database WHERE datistemplate = false;`;
+            const queryResult = await runPostgresQuery(config, query);
+            const listOfDatabases = queryResult.rows;
+            return this.createTreeItemsWithErrorHandling(
```
EricJizbaMSFT, line 101: Can you find a way to get rid of some of this duplicate code with `createTreeItemsWithErrorHandling `

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -96,24 +125,26 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
         const databaseName = await ext.ui.showInputBox({
             placeHolder: "Database Name",
             prompt: "Enter the name of the database",
             validateInput: (name: string) => validateDatabaseName(name, getChildrenTask)
         });
+        const config = await getClientConfig(this, postgresDefaultDatabase);
         context.showCreatingTreeItem(databaseName);
-        const database: Database = { name: databaseName };
-        await client.databases.createOrUpdate(this.resourceGroup, this.name, databaseName, database);
+        await runPostgresQuery(config, `Create Database "${databaseName}";`);
```
EricJizbaMSFT, line 139: Use `wrapArgInQuotes`

src/postgres/tree/PostgresTableTreeItem.ts
```diff
@@ -51,8 +51,7 @@ export class PostgresTableTreeItem extends AzureParentTreeItem<ISubscriptionCont
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
-        const pool = new Pool(this.parent.clientConfig);
-        await pool.query(`Drop Table ${this.table.schema.name}."${this.label}";`);
+        await runPostgresQuery(this.parent.clientConfig, `Drop Table ${this.table.schema.name}."${this.table.name}";`);
```
EricJizbaMSFT, line 54: Use `wrapArgInQuotes`

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -68,16 +76,37 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
+            return this.createTreeItemsWithErrorHandling(
+                listOfDatabases,
+                'invalidPostgresServer',
+                (database) => {
+                    return database.name && !['azure_maintenance', 'azure_sys'].includes(database.name) ? new PostgresDatabaseTreeItem(this, database.name) : undefined;
+                },
+                (database) => database.name
+            );
+        } else {
+            const config = await getClientConfig(this, postgresDefaultDatabase);
```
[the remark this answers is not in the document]
neelip: It won't matter in the current case where we are displaying all the databases under an account with or without database name being provided in the connection string. Earlier the specific database name was needed to manage account and non-account scenarios.

src/postgres/tree/PostgresServerTreeItem.ts
neelip: I tried doing that but it would just create too many extra variables, given listOfDatabases and individual databases are both different types for azure and non azure scenario.

src/postgres/commands/copyConnectionString.ts
```diff
@@ -17,13 +15,17 @@ export async function copyConnectionString(context: IActionContext, node: Postgr
...
-    const port: string = String(nonNullProp(clientConfig, 'port'));
-    const connectionString = `postgres://${user}:${password}@${host}:${port}/'${node.databaseName}'`;
+    const parsedConnectionString = node.parent.connectionString;
+
+    if (!(parsedConnectionString.username && parsedConnectionString.password)) {
+        await node.parent.updateConnectionString();
+    }
+
+    const encodedDatabaseName = encodeURIComponent(node.databaseName);
+    let connectionString: string = parsedConnectionString.getEncodedConnectionString();
```
EricJizbaMSFT, line 27: The method is called `getEncodedConnectionString`, so it should encode the whole connection string including database name

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,57 @@
...
+    let connectionString: string = `postgres://`;
+    if (username && password) {
+        connectionString += `${username}:${password}@`;
+    }
+    connectionString += `${host}:${port}`;
+    return parsePostgresConnectionString(connectionString);
+}
+
+export function fixedEncodeURIComponent(component: string): string {
+    return encodeURIComponent(component).replace(/[!'()*]/g, escape);
```
EricJizbaMSFT, line 26: Why are you doing this? If you're going to do something weird (like modifying the Node.js default "encodeURIComponent"), you need to explain why and probably link to docs

src/postgres/postgresConnectionStrings.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export class ParsedPostgresConnectionString extends ParsedConnectionString {
+    public readonly hostName: string;
+    public readonly username: string | undefined;
+    public readonly password: string | undefined;
+    public readonly port: string;
+
+    constructor(connectionString: string, config: ConnectionOptions) {
+        super(connectionString, config.database?.replace(/^'(.*)'$/, '$1'));
```
EricJizbaMSFT, line 36: Please remove this regex quote stuff now that we're encoding the database name

src/postgres/commands/copyConnectionString.ts
```diff
@@ -17,13 +15,17 @@ export async function copyConnectionString(context: IActionContext, node: Postgr
...
+    await checkAuthentication(context, node);
 
-    const user: string = nonNullProp(clientConfig, 'user');
-    const password: string = nonNullProp(clientConfig, 'password');
-    const host: string = nonNullProp(clientConfig, 'host');
-    const port: string = String(nonNullProp(clientConfig, 'port'));
-    const connectionString = `postgres://${user}:${password}@${host}:${port}/'${node.databaseName}'`;
+    const parsedConnectionString = node.parent.connectionString;
+
+    if (!(parsedConnectionString.username && parsedConnectionString.password)) {
```
EricJizbaMSFT, line 22: I don't like the ordering of when/where you're calling `updateConnectionString`. You shouldn't need to call it here. There are two places you should call it:
1. Inside `PostgresServerTreeItem.getCredentials`. That way the connection string is always up to date with what you just retrieved from keytar and you won't have to call it in places like this.
1. Inside `enterPostgresCredentials.ts` - pretty self-explanatory

Also, I don't think you need an actual `updateConnectionString` method. Just make the two properties read-write and do something like this:
```
this.connectionString.username = username;
this.connectionString.password = password;
```

src/postgres/tree/PostgresServerTreeItem.ts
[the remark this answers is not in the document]
neelip: Actually, for this particular line, we need to use postgresDefaultDatabase; cannot use any other databases. To query all the databases listed under a server we can only do it using 'postgres' database.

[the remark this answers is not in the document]
neelip: I was going with the first option of displaying only the account with attaching connectionString, similar to cosmosdb behavior. This release will no longer support attaching database specific connectionString. We need the '/database' in the id for both the behaviors to work.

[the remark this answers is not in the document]
EricJizbaMSFT: > Actually, for this particular line, we need to use postgresDefaultDatabase; cannot use any other databases. To query all the databases listed under a server we can only do it using 'postgres' database.

We don't need to query the databases, we already know which database they want from the connection string. 

> We need the '/database' in the id for both the behaviors to work.

No we don't. We only need it for a specific edge case bug if the user attaches both at the same time. Mongo attached accounts have supported database-specific connection strings for a while, even with that bug, and no one has complained.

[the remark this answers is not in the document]
neelip: I see you are talking about adding database-specific code, I assume you meant modifying account-specific connection string code....
If the bug is not a common occurence then I don't see why we can't have both connection strings supported. The new commit supports database-specific connection strings as well.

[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I don't think the bug is common. You're welcome to fix it - I just ask that it's part of a separate issue/PR, because I'm not sure changing the id like that is the best way to fix it. Also, ideally you would fix the bug for both PostgreSQL and Mongo at the same time

src/postgres/postgresConnectionStrings.ts
EricJizbaMSFT: You still haven't answered why you need to escape extra characters?

src/postgres/getClientConfig.ts
```diff
@@ -5,26 +5,34 @@
...
-        const clientConfig: ClientConfig = { user: username, password, ssl, host, port: 5432, database: databaseName };
+    const ssl: ConnectionOptions = {
+        // Always provide the certificate since it is accepted even when SSL is disabled
+        // Certificate source: https://aka.ms/AA7wnvl
+        ca: BaltimoreCyberTrustRoot
+    };
+
+    if (username && password) {
+        const host = nonNullProp(treeItem.connectionString, 'hostName');
+        const clientConfig = { user: username, password: password, ssl, host, port: postgresDefaultPort, database: databaseName };
```
EricJizbaMSFT, line 32: Use `treeItem.connectionString.port` instead of hard-coding the default

src/postgres/postgresConnectionStrings.ts
neelip: I put a comment in the new commit. So, encodeURIComponent does not escape `A-Z a-z 0-9 - _ . ! ~ * ' ( )`, which include characters that are allowed for passwords.

EricJizbaMSFT: Yes I read your comment - why do you need to encode those characters?

PR #1663, opened 2020-08-20
Fix for Expected non-null or undefined 'username' error

### LINKED ISSUE #1661: Failed to view databases for attached local postgres server
Repro steps:
1. Run a postgres server locally. I used the `brew` method in [these docs](https://dataschool.com/learn-sql/how-to-start-a-postgresql-server-on-mac-os-x/)
1. Attached a local postgres server with connection string `postgres://localhost:5432`

Expected:
I can view databases

Actual:
Screen Shot 2020-08-19 at 11 17 41 AM

neelip:
Fixes #1661

### DIFF
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -38,14 +38,14 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
         } else {
             ssl = treeItem.connectionString.ssl;
         }
-        const clientConfig: ClientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
+        let clientConfig: ClientConfig;
         let client: Client;
         if (treeItem.azureName) {
+            clientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
             client = new Client(clientConfig);
         } else {
-            client = new Client({
-                connectionString: treeItem.connectionString.connectionString,
-            });
+            clientConfig = { connectionString: treeItem.connectionString.connectionString };
+            client = new Client(clientConfig);
         }
 
         // Ensure the client config is valid before returning
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -33,13 +33,18 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
         const host = nonNullProp(treeItem.connectionString, 'hostName');
         const port: number = treeItem.connectionString.port ? parseInt(treeItem.connectionString.port) : postgresDefaultPort;
         let ssl: boolean | ConnectionOptions | undefined;
-        if (treeItem.connectionString.hostName.endsWith(`.postgres.database.azure.com`)) {
+        if (treeItem.azureName) {
             ssl = sslAzure;
         } else {
             ssl = treeItem.connectionString.ssl;
         }
         const clientConfig: ClientConfig = { user: username, password: password, ssl, host, port, database: databaseName };
-        const client = new Client(clientConfig);
+        let client: Client;
+        if (treeItem.azureName) {
+            client = new Client(clientConfig);
+        } else {
+            client = new Client(treeItem.connectionString.connectionString);
+        }
 
         // Ensure the client config is valid before returning
         try {
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -12,54 +12,46 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
 import { PostgresServerTreeItem } from "./tree/PostgresServerTreeItem";
 
 export async function getClientConfig(treeItem: PostgresServerTreeItem, databaseName: string): Promise<ClientConfig> {
-    let username: string | undefined = treeItem.connectionString.username;
-    let password: string | undefined = treeItem.connectionString.password;
+    let clientConfig: ClientConfig;
+    if (treeItem.azureName) {
+        let username: string | undefined = treeItem.connectionString.username;
+        let password: string | undefined = treeItem.connectionString.password;
 
-    if (!(username && password)) {
-        const credentials = await treeItem.getCredentials();
-        if (credentials.username && credentials.password) {
-            username = credentials.username;
-            password = credentials.password;
+        if (!(username && password)) {
+            const credentials = await treeItem.getCredentials();
+            if (credentials.username && credentials.password) {
+                username = credentials.username;
+                password = credentials.password;
+            }
         }
-    }
 
-    const sslAzure: ConnectionOptions = {
-        // Always provide the certificate since it is accepted even when SSL is disabled
-        // Certificate source: https://aka.ms/AA7wnvl
-        ca: BaltimoreCyberTrustRoot
-    };
+        const sslAzure: ConnectionOptions = {
+            // Always provide the certificate since it is accepted even when SSL is disabled
+            // Certificate source: https://aka.ms/AA7wnvl
+            ca: BaltimoreCyberTrustRoot
+        };
 
-    if ((username && password) || username === 'postgres') {
-        const host = nonNullProp(treeItem.connectionString, 'hostName');
-        const port: number = treeItem.connectionString.port ? parseInt(treeItem.connectionString.port) : postgresDefaultPort;
-        let ssl: boolean | ConnectionOptions | undefined;
-        if (treeItem.azureName) {
-            ssl = sslAzure;
+        if ((username && password) || username === 'postgres') {
+            const host = nonNullProp(treeItem.connectionString, 'hostName');
+            const port: number = treeItem.connectionString.port ? parseInt(treeItem.connectionString.port) : postgresDefaultPort;
+            clientConfig = { user: username, password: password, ssl: sslAzure, host, port, database: databaseName };
         } else {
-            ssl = treeItem.connectionString.ssl;
-        }
-        let clientConfig: ClientConfig;
-        let client: Client;
-        if (treeItem.azureName) {
-            clientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
-            client = new Client(clientConfig);
-        } else {
-            clientConfig = { connectionString: treeItem.connectionString.connectionString };
-            client = new Client(clientConfig);
-        }
-
-        // Ensure the client config is valid before returning
-        try {
-            await client.connect();
-            return clientConfig;
-        } finally {
-            await client.end();
+            throw {
+                message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.'),
+                code: invalidCredentialsErrorType
+            };
         }
     } else {
-        throw {
-            message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.'),
-            code: invalidCredentialsErrorType
-        };
+        clientConfig = { connectionString: treeItem.connectionString.connectionString };
+    }
+
+    const client = new Client(clientConfig);
+    // Ensure the client config is valid before returning
+    try {
+        await client.connect();
+        return clientConfig;
+    } finally {
+        await client.end();
     }
 }
 
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -39,7 +39,14 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
             ssl = treeItem.connectionString.ssl;
         }
         const clientConfig: ClientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
-        const client = new Client(clientConfig);
+        let client: Client;
+        if (treeItem.azureName) {
+            client = new Client(clientConfig);
+        } else {
+            client = new Client({
+                connectionString: treeItem.connectionString.connectionString,
+            });
+        }
 
         // Ensure the client config is valid before returning
         try {
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -23,7 +23,7 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
         }
     }
 
-    const ssl: ConnectionOptions = {
+    const sslAzure: ConnectionOptions = {
         // Always provide the certificate since it is accepted even when SSL is disabled
         // Certificate source: https://aka.ms/AA7wnvl
         ca: BaltimoreCyberTrustRoot
@@ -32,8 +32,13 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
     if ((username && password) || username === 'postgres') {
         const host = nonNullProp(treeItem.connectionString, 'hostName');
         const port: number = treeItem.connectionString.port ? parseInt(treeItem.connectionString.port) : postgresDefaultPort;
-        const sslSupport = treeItem.connectionString.sslSupport;
-        const clientConfig: ClientConfig = { user: username, password: password, ssl: sslSupport ? ssl : sslSupport, host, port, database: databaseName };
+        let ssl: boolean | ConnectionOptions | undefined;
+        if (treeItem.connectionString.hostName.endsWith(`.postgres.database.azure.com`)) {
+            ssl = sslAzure;
+        } else {
+            ssl = treeItem.connectionString.ssl;
+        }
+        const clientConfig: ClientConfig = { user: username, password: password, ssl, host, port, database: databaseName };
         const client = new Client(clientConfig);
 
         // Ensure the client config is valid before returning
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -14,16 +14,21 @@ import { PostgresServerTreeItem } from "./tree/PostgresServerTreeItem";
 export async function getClientConfig(treeItem: PostgresServerTreeItem, databaseName: string): Promise<ClientConfig> {
     let clientConfig: ClientConfig;
     if (treeItem.azureName) {
-        const username: string | undefined = treeItem.connectionString.username;
-        const password: string | undefined = treeItem.connectionString.password;
+        let username: string | undefined = treeItem.connectionString.username;
+        let password: string | undefined = treeItem.connectionString.password;
+
+        if (!(username && password)) {
+            const credentials = await treeItem.getCredentials();
+            username = nonNullProp(credentials, 'username');
+            password = nonNullProp(credentials, 'password');
+        }
 
         const sslAzure: ConnectionOptions = {
             // Always provide the certificate since it is accepted even when SSL is disabled
             // Certificate source: https://aka.ms/AA7wnvl
             ca: BaltimoreCyberTrustRoot
         };
-
-        if ((username && password) || username === 'postgres') {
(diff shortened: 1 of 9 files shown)


EricJizbaMSFT:
I'm not sure you understood the issue... local databases don't need credentials to view databases. A user should be able to connect with `postgres://localhost:5432` and it should "just work"

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -82,6 +82,7 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
             const listOfDatabases: DatabaseListResult = await client.databases.listByServer(nonNullProp(this, 'resourceGroup'), nonNullProp(this, 'azureName'));
             dbNames = listOfDatabases.map(db => db?.name);
         } else if (this.connectionString.databaseName) {
+            await getClientConfig(this, this.connectionString.databaseName);
```
neelip, line 85: Without this getClientConfig check, the attached database-specific account will not throw "Must enter credentials..." but would direct to a prompt to enter credentials. The enter credentials prompt won't work with the current setup.

src/postgres/postgresConnectionStrings.ts
```diff
@@ -35,13 +40,19 @@ export class ParsedPostgresConnectionString extends ParsedConnectionString {
...
     constructor(connectionString: string, config: ConnectionOptions) {
         super(connectionString, config.database ? config.database : undefined);
         this.hostName = nonNullProp(config, 'host');
         this.port = config.port ? config.port : `${postgresDefaultPort}`;
         this.username = nonNullProp(config, 'user');
         this.password = nonNullProp(config, 'password');
+        if (config.ssl === false) {
+            this.sslSupport = false;
+        } else {
+            this.sslSupport = true;
```
neelip, line 54: This is assuming ssl is supported in all other cases --> config.ssl = string | undefined | true.

neelip:
> I'm not sure you understood the issue... local servers don't need credentials to view databases. A user should be able to connect with `postgres://localhost:5432` and it should "just work"

You need to use something like this for the localhost connection to work (assuming your localhost configuration does not support ssl connections, otherwise leave out the ssl part) --> `postgres://localhost:5432?ssl=0`

`pg-connection-string` library only considers `ssl=0` as `false`.

src/postgres/getClientConfig.ts
```diff
@@ -12,25 +12,33 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
...
         ca: BaltimoreCyberTrustRoot
     };
 
-    if (username && password) {
+    if ((username && password) || username === 'postgres') {
         const host = nonNullProp(treeItem.connectionString, 'hostName');
         const port: number = treeItem.connectionString.port ? parseInt(treeItem.connectionString.port) : postgresDefaultPort;
-        const clientConfig = { user: username, password: password, ssl, host, port, database: databaseName };
+        let ssl: boolean | ConnectionOptions | undefined;
+        if (treeItem.connectionString.hostName.endsWith(`.postgres.database.azure.com`)) {
```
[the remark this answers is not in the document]
neelip: The problem is that for attaching azure accounts we need to use ssl = "BaltimoreCyber...". We can include this in the connection string but that would include a larger code change as we need to store the certificate and have to figure out an appropriate way to include that value in the connection string so that it would work with other postgres platforms. I would prefer that we handle it this way for this release and create an issue for this. I am guessing the the ending of the host name ".com" would differ for different countries. How about we check if the name contains the rest of the substring? 
What do you think?

src/postgres/postgresConnectionStrings.ts
```diff
@@ -32,16 +37,21 @@ export function fixedEncodeURIComponent(component: string): string {
 
 export class ParsedPostgresConnectionString extends ParsedConnectionString {
     public readonly hostName: string;
-    public username: string | undefined;
-    public password: string | undefined;
+    public username: string;
```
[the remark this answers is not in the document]
neelip: Even for localhost,  username is required but password can remain empty. And the parse() function assumes values for username and password. And same goes for ClientConfig. 

`export interface ConnectionOptions {
  host: string | null
  password?: string
  user?: string
  port?: string | null
  database: string | null | undefined
  client_encoding?: string
  ssl?: boolean | string`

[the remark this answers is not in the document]
EricJizbaMSFT: In the type that you copied, "user" has a question mark, meaning it's optional

For more detail on why these are optional, I took a look at these docs: https://www.postgresql.org/docs/10/libpq-connect.html

- password: "Password to be used if the server demands password authentication."
    - The key word here is "if". We don't know which cases require a password and which don't, so we need to support both (Plus - as mentioned before - localhost is the most important attached scenario and it doesn't need a password)
- user: "PostgreSQL user name to connect as. Defaults to be the same as the operating system name of the user running the application."
    - Meaning "postgres" is _not_ the default in all scenarios. Also means we don't need to give a username because the server will determine a default by itself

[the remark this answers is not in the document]
neelip: That makes sense. So we are assuming then that users will provide both the ssl and credentials with the connection string?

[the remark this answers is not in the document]
EricJizbaMSFT: Yes - it's up to the user to provide (or not provide) ssl and credentials

src/postgres/getClientConfig.ts
```diff
@@ -17,21 +17,36 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
...
+        } else {
+            ssl = treeItem.connectionString.ssl;
+        }
+        const clientConfig: ClientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
+        let client: Client;
+        if (treeItem.azureName) {
+            client = new Client(clientConfig);
+        } else {
+            client = new Client({
+                connectionString: treeItem.connectionString.connectionString,
```
EricJizbaMSFT, line 47: So this line is good - I like directly using the connectionString for the attached scenario. However, the problem is that the client created here is never returned or used except for a few lines down. That could cause problems if the `ClientConfig` returned by this function is different than what's specified in the connection string. I think the best solution is to refactor `getClientConfig` to `createPgClient`. Something like this:

```
export async function createPgClient(treeItem: PostgresServerTreeItem, databaseName: string): Promise<Client> {
    if (treeItem.azureName) {
        // all the stuff to create clientConfig for the Azure case
        const clientConfig: ClientConfig = { user: username, password: password, ssl: ssl, host, port, database: databaseName };
        return new Client(clientConfig);
    } else {
        return new Client(treeItem.connectionString.connectionString);
    }
}
```

That way, it's the "pg" module's job to create a ClientConfig from the connection string, not us.

neelip: Passing Client object will cause problems as Client can only be connected once. Also, for functions and tables the same client will be passed down. We can use Pool which will allow for multiple client connections but it will require a lot more code change and each client needs to be released to make sure there is no lag and client availability is guaranteed. 

We can still return ClientConfig for both cases, with the exception that ClientConfig for attach accounts scenario would only contain `connectionString:`.

EricJizbaMSFT: > We can still return ClientConfig for both cases, with the exception that ClientConfig for attach accounts scenario would only contain connectionString:.

Yeah that'll work. I still think the above proposed solution is better (and I'll explain why below), but I agree this is sufficient for this PR and would require fewer changes.

> Passing Client object will cause problems as Client can only be connected once.

It's a bit excessive to call `client.connect` every time we call `getClientConfig`. We should only need to do it once per database, probably as a part of `checkAuthentication`.

> Also, for functions and tables the same client will be passed down.

I don't think `clientConfig` should be passed down in the first place. It would be better if the child tree items (like functions/tables) called something like `this.parent.getClientConfig()` to make sure they always get the latest clientConfig. Today, the parent might've changed clientConfig with a different username/password and the children may not get that unless the user refreshes the tree.

My point is that the two obstacles you mentioned should probably be refactored eventually anyways.

src/postgres/getClientConfig.ts
```diff
@@ -17,21 +17,36 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
...
-        const client = new Client(clientConfig);
+        let ssl: boolean | ConnectionOptions | undefined;
+        if (treeItem.azureName) {
+            ssl = sslAzure;
+        } else {
+            ssl = treeItem.connectionString.ssl;
+        }
+        let clientConfig: ClientConfig;
+        let client: Client;
+        if (treeItem.azureName) {
```
EricJizbaMSFT, line 43: Even though I'm fine with you not doing `createPgClient` for this PR, I still want you to take the other part of that suggestion, which more clearly separated out the azure vs. attached scenarios. Aka something like this:
```typescript
export async function getClientConfig(treeItem: PostgresServerTreeItem, databaseName: string): Promise<ClientConfig> {
    let clientConfig: ClientConfig;
    if (treeItem.azureName) {
        // all the stuff to create clientConfig for the Azure case
    } else {
        clientConfig = { connectionString: treeItem.connectionString.connectionString };
    }

    // Ensure the client config is valid before returning
    const client = new Client(clientConfig);
    try {
        await client.connect();
        return clientConfig;
    } finally {
        await client.end();
    }
}
```

src/postgres/tree/PostgresServerTreeItem.ts
EricJizbaMSFT: I'm confused by this comment - is it still accurate? We don't want "Must enter credentials..." to show up because they're optional for attached scenarios

src/postgres/postgresConnectionStrings.ts
```diff
@@ -27,13 +28,18 @@ export class ParsedPostgresConnectionString extends ParsedConnectionString {
     public username: string | undefined;
     public password: string | undefined;
     public readonly port: string;
+    public readonly ssl: boolean | ConnectionOptions | undefined;
```
EricJizbaMSFT, line 31: Do we still need these ssl changes? Since the azure scenario always uses `BaltimoreCyberTrustRoot` and the attached scenario always uses the connectionString itself

neelip:
https://github.com/microsoft/vscode-cosmosdb/pull/1663#discussion_r477472639
As connectionString is a ServerTreeItem property, databaseName would only be present for the attached scenario. But this modification is unnecessary change for this particular PR, I will remove it here and include it in the other one.

src/postgres/tree/PostgresServerTreeItem.ts
neelip: As connectionString is a ServerTreeItem property, databaseName would only be present for the attached scenario. But this modification is unnecessary change for this particular PR, I will remove it here and include it in the other one.

src/postgres/getClientConfig.ts
neelip: Yes I agree with you on that, the way we are connecting to client and getting clientConfig is not systematic and there could gaps in this process. I will create an issue for it for next release.

src/postgres/postgresConnectionStrings.ts
neelip: > Make sure to test out the localhost scenario - I'm still seeing "Must enter credentials" when I should be seeing the databases _without_ needing to enter credentials
> Screen Shot 2020-08-26 at 10 37 45 AM

I am able to connect, you just have to include the username in the connection string for it to work.

neelip:
> Make sure to test out the localhost scenario - I'm still seeing "Must enter credentials" when I should be seeing the databases _without_ needing to enter credentials
> Screen Shot 2020-08-26 at 10 37 45 AM

I am able to connect, you just have to include the username in the connection string for it to work.

src/postgres/postgresConnectionStrings.ts
neelip: That's true, we don't need ssl field.

EricJizbaMSFT:
> I am able to connect, you just have to include the username in the connection string for it to work.

People don't typically add a username to a localhost connection string, so it should work without that

neelip:
> > I am able to connect, you just have to include the username in the connection string for it to work.
> 
> People don't typically add a username to a localhost connection string, so it should work without that

The changes in the latest commits should resolve this issue. The connection string `postgres://localhost:5432` should work as is and there shouldn't be a `Must enter credentials to connect to server` either.

src/postgres/getClientConfig.ts
```diff
@@ -12,39 +12,38 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
...
+                message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.'),
+                code: invalidCredentialsErrorType
+            };
         }
     } else {
-        throw {
-            message: localize('mustEnterCredentials', 'Must enter credentials to connect to server.'),
-            code: invalidCredentialsErrorType
-        };
+        clientConfig = { connectionString: treeItem.connectionString.connectionString };
```
EricJizbaMSFT, line 37: I'm getting this now:
Screen Shot 2020-08-28 at 10 51 47 AM

Seems like "pg" has a defaults object, which uses this:
```
  // name of database to connect
  database: process.platform === 'win32' ? process.env.USERNAME : process.env.USER,
```

I'm not sure why they default to the username, and seems like it [already changed](https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/defaults.js#L11) in the latest version of the "pg" package, but I think the fix is for us to change that `defaults.database` to "postgres" (or eventually try out the new version of the "pg" package)

src/postgres/getClientConfig.ts
```diff
@@ -12,39 +12,38 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
...
+            ca: BaltimoreCyberTrustRoot
+        };
 
-        // Ensure the client config is valid before returning
-        try {
-            await client.connect();
-            return clientConfig;
-        } finally {
-            await client.end();
+        if ((username && password) || username === 'postgres') {
```
EricJizbaMSFT, line 26: What's `username === 'postgres'` for? Can we get rid of that?

src/postgres/postgresConnectionStrings.ts
```diff
@@ -27,13 +27,14 @@ export class ParsedPostgresConnectionString extends ParsedConnectionString {
...
     public readonly port: string;
+    public readonly options: string;
 
     constructor(connectionString: string, config: ConnectionOptions) {
         super(connectionString, config.database ? config.database : undefined);
         this.hostName = nonNullProp(config, 'host');
         this.port = config.port ? config.port : `${postgresDefaultPort}`;
-        this.username = config.user;
-        this.password = config.password;
+        this.username = nonNullProp(config, 'user');
```
EricJizbaMSFT, line 36: These shouldn't be using `nonNullProp` since user/password are optional. I looked into it and seems like we're just getting lucky that `parse` is setting it to an empty string instead of `undefined`

src/postgres/postgresConnectionStrings.ts
```diff
@@ -27,13 +27,14 @@ export class ParsedPostgresConnectionString extends ParsedConnectionString {
     public username: string | undefined;
     public password: string | undefined;
     public readonly port: string;
+    public readonly options: string;
```
EricJizbaMSFT, line 30: Remove `options` since it's unused

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -154,8 +155,10 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
                     }
                 }
             }
-            this.connectionString.username = username;
-            this.connectionString.password = password;
+            if (username && password) {
```
EricJizbaMSFT, line 158: This check shouldn't be necessary

src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -77,7 +77,8 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
 
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
         let dbNames: (string | undefined)[];
-        if (this._azureId) {
+        if (this.azureName) {
+            await this.getCredentials();
```
EricJizbaMSFT, line 81: How come you added getCredentials here?

neelip: Adding `getCredentials()` here would make sure that the `connectionString.username` and `connectionString.password` are initialized. Because the connectionString is not stored in persistedId like username every time loadMore...() runs the connectionString is reset and the username and password are undefined again. 

With `getCredentials()` here username and password values of the conenctionString are available to use in the `getClientConfig()` ( do not have to call `getCredentials()` again in `getClientConfig()`)

EricJizbaMSFT: I see what you're trying to fix, but I'm worried about two things:
1. There's no guarantee "loadMoreChildrenImpl" is called before connectionString is accessed, so we might run into timing issues
1. getCredentials accesses keytar every time its called, and we want to minimize the number of times we use keytar. One reason is that operating systems will prompt users something like "Do you authorize VS Code to access keychain?". Users can usually click "Always Allow", but we don't want to bother them if they don't want to click that.

I think I prefer if fixing this issue is done separately, as I don't think it'll be a one-line fix like you have here

src/postgres/getClientConfig.ts
neelip: Updating "pg" package would cause the same problem... Here is the latest version:
In defaults, `database: undefined,`
In connection-parameters, `if (this.database === undefined) {  this.database = this.user }`
That said the latest version is more easier to handle. What is the best place for `default.database = 'postgres'` to go, I am not sure where to place it in the general or 'utils' folder?

src/postgres/getClientConfig.ts
neelip: Yes, this is not necessary now

src/postgres/tree/PostgresServerTreeItem.ts
neelip: We still need this check for azure accounts as invalidCredentialsErrorType would lead to `enterPostgresCredentials()` when trying to load contents of a database. This for only when the account is created in the portal and not in the extension.

src/postgres/getClientConfig.ts
EricJizbaMSFT: > What is the best place for default.database = 'postgres' to go, I am not sure where to place it in the general or 'utils' folder?

Well we only need to do it once per session of VS Code. So maybe just in the extension.ts/activate method. Or in `registerPostgresCommands` if you want it in a postgres-specific place

src/postgres/tree/PostgresServerTreeItem.ts
EricJizbaMSFT: Can you explain this code path? I'm still a bit confused. It seems like if "username" and "password" are undefined (aka not persisted or in keytar), then `this.connectionString.username` and `this.connectionString.password` would also be undefined and so it wouldn't matter if you have this check or not

neelip: You're right, the check here is not necessary

PR #1664, opened 2020-08-20
Postgres api support for App Service connections

neelip:


src/postgres/getClientConfig.ts
```diff
@@ -12,13 +12,15 @@ import { invalidCredentialsErrorType } from "./tree/PostgresDatabaseTreeItem";
 import { PostgresServerTreeItem } from "./tree/PostgresServerTreeItem";
 
 export async function getClientConfig(treeItem: PostgresServerTreeItem, databaseName: string): Promise<ClientConfig> {
-    let username: string | undefined = treeItem.connectionString.username;
-    let password: string | undefined = treeItem.connectionString.password;
+    let username: string = treeItem.connectionString.username;
+    let password: string = treeItem.connectionString.password;
```
EricJizbaMSFT, line 16: Please revert all these username/password changes - I assume it's something that got accidentally pulled in from https://github.com/microsoft/vscode-cosmosdb/pull/1663?

src/vscode-cosmosdb.api.d.ts
```diff
@@ -59,7 +65,7 @@ export interface DatabaseTreeItem extends DatabaseAccountTreeItem {
 
 export type CosmosDBResourceType = 'DatabaseAccount' | 'Database';
 
-export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table';
+export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table' | 'Postgres';
```
EricJizbaMSFT, line 68: So all of the naming in this file is still Cosmos DB-specific. Are we sure that it will work as-is for PostgreSQL? For example, Postgres is not a "_Cosmos DB_ API Type".

src/vscode-cosmosdb.api.d.ts
```diff
@@ -51,6 +51,12 @@ export interface DatabaseAccountTreeItem extends CosmosDBTreeItem {
         masterKey: string;
         documentEndpoint: string;
     }
+
+    postgresData?: {
```
EricJizbaMSFT, line 55: Whenever you change the api, you also need to change the version:
https://github.com/microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L175

src/commands/api/DatabaseAccountTreeItemInternal.ts
```diff
@@ -54,14 +63,34 @@ export class DatabaseAccountTreeItemInternal implements DatabaseAccountTreeItem
...
+            // tslint:disable-next-line: no-any
+            let apiType: any;
+            if (this._parsedCS instanceof ParsedMongoConnectionString) {
+                apiType = API.MongoDB;
+            } else if (this._parsedCS instanceof ParsedPostgresConnectionString) {
+                apiType = API.Postgres;
+            } else {
+                apiType = API.Core;
+            }
             this._accountNode = await ext.attachedAccountsNode.attachConnectionString(this.connectionString, apiType);
```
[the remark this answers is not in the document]
neelip: You are right, I was using the other format when there wasn't a connection string property on PostgresServerTreeItem nodes.

src/commands/api/apiCache.ts
```diff
@@ -25,8 +26,13 @@ export function removeTreeItemFromCache(expected: ParsedConnectionString): void
     if (!expected.databaseName) {
         // If parsedCS represents an account, remove the account and any databases that match that account
         for (const [key, value] of sessionCache.entries()) {
-            const actual = new ParsedMongoConnectionString(value.connectionString, value.hostName, value.port, undefined);
-            if (actual.accountId === expected.accountId) {
+            let actual: ParsedConnectionString | undefined;
+            if (expected instanceof ParsedPostgresConnectionString) {
+                actual = parsePostgresConnectionString(value.connectionString);
+            } else if (expected instanceof ParsedMongoConnectionString) {
```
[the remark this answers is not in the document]
neelip: Not sure why. I think DocDB should be treated the same way as it can have both account-specific and database-specific connection string.

src/vscode-cosmosdb.api.d.ts
neelip: Would it be reasonable to make it 1.2.0 or 1.1.1?

src/vscode-cosmosdb.api.d.ts
neelip: The checked the properties mentioned in this api file are very specifically created for the api, don't have direct connection to the rest of the Databases code.

src/vscode-cosmosdb.api.d.ts
EricJizbaMSFT: Per https://semver.org/

> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards compatible manner, and
> 3. PATCH version when you make backwards compatible bug fixes.

This is definitely a functional change, not just a bug fix. So it's either 1.2.0 or 2.0.0, depending if we have breaking changes in the api

neelip: Not sure if it is incompatible, doesn't really break older App Service versions. I would change it to 1.2.0 for now and update before release if needed.

src/commands/api/findTreeItem.ts
```diff
@@ -114,3 +114,44 @@ async function searchDbAccounts(dbAccounts: AzExtTreeItem[], expected: ParsedCon
...
+            // We found the right account - just not the db. In this case we can still 'reveal' the account
+            return new DatabaseTreeItemInternal(expected, expected.databaseName, dbAccount);
+        }
+        return new DatabaseAccountTreeItemInternal(expected, dbAccount);
+    }
+
+    return undefined;
+}
+
+async function searchPostgresServerAccount(dbAccount: PostgresServerTreeItem, expected: ParsedConnectionString, actual: ParsedConnectionString, context: IActionContext): Promise<DatabaseAccountTreeItem | DatabaseTreeItem | undefined> {
```
EricJizbaMSFT, line 137: > Code Changes to findTreeItem.ts to fix funtionality issues

Can you explain what the functionality issues were? I'm just not sure why we need a separate "searchPostgresServerAccount"

src/vscode-cosmosdb.api.d.ts
EricJizbaMSFT: Well I think this release is the right time to change the naming. It was accurate before "Postgres" was added - and now it's not. Better to clump the naming changes and functionality changes together when it comes to an API

src/commands/api/findTreeItem.ts
neelip: The current search function has an odd structure, where when the `database` is not found in the attachedAccounts list however if the `account` is found, it will immediately return a value rather than undefined. 
`// We found the right account - just not the db. In this case we can still 'reveal' the account`
`return new DatabaseTreeItemInternal(expected, expected.databaseName, dbAccount);`
With undefined returned the call will move down to searching azureSubscriptions instead and usually the right account is found in this step. 

We can have one single function for all but it requires changing existing code.

EricJizbaMSFT: Remember the attached scenario isn't really for Azure users. We don't expect someone to attach an Azure server when they can already see it in their subscriptions (I know you're doing it, but that's really only for testing purposes). So it's actually unlikely that "the right account is found" by searching the azure subscriptions.

neelip: That's true I forgot about that. In that case sticking with the old format for postgres as well should work.

src/extension.ts
```diff
@@ -172,7 +172,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         findTreeItem,
         pickTreeItem,
         revealTreeItem,
-        apiVersion: '1.1.0'
+        apiVersion: '2.0.0'
```
EricJizbaMSFT, line 175: I think you can still do 1.2.0. The only stuff you changed was interface and type names, but those disappear when typescript gets compiled to javascript anyways.

PR #1672, opened 2020-08-28
App Service Postgres connections "open in portal" support for Firewall rules

neelip:


src/commands/api/DatabaseAccountTreeItemInternal.ts
```diff
@@ -58,6 +59,12 @@ export class DatabaseAccountTreeItemInternal implements DatabaseAccountTreeItem
         ext.treeView.reveal(await this.getAccountNode());
     }
 
+    public async openInPortal(resourceID: string): Promise<void> {
+        if (this._accountNode) {
```
nturinski, line 63: If you don't have the accountNode, I think you should call getAccountNode then rather than just silently failing.

src/vscode-cosmosdb.api.d.ts
```diff
@@ -51,6 +51,8 @@ export interface DatabaseAccountTreeItem extends CosmosDBTreeItem {
         masterKey: string;
         documentEndpoint: string;
     }
+
+    openInPortal(resourceID: string): Promise<void>;
```
EricJizbaMSFT, line 55: How about we add `accountId` to `azureData` and just let the app service repo take care of the actual openInPortal logic. I feel like that's a simpler and more generic solution, both important qualities for an API

neelip: This might be better option. I will just add the accountId in Databases and openInPortal will be through app service.

src/vscode-cosmosdb.api.d.ts
```diff
@@ -45,6 +45,7 @@ export interface DatabaseAccountTreeItem extends AzureDatabasesTreeItem {
      */
     azureData?: {
         accountName: string;
+        accountId?: string;
```
EricJizbaMSFT, line 48: I would add accountId in the cosmos case, too - so that we don't have to make this optional and add docs explaining why it only works for postgres

EricJizbaMSFT (approved):

PR #1680, opened 2020-09-03
Fixes Postgres server name issue with strings containing only numbers

### LINKED ISSUE #1593: There is no validation when inputting pure numbers for Administrator Username
**OS:** Win10
**Build Version:** 20200626.4
Does this occur consistently? Yes 

**Repro steps:**

1. F1 --> Type and select "Azure Databases: Create Server..." --> Select "PostgreSQL" --> Input "1234567"  for Administrator Username.
2. Check the result.

**Expect:**
Create a PostgreSQL server without any error.

**Actual:**
There is an error "'123456' is not a valid name because it contains invalid characters".
image

Action: azureDatabases.createServer
Error type: Error
Error Message: '123456' is not a valid name because it contains invalid characters.

Version: 0.13.1-alpha
OS: win32
OS Release: 10.0.18363
Product: Visual Studio Code - Insiders
Product Version: 1.47.0-insider
Language: en

<details>
<summary>Call Stack</summary>

```
Object.<anonymous> extension.bundle.js:535:475109
rejected extension.bundle.js:535:474536
processTicksAndRejections task_queues.js:93:5
```

</details>

neelip:


src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -42,6 +42,8 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
         return localize('serverNameLengthCheck', 'The name must be between {0} and {1} characters.', min, max);
     } else if (!(/^[a-z0-9-]+$/).test(name)) {
         return localize('serverNameCharacterCheck', 'Server name must only contain lowercase letters, numbers, and hyphens.');
+    } else if ((/^[0-9]+$/).test(name)) {
```
EricJizbaMSFT, line 45: I think the issue was talking about the admin username, not the server name

neelip: That's true, created another PR to fix username. I noticed that serverName seems to have a similar problem with validation.

EricJizbaMSFT: So "12345" is invalid but "123-45" is valid? That seems like a bug, not intentional design. Did you find any docs on this? If not, I think I lean towards "wont-fix" and calling this Azure's problem

neelip: It is odd. I didn't find any docs but when I tried to create using only numbers on the portal, it is not allowing. 
image
image

neelip: Actually, same thing goes for username. The invalid case with just numbers is not mentioned on the portal, but it doesn't allow to create with it.

EricJizbaMSFT: Right, but the portal's message doesn't saying anything about "it can't be all numbers", so I think it's a bug in the portal.

I just tried creating a server from our extension with all numbers and it worked.

neelip: I agree, it must be in the portal, creating through server works.

PR #1690, opened 2020-09-09
Fixes invalid characters error for valid postgres inputs

### LINKED ISSUE #1689: Fail to create a PostgreSQL server with a **valid** Administrator username on the Mac and Windows
Build Version: 20200831.4
Fail to create a PostgreSQL server with a **valid** Administrator username on the Mac and Windows.

_Originally posted by @v-mengwe in https://github.com/microsoft/vscode-cosmosdb/issues/1593#issuecomment-685441045_

neelip:
Fixes #1689

### DIFF
--- a/src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
+++ b/src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
@@ -8,9 +8,13 @@ import { IAzureDBWizardContext } from "../../../tree/IAzureDBWizardContext";
 
 export interface IPostgresServerWizardContext extends IAzureDBWizardContext {
 
-    // username
+    /**
+     * Username without server, i.e. "user1"
+     */
     shortUserName?: string;
-    // username@serverName
+    /**
+     * Username with server, i.e. "user1@server1"
+     */
     longUserName?: string;
     adminPassword?: string;
 
--- a/src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
+++ b/src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
@@ -8,7 +8,10 @@ import { IAzureDBWizardContext } from "../../../tree/IAzureDBWizardContext";
 
 export interface IPostgresServerWizardContext extends IAzureDBWizardContext {
 
-    adminUser?: string;
+    // username
+    shortUserName?: string;
+    // username@serverName
+    longUserName?: string;
     adminPassword?: string;
 
     addFirewall?: boolean;
--- a/src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
+++ b/src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
@@ -31,7 +31,7 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresSe
                 const options = {
                     location: locationName,
                     properties: {
-                        administratorLogin: user,
+                        administratorLogin: user.split('@')[0],
                         administratorLoginPassword: password,
                         sslEnforcement: "Enabled",
                         createMode: "Default"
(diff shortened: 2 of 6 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -90,7 +90,11 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             ext.outputChannel.appendLog(createMessage);
             const server = nonNullProp(wizardContext, 'server');
             const host = nonNullProp(server, 'fullyQualifiedDomainName');
-            const username: string = nonNullProp(wizardContext, 'adminUser');
+            let username: string = nonNullProp(wizardContext, 'adminUser');
```
EricJizbaMSFT, line 93: You're trying to fix a symptom of the problem, rather than the root cause. This is not the first time you've moved these few lines of code because they weren't executed at the right time. I still don't think it's right because I believe `PostgresServerSetCredentialsStep` needs the username _with_ the suffix.

Try to come up with a solution that addresses the root cause - something that makes these kinds of bugs less likely to happen in the future. It will probably require some sort of renaming or refactoring.

neelip: That's true. `createAzureClient` in the PostgresServerCreatStep is the only place we need to use username without the suffix, and anything after that will require the full username including the suffix.

src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
```diff
@@ -8,7 +8,10 @@ import { IAzureDBWizardContext } from "../../../tree/IAzureDBWizardContext";
 
 export interface IPostgresServerWizardContext extends IAzureDBWizardContext {
 
-    adminUser?: string;
+    // username
```
EricJizbaMSFT, line 11: I think we can be a little more descriptive than this 😝 How about 
> Username without server, i.e. "user1"

and

> UserName with server, i.e. "user1@server1"

src/postgres/commands/createPostgresServer/IPostgresServerWizardContext.ts
```diff
@@ -8,7 +8,10 @@ import { IAzureDBWizardContext } from "../../../tree/IAzureDBWizardContext";
...
+    shortUserName?: string;
```
EricJizbaMSFT, line 12: There's a standard way to write comments for properties and it looks like this:
```
/**
 * Username without server, i.e. "user1"
 */
shortUserName?: string;
```

One reason we do it this way is so that you can see the comment when you hover over the property or during autocomplete
Screen Shot 2020-09-09 at 2 13 59 PM

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -31,7 +31,7 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresSe
                 const options = {
                     location: locationName,
                     properties: {
-                        administratorLogin: user,
+                        administratorLogin: user.split('@')[0],
```
EricJizbaMSFT, line 34: No need to split anymore

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -19,7 +19,7 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresSe
         const locationName = nonNullProp(nonNullProp(wizardContext, 'location'), 'name');
         const rgName: string = nonNullProp(nonNullProp(wizardContext, 'resourceGroup'), 'name');
         const newServerName = nonNullProp(wizardContext, 'newServerName');
-        const user: string = nonNullProp(wizardContext, 'adminUser');
+        const user: string = nonNullProp(wizardContext, 'shortUserName');
```
EricJizbaMSFT, line 22: I would get rid of the `user` variable since it's only used once and use `nonNullProp(wizardContext, 'shortUserName')` directly below

Alternatively, rename "user" to "shortUserName" so that we avoid the same confusion we had before

EricJizbaMSFT (approved):

PR #1697, opened 2020-09-09
Fix Postgres client connection with connection string

### LINKED ISSUE #1696: The functions of database display incorrectly for attached PostgreSQL server
**OS:** All
**Build Version:** 20200908.6

**Repro steps:**

1. Attach PostgreSQL server without database (remove the database name from the connection string).
2. Check the Functions of database and postgres.

**Expect:**
The functions of database display correctly.

**Actual:**
The functions of database display incorrectly.
image

neelip:
Fixes #1696 
To fix missing databaseName value for Client connections with connectionString.

### DIFF
--- a/src/postgres/getClientConfig.ts
+++ b/src/postgres/getClientConfig.ts
@@ -13,10 +13,10 @@ import { PostgresServerTreeItem } from "./tree/PostgresServerTreeItem";
 
 export async function getClientConfig(treeItem: PostgresServerTreeItem, databaseName: string): Promise<ClientConfig> {
     let clientConfig: ClientConfig;
-    const connectionString = await treeItem.getFullConnectionString();
+    const parsedCS = await treeItem.getFullConnectionString();
     if (treeItem.azureName) {
-        const username: string | undefined = connectionString.username;
-        const password: string | undefined = connectionString.password;
+        const username: string | undefined = parsedCS.username;
+        const password: string | undefined = parsedCS.password;
 
         const sslAzure: ConnectionOptions = {
             // Always provide the certificate since it is accepted even when SSL is disabled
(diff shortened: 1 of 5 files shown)


src/postgres/getClientConfig.ts
```diff
@@ -34,7 +34,11 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
             };
         }
     } else {
-        clientConfig = { connectionString: connectionString.connectionString };
+        let connectionString = parsedCS.connectionString;
+        if (!parsedCS.databaseName) {
+            connectionString += databaseName;
```
EricJizbaMSFT, line 39: Unfortunately it's not this simple - we have to really think about edge cases with the attached scenario because users could enter any type of connection string. For example, users could have a query in the uri, meaning you'd be appending to the query instead of the pathname and it wouldn't work. Also they may or may not have a trailing slash already in the uri.

Check out Node.js's built-in "URL" class. You should be able to do something like this, which is more robust:
```
url = new URL(connectionString)
url.pathname = databaseName;
url.toString()
```

Also add a few unit tests, similar to what we did for mongo: https://github.com/microsoft/vscode-cosmosdb/blob/24843d11283ad5ed820db90fec47e84f0f73b840/test/mongoConnectionStrings.test.ts#L77

test/postgresConnectionStrings.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import * as assert from 'assert';
+import { addDatabaseToConnectionString } from '../src/postgres/postgresConnectionStrings';
+
+function testAddDatabaseToConectionString(connectionString: string, databaseName: string, expectedConnectionString: string | undefined): void {
+    const modifiedConnectionString = addDatabaseToConnectionString(connectionString, databaseName);
+    assert.equal(modifiedConnectionString, expectedConnectionString);
+}
+
+suite(`postgresCollectionStrings`, () => {
```
EricJizbaMSFT, line 14: I assume this should be postgresConnectionStrings instead of co**ll**ection

EricJizbaMSFT (approved):

test/postgresConnectionStrings.test.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { addDatabaseToConnectionString } from '../src/postgres/postgresConnectionStrings';
```
EricJizbaMSFT, line 7: I think the build is failing because you have to import this from `extension.bundle` for tests to work when webpacked (which is how the build machine runs tests)

EricJizbaMSFT: Also need to export `addDatabaseToConnectionString` here: 
https://github.com/microsoft/vscode-cosmosdb/blob/master/extension.bundle.ts

PR #1698, opened 2020-09-09
Enable display of storeProcedures for attached accounts

### LINKED ISSUE #1695: Display all subitems of functions and stored procedures under the attached database
**OS:** All
**Build Version:** 20200908.6

**Repro steps:**

1. Attach PostgreSQL server with a database (add the database name next to the connection string).
2. Check the Functions of database.

**Expect:**
Only display its functions under the Functions of database.

**Actual:**
Display all subitems of functions and stored procedures under the Functions of database.
image

**More Info:**
This issue also reproduces for attach PostgreSQL server without database.

neelip:
Fixes #1695

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -59,7 +59,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
                 new PostgresTablesTreeItem(this, clientConfig)
             ];
 
-            if (this.parent.supportsStoredProcedures(clientConfig)) {
+            if (await this.parent.supportsStoredProcedures(clientConfig)) {
                 children.push(new PostgresStoredProceduresTreeItem(this, clientConfig));
             }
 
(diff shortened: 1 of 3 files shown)


src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -143,11 +144,14 @@ export class PostgresServerTreeItem extends AzureParentTreeItem<ISubscriptionCon
...
     }
 
-    public supportsStoredProcedures(): boolean {
+    public async supportsStoredProcedures(clientConfig: ClientConfig): Promise<boolean> {
         // `semver.gte` complains when a version doesn't have decimals (i.e. "10"), so attempt to convert version to SemVer
         let version: SemVer | null;
         if (this._serverVersion) {
             version = coerce(this._serverVersion);
+        } else {
+            const result = await runPostgresQuery(clientConfig, `SHOW server_version;`);
```
EricJizbaMSFT, line 153: It's inefficient to run this query so often when the result should never change - just cache the result inside `this._serverVersion`

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -59,7 +59,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
                 new PostgresTablesTreeItem(this, clientConfig)
             ];
 
-            if (this.parent.supportsStoredProcedures()) {
+            if (this.parent.supportsStoredProcedures(clientConfig)) {
```
EricJizbaMSFT, line 62: You have to await this since it's async now. Otherwise it returns the promise directly, which is always a truth-y value. I thought we had tslint for this, but might not be enabled in this repo

EricJizbaMSFT (approved):

PR #1701, opened 2020-09-10
Reverted to older version of pg-structure to support displaying Postgres Table

### LINKED ISSUE #1699: There is an error in PostgreSQL server "Tables" node
**OS:** All
**Build Version:** 20200910.1

**Repro steps:**
1. Create one PostgreSQL server in VS Code.
2. Expand this PostgreSQL server -> postgres database -> "Tables" node.

**Expect:**
There is no error in "Tables" node.

**Actual:**
There is an error in "Tables" node.
image

neelip:
Fixes #1699, #1702

### DIFF
--- a/package.json
+++ b/package.json
@@ -1052,7 +1052,7 @@
         "ms-rest-azure": "^2.3.1",
         "pg": "^8.3.3",
         "pg-connection-string": "^2.3.0",
-        "pg-structure": "^7.2.3",
+        "pg-structure": "^6.2.0",
         "public-ip": "^4.0.0",
         "semver": "^7.2.2",
         "underscore": "^1.8.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1052,7 +1052,7 @@
         "ms-rest-azure": "^2.3.1",
         "pg": "^8.3.3",
         "pg-connection-string": "^2.3.0",
-        "pg-structure": "^7.2.3",
+        "pg-structure": "^6.2.0",
```
EricJizbaMSFT, line 1055: Can you add more description about why we need an older version? For example, is there a bug we can track that will let us know when we can move to the latest version?

neelip: pg-structure packages version >6 only support Postgres versions >=10. Created an issue for this #1702

EricJizbaMSFT (approved):
I'm okay reverting back to an older version of pg-structure for this release, but I would like us to investigate this further in the next release. For example, I don't like how tables is the only thing using pg-structure (functions/sprocs are just using queries). I also think this bug will be easy to pop up again because we update packages somewhat frequently

Issue #1707, opened 2020-09-15
An error occurs when creating a function/stored procedure with name "case"


v-xinda:
**OS:** All
**Build:** 20200911.9

**Repro Steps:**
1. Drill down the PostgreSQL server --> Right click the "Functions" node --> Click "Create Function Query..." --> Enter a name with "case" --> Select a type --> Press Enter.
2. Open a .psql file and click "Execute Query".
3. Check the result.

**Expect:**
Succeed to create the function.

**Actual:**
Fail to create the function.
![image](https://user-images.githubusercontent.com/57662131/93178031-3ec45300-f766-11ea-9127-d06a2546b09b.png)

**More Info:**
This issue also reproduces when creating a stored procedure with name "case".

PR #1709, opened 2020-09-15
Check against reserved words when creating function/sproc queries

wwlorey:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/1707 although its probably not a blocker for the release


src/postgres/commands/createPostgresQuery/validateIdentifier.ts
```diff
@@ -21,5 +21,90 @@ export function validateIdentifier(identifier: string): string | undefined {
         return localize('canOnlyContainCertainCharacters', 'Name can only contain letters, underscores, digits (0-9), and dollar signs ($).');
     }
 
+    if (reservedWords.has(identifier.toLowerCase())) {
+        return localize('cantIncludeReservedWord', 'Name cannot include reserved word "{0}".', identifier);
```
EricJizbaMSFT, line 25: When you say "include", I would assume that means "all" _and_ "all123" are disallowed. I don't think that's what you mean, though...

```suggestion
        return localize('cantIncludeReservedWord', 'Name cannot be reserved word "{0}".', identifier);
```

EricJizbaMSFT (approved):

PR #1712, opened 2020-09-16
Updated Postgres default values to "Basic"

neelip:


src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -30,11 +30,11 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresSe
...
                     sku: {
-                        name: "GP_Gen5_4",
-                        tier: "GeneralPurpose",
-                        capacity: 4,
+                        name: "B_Gen5_1",
+                        tier: "Basic",
+                        capacity: 1,
                         family: "Gen5",
-                        size: "102400"
+                        size: "5120"
```
EricJizbaMSFT, line 37: I think you mean `51200`

Also, let's make a constant for this since it's used twice

neelip: The size variable takes values in MB and the lowest available size on portal is 5GB.

EricJizbaMSFT: Okay I get why we're picking cheaper options as the default, but do we really need the lowest of the low? 😝 5 GB seems really low to me, and it doesn't even save that much money. If we go that low, I propose we make sure 'storageAutoGrow' is set to 'Enabled'

EricJizbaMSFT: cc @fiveisprime @nturinski for your thoughts on storage size and/or [auto-grow](https://docs.microsoft.com/azure/postgresql/concepts-pricing-tiers#storage-auto-grow)

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -6,6 +6,7 @@
 import PostgreSQLManagementClient from 'azure-arm-postgresql';
 import { Progress } from 'vscode';
 import { AzureWizardExecuteStep, callWithMaskHandling, createAzureClient } from 'vscode-azureextensionui';
+import { postgresDefaultStorageSizeMB } from '../../../../constants';
```
EricJizbaMSFT, line 9: Move the constant into this file, close to where it's actually used. We don't want to use the "constants.ts" file for every single constant we have otherwise it'll get out of hand quickly

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -30,11 +31,11 @@ export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresSe
...
                     sku: {
-                        name: "GP_Gen5_4",
-                        tier: "GeneralPurpose",
-                        capacity: 4,
+                        name: "B_Gen5_1",
+                        tier: "Basic",
+                        capacity: 1,
                         family: "Gen5",
-                        size: "102400"
+                        size: `postgresDefaultStorageSizeMB`
```
EricJizbaMSFT, line 38: Does this work? I normally just do `String(storageSizeMB)`. If you use template literal, I thought you would need `${`

neelip: That's true, I forgot to put the ${. I tested it and it was working so didn't realize, maybe the value was pulled from the second variable instead of this.

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
nturinski: Nathan and Matt agreed with Eric, let's use the portal's basic default which is 50 GB with auto-grow enabled.

EricJizbaMSFT (approved):

PR #1717, opened 2020-09-22
Replace pgStructure in PostgresTablesTreeItem with queries

### LINKED ISSUE #1702: "pg-structure" versions higher than 6 support Postgres version >=10
Related #1699 

image

We see this error with only pg-structure versions >6 as "regexp_match()" is only available in Postgres >= 10.

neelip:
Fixes #1702

### DIFF
--- a/src/postgres/getPostgresTablesQueryRows copy.ts
+++ b/src/postgres/getPostgresTablesQueryRows copy.ts
@@ -1,0 +1,44 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ClientConfig } from "pg";
+import { runPostgresQuery } from "./runPostgresQuery";
+
+export interface IPostgresTable {
+    schemaName: string;
+    name: string;
+    oid: number;
+    columnNames: string[];
+}
+
+function getTableInfo(): string {
+    return `select schemaname, tablename
+            from pg_catalog.pg_tables
+            where schemaname != 'pg_catalog' AND
+            schemaname != 'information_schema'`;
+}
+
+function getTableOID(tableName: string): string {
+    return `select oid from pg_class where relname = '${tableName}'`;
+}
+
+function getColumnNames(tableName: string): string {
+    return `select column_name as name
+            from information_schema.columns
+            where table_name = '${tableName}'`;
+}
+
+export async function getTable(clientConfig: ClientConfig): Promise<IPostgresTable[]> {
+    const tableInfoRows = await runPostgresQuery(clientConfig, getTableInfo());
+    const tablesArray: IPostgresTable[] = [];
+    for (const row of tableInfoRows.rows) {
+        const schemaName = row.schemaname;
+        const tableName = row.tablename;
+        const tableOIDResult = await runPostgresQuery(clientConfig, getTableOID(tableName));
+        const columnsResult = await runPostgresQuery(clientConfig, getColumnNames(tableName));
+        tablesArray.push({ schemaName, name: tableName, oid: tableOIDResult.rows[0].oid, columnNames: columnsResult.rows.map(result => result.name) });
+    }
+    return tablesArray;
+}
--- a/src/postgres/getPostgresTablesQueryRows.ts
+++ b/src/postgres/getPostgresTablesQueryRows.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ClientConfig } from "pg";
-import { runPostgresQuery } from "./runPostgresQuery";
+import { runPostgresQuery, wrapArgInQuotes } from "./runPostgresQuery";
 
 export interface IPostgresTable {
     schemaName: string;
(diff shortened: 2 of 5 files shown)


src/postgres/getPostgresTablesQueryRows copy.ts
```diff
@@ -0,0 +1,44 @@
...
+
+function getTableInfo(): string {
+    return `select schemaname, tablename
+            from pg_catalog.pg_tables
+            where schemaname != 'pg_catalog' AND
+            schemaname != 'information_schema'`;
+}
+
+function getTableOID(tableName: string): string {
+    return `select oid from pg_class where relname = '${tableName}'`;
```
EricJizbaMSFT, line 24: Use `wrapArgInQuotes` around tableName

src/postgres/getPostgresTablesQueryRows copy.ts
```diff
@@ -0,0 +1,44 @@
+/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT, line 1: Remove " copy" from the file name

src/postgres/getPostgresTablesQueryRows copy.ts
```diff
@@ -0,0 +1,44 @@
...
+    return `select oid from pg_class where relname = '${tableName}'`;
+}
+
+function getColumnNames(tableName: string): string {
+    return `select column_name as name
+            from information_schema.columns
+            where table_name = '${tableName}'`;
+}
+
+export async function getTable(clientConfig: ClientConfig): Promise<IPostgresTable[]> {
```
EricJizbaMSFT, line 33: getTable -> getTables

src/postgres/getPostgresTablesQueryRows.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface IPostgresTable {
+    schemaName: string;
+    name: string;
+    oid: number;
+    columnNames: string[];
+}
+
+function getTableInfo(): string {
+    return `select schemaname, tablename
```
EricJizbaMSFT, line 17: I feel like @wwlorey should take over reviewing this PR since he wrote most of the existing pg queries in this repo. The rest of the code looks fine to me, but not that familiar with the pg queries themselves

For example, I wasn't sure if there was a way to combine the queries into one, more-efficient query

src/postgres/getPostgresTablesQueryRows.ts
```diff
@@ -0,0 +1,44 @@
...
+import { runPostgresQuery, wrapArgInQuotes } from "./runPostgresQuery";
+
+export interface IPostgresTable {
+    schemaName: string;
+    name: string;
+    oid: number;
+    columnNames: string[];
+}
+
+function getTableInfo(): string {
```
wwlorey, line 16: I think its worthwhile to combine these three queries into one to make it more efficient. Something like this would be a good starting point:

```sql
select schemaname,
tablename,
pg_class.oid,
columns.column_name

from pg_catalog.pg_tables
left join pg_class on relname = tablename
left join information_schema.columns columns on table_name = tablename

where schemaname != 'pg_catalog' AND
schemaname != 'information_schema'
```

But I'm not sure if table names are unique so it would be worth looking into that & running the query through some testing

src/postgres/getPostgresTablesQueryRows.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface IPostgresTable {
+    schemaName: string;
+    name: string;
+    columnNames: string[];
+}
+
+function getTableInfo(): string {
+    return `select schemaname, tablename, array_agg (columnname) as columnarray
+            from pg_catalog.pg_tables
+            left join (select column_name::text as columnname, table_name, table_schema from information_schema.columns) columns on table_name = tablename and table_schema = schemaname
```
wwlorey, line 18: Could you explain what `::text` does in `select column_name::text ...`?

wwlorey, line 18: > ```sql
>... on table_name = tablename and table_schema = schemaname
>```

Just making sure - table names are unique in each schema right?

neelip: Yes, table names are unique within a schema

neelip: The `column_name::text` parses column_name to text data type so that it can be aggregated later, otherwise the original data type of column_name is information_schema.sql_identifier.

src/postgres/tree/PostgresTableTreeItem.ts
```diff
@@ -3,38 +3,34 @@
...
 
-    constructor(parent: PostgresTablesTreeItem, table: Table, isDuplicate: boolean) {
+    constructor(parent: PostgresTablesTreeItem, table: IPostgresTable, isDuplicate: boolean) {
         super(parent);
         this.table = table;
         this._isDuplicate = isDuplicate;
     }
 
-    public get id(): string {
-        return String(this.table.oid);
```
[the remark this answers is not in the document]
neelip: I was having some issues with finding oid values in the relations initially, so this commit doesn't have oid code. But the next commit includes it.

src/postgres/getTables.ts
```diff
@@ -0,0 +1,33 @@
...
+                                from information_schema.columns
+                            ) columns on table_name = tablename and table_schema = schemaname
+                            where schemaname != 'pg_catalog' AND schemaname != 'information_schema'
+                            group by schemaname, tablename;`;
+
+export async function getTables(clientConfig: ClientConfig): Promise<IPostgresTable[]> {
+    const tableInfoRows: QueryResult = await runPostgresQuery(clientConfig, tablesQuery);
+    const tablesArray: IPostgresTable[] = [];
+    for (const row of tableInfoRows.rows) {
+        tablesArray.push({ schemaName: row.schemaname, name: row.tablename, oid: row.oid, columnNames: row.columnarray });
```
wwlorey, line 30: This looks good!

One last thing:  I'd rename `columnarray` to `columnnames` to more closely match `columnNames` from `IPostgresTable`. `columnarray` makes me think there might be more than just names in the array

wwlorey (approved):

PR #1724, opened 2020-10-20
Added "Attach Emulator..." to menu options

### LINKED ISSUE #1520: Add "Attach Emulator..." tree item
**OS:** Windows 10
**Build Version:** 20200522.22

**Repro Steps:**
Check the "Attach Database Accounts" and its context menu.
image

**Actual:**
It only shows "Attach Database Account".

**Suggestion:**
Change "Attach Database Account" to "Attach Database Account or Emulator".

neelip:
Fixes #1520


wwlorey:
I think there might be some confusion about what the issue was asking for. We want to add a tree item under "Attached Database Accounts" called "Attach Emulator...". And it should only show up on windows 
(which is the only OS that shows the "Attach Emulator..." menu item: https://github.com/microsoft/vscode-cosmosdb/blob/neelima/issue-1520-attach-emulator-option/src/tree/AttachedAccountsTreeItem.ts#L41)

nturinski:
Closing/reopening to trigger build

package.json
```diff
@@ -697,7 +697,7 @@
                 },
                 {
                     "command": "cosmosDB.attachEmulator",
-                    "when": "view == azureDatabasesExplorer && viewItem == cosmosDBAttachedAccountsWithEmulator"
+                    "when": "view == azureDatabasesExplorer && viewItem =~ /^cosmosDBAttachedAccounts/"
```
wwlorey, line 700: We only want to show the "Attach Emulator..." menu item if the context value is exactly `cosmosDBAttachedAccountsWithEmulator`. This regular expression shows that menu item for context values including `cosmosDBAttachedAccountsWithoutEmulator`

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -56,6 +56,10 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
         this._loadPersistedAccountsTask = this.loadPersistedAccounts();
     }
 
+    public static isPlatformWindows(): boolean {
+        return process.platform === 'win32';
+    }
```
wwlorey, line 61: It looks like this already exists here: https://github.com/microsoft/vscode-cosmosdb/blob/8d86a075ef090aec387c9a9c3861b42f1c383e8b/src/constants.ts#L6

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -107,12 +111,24 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
+            return attachedAccounts;
+        } else {
+            const genericTreeItemWithoutEmulator = new GenericTreeItem(this, {
+                contextValue: 'cosmosDBAttachDatabaseAccount',
+                label: 'Attach Database Account...',
+                commandId: 'cosmosDB.attachDatabaseAccount',
+                includeInTreeItemPicker: true
+            });
+            const genericTreeItemWithEmulator = new GenericTreeItem(this, {
+                contextValue: 'cosmosDBAttachDatabaseAccount',
```
wwlorey, line 124:
```suggestion
                contextValue: 'cosmosDBAttachEmulator',
```

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -107,12 +111,24 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
-        return attachedAccounts.length > 0 ? attachedAccounts : [new GenericTreeItem(this, {
-            contextValue: 'cosmosDBAttachDatabaseAccount',
-            label: 'Attach Database Account...',
-            commandId: 'cosmosDB.attachDatabaseAccount',
-            includeInTreeItemPicker: true
-        })];
+        if (attachedAccounts.length > 0) {
+            return attachedAccounts;
+        } else {
+            const genericTreeItemWithoutEmulator = new GenericTreeItem(this, {
```
wwlorey, line 117: Could you update the `genericTreeItemWithoutEmulator` and `genericTreeItemWithEmulator` variable names? We show `genericTreeItemWithoutEmulator` whether or not the emulator is enabled.

Maybe something like `attachDatabaseAccount` and `attachEmulator` could be more fitting

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -107,12 +111,24 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
+                includeInTreeItemPicker: true
+            });
+            const genericTreeItemWithEmulator = new GenericTreeItem(this, {
+                contextValue: 'cosmosDBAttachDatabaseAccount',
+                label: 'Attach Emulator...',
+                commandId: 'cosmosDB.attachEmulator',
+                includeInTreeItemPicker: true
+            });
+            return AttachedAccountsTreeItem.isPlatformWindows() ? [genericTreeItemWithoutEmulator, genericTreeItemWithEmulator] :
+                [genericTreeItemWithoutEmulator];
```
wwlorey, line 130: nit: Minor whitespace change 
```suggestion
            return AttachedAccountsTreeItem.isPlatformWindows() ? 
                [genericTreeItemWithoutEmulator, genericTreeItemWithEmulator] :
                [genericTreeItemWithoutEmulator];
```

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -107,12 +111,24 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
 
         const attachedAccounts: AzureTreeItem[] = await this.getAttachedAccounts();
 
-        return attachedAccounts.length > 0 ? attachedAccounts : [new GenericTreeItem(this, {
-            contextValue: 'cosmosDBAttachDatabaseAccount',
-            label: 'Attach Database Account...',
-            commandId: 'cosmosDB.attachDatabaseAccount',
-            includeInTreeItemPicker: true
-        })];
```
wwlorey, line 115: Please verify that adding this new tree item isn't breaking the detach logic, which checks for a specific tree item when determining if its possible to detach.
https://github.com/microsoft/vscode-cosmosdb/blob/8d86a075ef090aec387c9a9c3861b42f1c383e8b/src/extension.ts#L105-L123

package.json
neelip: That makes sense, I forgot to change it back to the original value.

src/tree/AttachedAccountsTreeItem.ts
neelip: That's true. Line 41 can use the same value as well. Will update both lines.

src/tree/AttachedAccountsTreeItem.ts
neelip: Does this context value have to match with 
`{
                    "command": "cosmosDB.attachEmulator",
                    "when": "view == azureDatabasesExplorer && viewItem == cosmosDBAttachedAccountsWithEmulator"
},` ? I am trying understand the behavior of the context value for GenericTreeItem.

wwlorey: So the context value for this generic tree item doesn't matter too much - it should just have a name that matches with the purpose of the tree item.

That rule you mentioned only has to do with the "cosmosDB.attachEmulator" command showing up in a context menu. Which we only want for the `AttachedAccountsTreeItem` that has a context value of "cosmosDBAttachedAccountsWithEmulator ".

Since we define `commandId: 'cosmosDB.attachEmulator'` when this generic tree item is created, clicking it will run that command and we don't have to mess with anything in package.json

wwlorey:
@neelip have you had a chance to take another look at this?

src/tree/AttachedAccountsTreeItem.ts
neelip: I just checked, detach is working for both scenarios.

neelip:
Sorry was busy working on the package updates PR, didn't get a chance to work on this.

wwlorey (approved):

PR #1730, opened 2020-10-23
Updated deprecated packages "azure-arm-cosmosdb" and "azure-arm-postgresql"

neelip:
Deprecated packages: `azure-arm-cosmosdb`, `azure-arm-postgresql` --> `@azure/arm-cosmosdb`, `@azure/arm-postgresql`
Dependencies: `vscode-azureextensionui@latest`, `@azure/ms-rest-js`, `@azure/ms-rest-azure`, `@azure/arm-resources`, 
                          `@azure/ms-rest-nodeauth`


src/AzureDBExperiences.ts
```diff
@@ -31,20 +31,25 @@ export function getExperienceFromApi(api: API): Experience {
     return info;
 }
 
-export function getExperienceLabel(resource: DatabaseAccount & Server): string {
-    const experience: Experience | undefined = tryGetExperience(resource);
-    if (experience) {
-        return experience.shortName;
+export function getExperienceLabel(databaseAccount?: DatabaseAccountGetResults, server?: Server): string | undefined {
```
EricJizbaMSFT, line 34: I think you can get rid of `server`. I only see you call `getExperienceLabel` once, and you're passing `undefined` for server anyways

package.json
```diff
@@ -6,7 +6,7 @@
     "displayName": "Azure Databases",
     "description": "Create, browse, and update globally distributed, multi-model databases in Azure.",
     "engines": {
-        "vscode": "^1.40.0"
+        "vscode": "^1.50.0"
```
EricJizbaMSFT, line 9: Not quite - when you update the VS Code engine, you always want to be on the _minimum_ version you need (in our case 1.48.0). Otherwise, people on older versions of VS Code won't get the update to our extension until they update VS Code, which just slows down the adoption of our latest extension

package.json
```diff
@@ -1012,14 +1012,15 @@
...
+        "@azure/arm-resources": "^4.0.0",
         "@azure/cosmos": "^3.6.3",
         "@types/documentdb": "^1.10.2",
         "@types/fs-extra": "^8.1.0",
         "@types/keytar": "4.0.1",
         "@types/mocha": "^7.0.2",
         "@types/mongodb": "^3.3.2",
-        "@types/node": "^12.0.0",
-        "@types/vscode": "1.40.0",
+        "@types/node": "^14.14.5",
```
EricJizbaMSFT, line 1022: You should keep the Node types at 12. Our version of Node is controlled by VS Code. You can find that info by doing "Code -> About Visual Studio Code" (the minor/patch version doesn't really matter, just the major version). For example:
Screen Shot 2020-10-27 at 3 27 19 PM

package.json
```diff
@@ -6,7 +6,7 @@
     "displayName": "Azure Databases",
```
EricJizbaMSFT, line 6: In other repos, I had to add this to get the build to pass when updating the azure packages:
https://github.com/microsoft/vscode-azurefunctions/blob/main/tsconfig.json#L23

May or may not help the build error you're seeing for this PR

EricJizbaMSFT (approved):

package.json
neelip: It works, thanks! I was a `'async' modifier cannot be used in an ambient context` error in the `vscode-azureextensionui` package.

PR #1737, opened 2020-11-03
Update from "documentdb" package to "@azure/comos"

neelip:


package.json
```diff
@@ -1040,6 +1040,7 @@
         "webpack-cli": "^3.1.2"
     },
     "dependencies": {
+        "@azure/cosmos": "^3.6.3",
         "@azure/arm-cosmosdb": "^9.1.0",
```
EricJizbaMSFT, line 1044: Can you remove "documentdb" and "@types/documentdb" now? Also we probably don't need "@azure/cosmos" listed in devDependencies now that it's in "dependencies"

neelip: Thank you, I missed to remove them. Will do!

EricJizbaMSFT:
Looks like the build is failing because of some more references to "documentdb" that you need to change. I'm gonna wait to review until you can get that build passing

neelip:
"DisableSSLVerification" is no longer available with CosmosClient. I am looking at ways to bypass ssl for the new CosmosClient to support Emulator usage. I thought of creating an issue for this and move along with the current PR, what do you think?

src/docdb/getDocumentClient.ts
```diff
@@ -3,27 +3,18 @@
...
-    const connectionPolicy = new documentBase.ConnectionPolicy();
-
-    const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
-    const strictSSL = !isEmulator && vscodeStrictSSL;
-    connectionPolicy.DisableSSLVerification = !strictSSL;
-    const client = new DocumentClient(documentEndpoint, { masterKey: masterKey }, connectionPolicy);
+export function getDocumentClient(endpoint: string, key: string): CosmosClient {
+    const client = new CosmosClient({ endpoint, key });
 
     // User agent isn't formally exposed on the client (https://github.com/Azure/azure-documentdb-node/issues/244) but nevertheless can be accessed via defaultHeaders
```
EricJizbaMSFT, line 12: Is there a better way to do this now that we're on a new sdk?

src/docdb/getDocumentClient.ts
```diff
@@ -3,27 +3,18 @@
...
-export function getDocumentClient(documentEndpoint: string, masterKey: string, isEmulator: boolean | undefined): DocumentClient {
-    const documentBase = DocDBLib.DocumentBase;
-    const connectionPolicy = new documentBase.ConnectionPolicy();
-
-    const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
-    const strictSSL = !isEmulator && vscodeStrictSSL;
-    connectionPolicy.DisableSSLVerification = !strictSSL;
-    const client = new DocumentClient(documentEndpoint, { masterKey: masterKey }, connectionPolicy);
+export function getDocumentClient(endpoint: string, key: string): CosmosClient {
+    const client = new CosmosClient({ endpoint, key });
```
EricJizbaMSFT, line 10: > "DisableSSLVerification" is no longer available with CosmosClient. I am looking at ways to bypass ssl for the new CosmosClient to support Emulator usage. I thought of creating an issue for this and move along with the current PR, what do you think?

Yes I think that's fine - just make extra sure we fix that bug before releasing because it would represent a regression if we didn't

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -109,21 +94,23 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
             validateInput: (input: string) => validateThroughput(isFixed, input)
         }));
 
-        const options = { offerThroughput: throughput };
+        let partitionKey: PartitionKeyDefinition | undefined;
+        if (!isFixed) {
+            if (!partitionKeyInput.startsWith('/')) {
+                partitionKeyInput = '/' + partitionKeyInput;
+            }
+            partitionKey = { paths: [partitionKeyInput] };
```
EricJizbaMSFT, line 102: We don't need to specify "kind" anymore?

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -35,27 +35,27 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Collec
...
     }
 
     public get databaseName(): string {
         return this._database.id;
     }
 
-    public async getIterator(client: DocumentClient, feedOptions: FeedOptions): Promise<QueryIterator<CollectionMeta>> {
-        return client.readCollections(this.link, feedOptions);
+    public async getIterator(client: CosmosClient, feedOptions: FeedOptions): Promise<QueryIterator<ContainerDefinition & Resource>> {
+        return client.database(this._database.id).containers.readAll(feedOptions);
```
EricJizbaMSFT, line 58: So these "readAll" methods don't _actually_ try to list _all_ items right? They still paginate if there's a bunch?

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -57,39 +58,33 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
 
         if (docID || docID === "") {
             docID = docID.trim();
-            let body = { id: docID };
-            body = <NewDocument>(await this.promptForPartitionKey(body));
+            let body: ItemDefinition = { id: docID };
+            body = (await this.promptForPartitionKey(body));
             context.showCreatingTreeItem(docID);
-            const document = await this.createDocument(body);
+            const item: ItemResponse<ItemDefinition> = await this.root.getDocumentClient().database(this.parent.parent.id).container(this.parent.id).items.create(body);
```
EricJizbaMSFT, line 64: These lines and `this.createDocument` just below it are duplicated. Either remove `createDocument` or call it here

src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -57,39 +58,33 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<RetrievedDocument>
...
+    public async createDocument(body: ItemDefinition): Promise<ItemDefinition> {
+        const item: ItemResponse<ItemDefinition> = await this.root.getDocumentClient().database(this.parent.id).container(this.id).items.create(body);
+        return nonNullProp(item, 'resource');
     }
 
     public documentHasPartitionKey(doc: Object): boolean {
         let interim = doc;
-        let partitionKey: string | undefined = this.parent.partitionKey && this.parent.partitionKey.paths[0];
-        if (!partitionKey) {
+        const partitionKey = this.parent.partitionKey && this.parent.partitionKey;
```
EricJizbaMSFT, line 79: Please add the new type here. Also how come it has a "paths" object that's an array now? And are we sure we should only use the first path?

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -20,10 +23,12 @@ export class DocDBStoredProcedureTreeItem extends AzureTreeItem<IDocDBTreeRoot>
     public readonly contextValue: string = DocDBStoredProcedureTreeItem.contextValue;
     public readonly commandId: string = 'cosmosDB.openStoredProcedure';
     public readonly cTime: number = Date.now();
+    public readonly parent: DocDBStoredProceduresTreeItem;
     public mTime: number = Date.now();
 
-    constructor(parent: DocDBStoredProceduresTreeItem, public procedure: ProcedureMeta) {
+    constructor(parent: DocDBStoredProceduresTreeItem, public procedure: (StoredProcedureDefinition & Resource)) {
         super(parent);
+        this.parent = parent;
```
EricJizbaMSFT, line 31: Remove this line - the `super` call will handle setting the parent

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -44,7 +49,11 @@ export class DocDBStoredProcedureTreeItem extends AzureTreeItem<IDocDBTreeRoot>
     }
 
     public async getFileContent(): Promise<string> {
-        return this.procedure.body;
+        const clonedProcedure: {} = { ...this.procedure?.body?.toString };
```
EricJizbaMSFT, line 52: How come the behavior changed here?

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -3,20 +3,37 @@
...
 import { DocDBStoredProcedureTreeItem } from './DocDBStoredProcedureTreeItem';
 import { DocDBTreeItemBase } from './DocDBTreeItemBase';
 
 /**
  * This class represents the DocumentDB "Stored Procedures" node in the tree
  */
-export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<ProcedureMeta> {
+export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProcedureDefinition> {
+
+    public get iconPath(): string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } {
```
EricJizbaMSFT, line 22: Please revert the order of the fields so that it's easier to see what you changed

src/docdb/tree/DocDBTreeItemBase.ts
```diff
@@ -43,22 +43,16 @@ export abstract class DocDBTreeItemBase<T> extends AzureParentTreeItem<IDocDBTre
...
-                });
-            });
-            if (resource === undefined) {
+            const resourceArray: T[] | undefined = (await this._iterator.fetchNext()).resources;
+            if (resourceArray === undefined) {
                 this._hasMoreChildren = false;
                 break;
             } else {
-                resources.push(resource);
+                resources.push(...resourceArray);
```
EricJizbaMSFT, line 51: If it's an array now instead of a single object, we'll have to re-do how we calculate batchSize

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -53,17 +62,8 @@ export class DocDBStoredProcedureTreeItem extends AzureTreeItem<IDocDBTreeRoot>
...
-            { body: content, id: this.procedure.id },
-            (err, updated: ProcedureMeta) => {
-                if (err) {
-                    reject(err);
-                } else {
-                    resolve(updated);
-                }
-            })
-        );
+        const replace = await client.database(this.parent.parent.parent.id).container(this.parent.parent.id).scripts.storedProcedure(this.id).replace({ id: this.id, body: content });
```
EricJizbaMSFT, line 65: I think it's too easy to accidentally use the wrong parent id when you're doing this kind of stuff. How about at each tree-level you make a method like `getDatabaseClient` and `getContainerClient`. I imagine `getContainerClient` would be something like this:
```
return (await this.parent.getDatabaseClient()).container(this.id);
```

The benefit is that you're only specifying 'id' within that method, so if you get it wrong it would affect all usages and we would find out faster.

EricJizbaMSFT (changes requested):
Okay so there's a lot of changes in this PR, which is fair given what you're doing. _However_, that means I would like to see a decent amount of "bake" time and/or a bug bash after this is merged to weed out any regressions. We may need to re-visit the release date with the team

src/docdb/getDocumentClient.ts
neelip: No, `UserAgentSuffix` is not exposed on the client. The only change is that with the new SDK we don't to access the suffix through `defaultHeaders`.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
neelip: `kind` as a property for partitionKey is not available with the new SDK and haven't found its usage anywhere in samples. I am trying to find documentation to confirm this.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
neelip: It does paginate based on the FeedOptions, and QueryIterator allows for accessing all in response or in batches.

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
neelip: image
Actually, modifications to the string that are made here are not necessary , but it requires a small fix as `body` returns `string | ((...inputs: any[]) => void`.

src/docdb/tree/DocDBTreeItemBase.ts
neelip: I checked this method again after we talked. The current version actually works, except that with the new sdk we need set the `maxItemCount = 1` so that `fetchNext()` behaves the same way as `nextItem()`.

ejizba: I'm not sure what you mean by "actually works". If `_batchSize` is 10, my theory is that it's actually loading 10 batches (aka 100 items) every time a user clicks load more. To fix, I would get rid of the while loop so that it only loads one batch - this should be simpler and more efficient than changing `maxItemCount` to 1

src/docdb/getCosmosClient.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CosmosClient } from "@azure/cosmos";
+import { appendExtensionUserAgent } from "vscode-azureextensionui";
+
+export function getCosmosClient(endpoint: string, key: string): CosmosClient {
+    const client = new CosmosClient({ endpoint, key });
```
ejizba, line 10: Seems to be exposed to me - like this:

```typescript
return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent() });
```

src/docdb/tree/DocDBDatabaseTreeItem.ts
```diff
@@ -12,7 +12,11 @@ export class DocDBDatabaseTreeItem extends DocDBDatabaseTreeItemBase {
...
     public readonly childTypeLabel: string = 'Collection';
 
-    public initChild(collection: CollectionMeta): DocDBCollectionTreeItem {
-        return new DocDBCollectionTreeItem(this, collection);
+    public initChild(container: ContainerDefinition & Resource): DocDBCollectionTreeItem {
+        return new DocDBCollectionTreeItem(this, container);
+    }
+
+    public async getDatabaseClient(client: CosmosClient): Promise<Database> {
+        return client.database(this.id);
```
ejizba, line 20: This and other methods like `getContainerClient` can be sync instead of async

src/docdb/tree/DocDBDocumentTreeItem.ts
```diff
@@ -145,4 +120,8 @@ export class DocDBDocumentTreeItem extends AzureTreeItem<IDocDBTreeRoot> impleme
         }
         return value;
     }
+
+    private async getDocumentClient(client: CosmosClient): Promise<Item> {
+        return (await this.parent.getContainerClient(client)).item(nonNullProp(this.document, 'id'), this.getPartitionKeyValue);
```
ejizba, line 125: `getPartitionKeyValue` -> `getPartitionKeyValue()`, right?

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -67,8 +61,14 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<ProcedureMe
...
     }
 
-    public async getIterator(client: DocumentClient, feedOptions: FeedOptions): Promise<QueryIterator<ProcedureMeta>> {
-        return client.readStoredProcedures(this.link, feedOptions);
+    public async getIterator(client: CosmosClient, feedOptions: FeedOptions): Promise<QueryIterator<StoredProcedureDefinition & Resource>> {
+        return (await this.getContainerClient(client)).scripts.storedProcedures.readAll(feedOptions);
+    }
+
+    public async getContainerClient(client: CosmosClient): Promise<Container> {
+        return (await (<GraphCollectionTreeItem>this.parent).getContainerClient(client)) ?
```
ejizba, line 69: I think this is over complicated - pretty sure this works:
```
return await this.parent.getContainerClient(client);
```

ejizba (approved):

PR #1740, opened 2020-11-06
Fixes checkNameExists issue that occurred with PR 1730

neelip:
Without this fix, all account names in the cosmos create scenario are getting a "Account name ... is not available."

### DIFF
--- a/src/tree/CosmosDBAccountWizard/CosmosDBAccountNameStep.ts
+++ b/src/tree/CosmosDBAccountWizard/CosmosDBAccountNameStep.ts
@@ -40,7 +40,7 @@ async function validateCosmosDBAccountName(name: string, client: CosmosDBManagem
         return `The name must be between ${min} and ${max} characters.`;
     } else if (name.match(/[^a-z0-9-]/)) {
         return "The name can only contain lowercase letters, numbers, and the '-' character.";
-    } else if ((await client.databaseAccounts.checkNameExists(name)).body.valueOf()) {
+    } else if ((await client.databaseAccounts.checkNameExists(name)).body) {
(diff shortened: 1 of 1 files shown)


src/tree/CosmosDBAccountWizard/CosmosDBAccountNameStep.ts
```diff
@@ -40,7 +40,7 @@ async function validateCosmosDBAccountName(name: string, client: CosmosDBManagem
         return `The name must be between ${min} and ${max} characters.`;
     } else if (name.match(/[^a-z0-9-]/)) {
         return "The name can only contain lowercase letters, numbers, and the '-' character.";
-    } else if (await client.databaseAccounts.checkNameExists(name)) {
+    } else if ((await client.databaseAccounts.checkNameExists(name)).body.valueOf()) {
```
wwlorey, line 43: Could you explain some more about what the problem was and how this fixes it? I don't have much context on it

neelip: With recent package update to "@azure/cosmos", `checkNameExists()` returns a `DatabaseAccountsCheckNameExistsResponse` instead of a boolean.

wwlorey: Do you have to use `valueOf`? It looks like `body` is already type boolean

Screen Shot 2020-11-09 at 12 55 06 PM

neelip: I might have missed it. That's true, valueOf is not required there.

wwlorey (approved):

PR #1764, opened 2020-11-30
Fix username validation for restricted usernames

### LINKED ISSUE #1756: There is no validation for the invalid Administrator Username "Admin" when creating PostgreSQL server
**OS:** All
**Build Version:** 20201126.3

**Repro Steps:**
F1 -> Azure Databases: Create Server... -> PostgreSQL -> Enter a valid server name -> Enter "Admin" as the Administrator Username.

**Expect:**
There is a validation for this invalid Administrator Username, like the validation in portal.
image

**Actual:**
There is no validation for this invalid Administrator Username.
image

**More Info:**
1. An error pops up when creating PostgreSQL server with the invalid Administrator Username “Admin”.
image
2. This issue does not reproduce in portal.
3. This issue does not reproduce for the invalid Administrator Username with "admin".
image
4. This issue also reproduces for other invalid Administrator Usernames, such as "Root/admiN...".

neelip:
Fixes #1756

### DIFF
--- a/src/postgres/commands/createPostgresServer/steps/PostgresServerCredUserStep.ts
+++ b/src/postgres/commands/createPostgresServer/steps/PostgresServerCredUserStep.ts
@@ -39,7 +39,7 @@ async function validateUser(username: string): Promise<string | undefined> {
         return localize('usernameCharacterCheck', 'The name can only contain letters, numbers, and the "_" character.');
     } else if (username.match(/^[0-9]+/)) {
         return localize('usernameBeginningMatch', 'The name cannot start with a number.');
-    } else if (username.startsWith('pg_')) {
+    } else if (username.toLowerCase().startsWith('pg_')) {
         return localize('usernameStartWithCheck', 'Admin username cannot start with "pg_".');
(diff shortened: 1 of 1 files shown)


src/postgres/commands/createPostgresServer/steps/PostgresServerCredUserStep.ts
```diff
@@ -41,7 +41,7 @@ async function validateUser(username: string): Promise<string | undefined> {
         return localize('usernameBeginningMatch', 'The name cannot start with a number.');
     } else if (username.startsWith('pg_')) {
```
ejizba, line 42: What about this check - do we need to do `toLowerCase` to prevent people from starting with "PG_"?

ejizba (approved):

src/postgres/commands/createPostgresServer/steps/PostgresServerCredUserStep.ts
neelip: That's true, "PG_" needs to be lowercase as well. Both versions are considered to be part of the restricted group.

PR #1765, opened 2020-12-01
Fix ssl verification issue with new sdk

### LINKED ISSUE #1749: "DisableSSLVerification" not available with new sdk to support Emulators

### LINKED ISSUE #1763: There is an error under the "SQL Emulator" node after attaching the SQL Emulator account
**OS:** Win10
**Build Version:** 20201125.4
**VS Code Version:** 1.51.1 (user setup)

**Repro Steps:**
1. Right Click "Attached Database Accounts" --> Click "Attached Emulator..." --> Select "Core(SQL)".
2. Expand the "Attached Database Accounts" node --> Expand "SQL Emulator" node.

**Expect:**
There is no error.

**Actual:**
There is an error under the "SQL Emulator" node.
image

**More Info:**
This issue doesn't reproduce on the release build.

neelip:
Fixed #1749 Fixes #1763

### DIFF
--- a/src/docdb/getCosmosClient.ts
+++ b/src/docdb/getCosmosClient.ts
@@ -4,8 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { CosmosClient } from "@azure/cosmos";
+import * as https from "https";
+import * as vscode from 'vscode';
 import { appendExtensionUserAgent } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
 
-export function getCosmosClient(endpoint: string, key: string): CosmosClient {
-    return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent() });
+export function getCosmosClient(endpoint: string, key: string, isEmulator: boolean | undefined): CosmosClient {
+
+    const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
+    return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent(), agent: new https.Agent({ rejectUnauthorized: !isEmulator || vscodeStrictSSL }) });
+
 }
--- a/src/docdb/tree/DocDBAccountTreeItemBase.ts
+++ b/src/docdb/tree/DocDBAccountTreeItemBase.ts
@@ -35,7 +35,7 @@ export abstract class DocDBAccountTreeItemBase extends DocDBTreeItemBase<Databas
             endpoint,
             masterKey,
             isEmulator,
-            getCosmosClient: () => getCosmosClient(endpoint, masterKey)
+            getCosmosClient: () => getCosmosClient(endpoint, masterKey, isEmulator)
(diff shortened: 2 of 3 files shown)


src/docdb/getCosmosClient.ts
```diff
@@ -4,8 +4,14 @@
...
 
 import { CosmosClient } from "@azure/cosmos";
+import * as https from "https";
+import * as vscode from 'vscode';
 import { appendExtensionUserAgent } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
+
+export function getCosmosClient(endpoint: string, key: string, isEmulator: boolean | undefined): CosmosClient {
+
+    const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
```
wwlorey, line 14: Was this setting taken into account when creating a cosmos client in the old SDK?

src/docdb/getCosmosClient.ts
```diff
@@ -4,8 +4,14 @@
...
+    return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent(), agent: new https.Agent({ rejectUnauthorized: !isEmulator || vscodeStrictSSL }) });
```
ejizba, line 15: Before, this was `!isEmulator && vscodeStrictSSL`, and I think it should stay that way. It's reasonably safe to ignore the `vscodeStrictSSL` setting (which will almost always be true) in the emulator case because it should only be traffic from their local machine.

src/docdb/getCosmosClient.ts
```diff
@@ -4,8 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { CosmosClient } from "@azure/cosmos";
+import * as https from "https";
+import * as vscode from 'vscode';
 import { appendExtensionUserAgent } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
+
+export function getCosmosClient(endpoint: string, key: string, isEmulator: boolean | undefined): CosmosClient {
+
```
ejizba, line 13: Can you confirm you tested this on Windows with the emulator and got it working? I assume this also fixes https://github.com/microsoft/vscode-cosmosdb/issues/1763?

src/docdb/getCosmosClient.ts
neelip: Yes, this was part of the last release with the old sdk. 
https://github.com/microsoft/vscode-cosmosdb/blob/9b4b3018fe34f1bccb9522b471bff6f1cccf1903/src/docdb/getDocumentClient.ts

src/docdb/getCosmosClient.ts
neelip: Yes, that's true. It make sense to have `!isEmulator && vscodeStrictSSL`.

src/docdb/getCosmosClient.ts
neelip: Yes, Nathan and I tested this on his windows machine; it is working.

ejizba (approved):

PR #1767, opened 2020-12-02
Fix for Upload to Mongo Collection issue

### LINKED ISSUE #1753: Pop up an error "Cannot convert undefined or null to object" after saving changes in the final result
**OS:** All
**Build Version:** 20201125.4

**Repro Steps:**

1. Create a Scrapbook and connect a database.
2. Type the following in Scrapbook.
  ```
  db.test2.drop()
  db.test2.insert({"_id":"a","abc":"def"})
  db.test2.insert({"_id":123,"abc":"def"})
  db.test2.find()
  ```
3. Execute All -> Change the value of "abc" to "ghi" in the final result -> Save it -> Click "Upload".
4. Check whether save it successfully or not.

**Expect:**
Succeed to save changes without any error.

**Actual:**
Pop up an error "Cannot convert undefined or null to object".
Screen Shot 2020-11-26 at 7 41 42 PM
```
Action: writeFile
Error type: TypeError
Error Message: Cannot convert undefined or null to object
```

**More Info:**
This issue also reproduces for the opened 'cosmos-collection.json' file.

>   Repro Steps: Right click a collection of MongoDB account -> Click "Open Collection" -> Make changes for the opened 'cosmos-collection.json' file -> Save it -> Upload.

neelip:
Fixes #1753

### DIFF
--- a/src/mongo/tree/MongoCollectionTreeItem.ts
+++ b/src/mongo/tree/MongoCollectionTreeItem.ts
@@ -60,7 +60,7 @@ export class MongoCollectionTreeItem extends AzureParentTreeItem<IMongoTreeRoot>
             return {
                 replaceOne: {
                     filter: { _id: document._id },
-                    update: _.omit(document, '_id'),
+                    replacement: _.omit(document, '_id'),
                     upsert: false
                 }
             };


wwlorey (commented):
Is there documentation about this change? I assume its required by the new SDK

neelip:
Found that `update` doesn't exist for `replaceOne` property. It could be update within the new sdk as the problem doesn't exist with the current release. 
image
https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite/

nturinski:
Do you know if the old behavior would insert a document if no document matched?  If so, to keep the behavior the same, I think we need to set `upsert` to `true`.

nturinski:
Are these commands executed by the shell or the MongoDB driver, btw?  I can't remember which commands are just passed off to the mongo shell, but for FWIW:

image

This should be the node.js API for MongoDB
http://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#replaceOne

neelip:
image
3.3 documentation: https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#replaceOne. 
3.6 documentation: https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#replaceOne
It looks like upsert behavior hasn't changed with the new sdk. 

BulkWrite and replaceOne seem to be from the Node.js driver.

ejizba (approved):
Discussed and did some testing offline and seems fine to me

PR #1768, opened 2020-12-02
Fix for importing documents without partition key 

neelip:
The existing release already accounts for importing documents without partition key. An error is thrown with specific partition key information when a document doesn't contain the required partition key. 

The only issue is with scenarios where the partition key paths array has only 1 element, such as "/a".

### DIFF
--- a/src/docdb/tree/DocDBDocumentsTreeItem.ts
+++ b/src/docdb/tree/DocDBDocumentsTreeItem.ts
@@ -85,7 +85,7 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<ItemDefinition> {
         }
         const keyPath = partitionKey.split('/');
         let i: number;
-        for (i = 0; i < keyPath.length - 1; i++) {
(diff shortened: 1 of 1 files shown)


src/docdb/tree/DocDBDocumentsTreeItem.ts
```diff
@@ -85,7 +85,7 @@ export class DocDBDocumentsTreeItem extends DocDBTreeItemBase<ItemDefinition> {
         }
         const keyPath = partitionKey.split('/');
         let i: number;
-        for (i = 0; i < keyPath.length - 1; i++) {
+        for (i = 0; i <= keyPath.length - 1; i++) {
```
ejizba, line 88: This is why I always prefer `for...const...of` instead so you can't have off-by-one errors. Something like this:
```typescript
const properties = partitionKey.split('/');
for (const prop of properties) {
    if (interim.hasOwnProperty(prop)) {
        interim = interim[prop];
```

ejizba (approved):

PR #1776, opened 2020-12-03
Removed deprecated properties Server, ReplSet, Mongos of "mongodb" sdk

### LINKED ISSUE #1770: The attached MongoDB account name shows as "Undefined..."
**OS:** All
**Build Version:** 20201203.7

**Repro Steps:**
1. Create one MongoDB account.
2. Attach this MongoDB account and check the attached account name.

**Expect:**
The attached MongoDB account name is the same with the created one.

**Actual:**
The attached MongoDB account name shows as "Undefined...".
image

**More Info:**
This issue does not reproduce for SQL/Table/Gremlin/PostgreSQL accounts.

neelip:
Fixes #1770

### DIFF
--- a/src/mongo/mongoConnectionStrings.ts
+++ b/src/mongo/mongoConnectionStrings.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { MongoClient, Mongos, ReplSet, Server } from "mongodb";
+import { MongoClient } from "mongodb";
 import { appendExtensionUserAgent, IParsedError, parseError } from "vscode-azureextensionui";
 import { testDb } from "../constants";
 import { ParsedConnectionString } from "../ParsedConnectionString";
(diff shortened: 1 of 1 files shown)


src/mongo/mongoConnectionStrings.ts
```diff
@@ -63,23 +63,16 @@ export async function parseMongoConnectionString(connectionString: string): Prom
         }
     }
 
-    const serverConfig: Server | ReplSet | Mongos = mongoClient.db(testDb).serverConfig;
-    // Azure CosmosDB comes back as a ReplSet
-    if (serverConfig instanceof ReplSet) {
```
neelip, line 68: All the mongo attached accounts are not passing this condition as ReplSet is deprecated. 
However, the code to access `host` and `port` values hasn't changed.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -63,23 +63,16 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
-    }
+    const serverConfig = mongoClient.db(testDb).serverConfig;
+
+    // get the first connection string from the servers list
+    // this may not be best solution, but the connection (below) gives
+    // host name of single server, mongos instance or the primany from replicaSet which is different than what is in the connection string (espcially for Replica sets)
+    // "s" is not part of the static definition but can't find any official documentation on it. Yet it is definitely there at runtime. Grandfathering in.
+    // tslint:disable-next-line:no-any
+    const rs: any = serverConfig;
+    host = rs.s.options.servers[0].host;
```
neelip, line 74: `s.options.servers ` property is available on all types of accounts. I have tested with local mongo connection (single server) and the referred code works for single server type as well.

src/mongo/mongoConnectionStrings.ts
neelip, line 68: All the mongo attached accounts are are not passing this condition and `host = (<any>serverConfig).host;` is returning undefined as `host` and `port` are not available at the top level of the object even for a single server.

src/mongo/mongoConnectionStrings.ts
```diff
@@ -63,23 +63,16 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
+    port = rs.s.options.servers[0].port;
```
neelip, line 76: `s.options.servers` property is available on all types of accounts. I have tested with local mongo connection (single server) and the referred code works for single server type as well.

src/mongo/mongoConnectionStrings.ts
wwlorey: Is there an equivalent way to perform this check using the new SDK? I'm concerned that removing it could lead to problems

neelip: Not really, with the new sdk the serverConfig is of type "any". 
image

However, oddly the documentation here http://mongodb.github.io/node-mongodb-native/3.6/api/Db.html, doesn't show these objects as deprecated.

wwlorey: Yeah it looks like those types still exist. Did you add the `@types` for the new SDK to package.json? I think excluding that may mess up the typing

ejizba: Whatever you figure out, I would prefer to wait for this change until after the release because of how often `parseMongoConnectionString` is used and because of the wide variety of mongo accounts that can be attached

src/mongo/mongoConnectionStrings.ts
```diff
@@ -63,22 +63,19 @@ export async function parseMongoConnectionString(connectionString: string): Prom
...
-        host = rs.s.options.servers[0].host;
-        port = rs.s.options.servers[0].port;
+    // tslint:disable-next-line:no-any
+    const serverConfig: any = mongoClient.db(testDb).serverConfig;
+
+    // get the first connection string from the servers list
+    // this may not be best solution, but the connection (below) gives
+    // host name of single server, mongos instance or the primany from replicaSet which is different than what is in the connection string (espcially for Replica sets)
+    // "s" is not part of the static definition but can't find any official documentation on it. Yet it is definitely there at runtime. Grandfathering in.
+    if (serverConfig.s) {
```
ejizba, line 73: Because this is type `any` and because we're close to a release, I think we want to be even more careful

something like this:
```
const host: string = rs?.s?.options?.servers[0]?.host || serverConfig.host;
const port: string = rs?.s?.options?.servers[0]?.port || serverConfig.port;
```

neelip: I see what you mean. I agree, it is better to use both values.

ejizba (approved):

PR #1777, opened 2020-12-03
RC for 0.15.1

neelip:
- [x] Update Notice.html after #1778  is closed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,19 +4,13 @@ All notable changes to the "azure-cosmosdb" extension will be documented in this
 
 ## 0.16.0 - 2020-12-04
 ### Added
-- MongoDB server defaults to version “3.6”
+- Newly created Azure CosmosDB - MongoDB servers default to version “3.6”
 
 ### Changed
-- Replaced deprecated azure sdk with new sdk:
-  - “documentdb” to “@azure/cosmos”
-  - “azure-arm-cosmosdb” to “@azure/arm-cosmosdb”
-  - “azure-arm-postgresql” to “@azure/arm-postgresql”
-  - Upgraded following packages:
-    - mongodb: "3.3" to "3.6"
-    - “vscode-languageclient" and “vscode-languageserver" to latest versions
+- Minimum version of VS Code is now 1.48.0
 
 ### Fixed
-- [Bugs Fixed](https://github.com/microsoft/vscode-cosmosdb/milestone/25)
+- [Bugs Fixed](https://github.com/microsoft/vscode-cosmosdb/milestone/25?closed=1)
 
 ## 0.15.0 - 2020-09-15
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
-## 0.16.0 - 2020-12-04
+## 0.15.1 - 2020-12-04
 ### Added
 - Newly created Azure CosmosDB - MongoDB servers default to version “3.6”
 
 ### Changed
 - Minimum version of VS Code is now 1.48.0
+- Improved extension activation time (by switching to the azure-sdk-for-js)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+
+### Changed
+- Replaced deprecated azure sdk with new sdk:
+  - “documentdb” to “@azure/cosmos”
+  - “azure-arm-cosmosdb” to “@azure/arm-cosmosdb”
+  - “azure-arm-postgresql” to “@azure/arm-postgresql”
+  - Upgraded following packages:
+    - mongodb: "3.3" to "3.6"
+    - “vscode-languageclient" and “vscode-languageserver" to latest versions
+
```
neelip, line 17: Do we need to mention all of these package updates here?

ejizba: No we don't mention engineering work like this unless it's in the context of a feature/enhancement. For example, like this:
> Improved extension activation time (by switching to the [azure-sdk-for-js](https://github.com/Azure/azure-sdk-for-js))

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
+## 0.16.0 - 2020-12-04
+### Added
+- MongoDB server defaults to version “3.6”
+
+### Changed
+- Replaced deprecated azure sdk with new sdk:
```
ejizba, line 10: Please add this bullet point under "Changed"
> Minimum version of VS Code is now 1.48.0

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
+## 0.16.0 - 2020-12-04
+### Added
+- MongoDB server defaults to version “3.6”
```
wwlorey, line 7: I would say something like `Newly created MongoDB servers default to version "3.6"`. 

Without the "newly created" part I could see people being confused about new vs existing servers and version 3.6

CHANGELOG.md
ejizba: That's fine if it means the changelog doesn't have much in it, but I might do 0.15.1 instead of 0.16.0 in that case

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+- Replaced deprecated azure sdk with new sdk:
+  - “documentdb” to “@azure/cosmos”
+  - “azure-arm-cosmosdb” to “@azure/arm-cosmosdb”
+  - “azure-arm-postgresql” to “@azure/arm-postgresql”
+  - Upgraded following packages:
+    - mongodb: "3.3" to "3.6"
+    - “vscode-languageclient" and “vscode-languageserver" to latest versions
+
+### Fixed
+- [Bugs Fixed](https://github.com/microsoft/vscode-cosmosdb/milestone/25)
```
wwlorey, line 19: That pesky `?closed=1`

```suggestion
- [Bugs Fixed](https://github.com/microsoft/vscode-cosmosdb/milestone/25?closed=1)
```

CHANGELOG.md
ejizba: Also mention Azure Cosmos DB in there somehow, as we have a lot of non-Azure users for Mongo and this will not affect them

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
+## 0.16.0 - 2020-12-04
```
ejizba, line 5: This should be 0.15.1 now as well

CHANGELOG.md
ejizba: To be clear, you can add that exact line I mentioned because your work to update the packages did help with activation time. We put that line in other changelogs when we updates those packages for other extensions

ejizba (approved):

PR #1779, opened 2020-12-03
Remove canConnectToLocalMongoDB function

### LINKED ISSUE #1771: It will takes more than 15 seconds to pop up the "Enter the connection string..." textbox when attaching a MongoDB account
**OS:** All
**Build Version:** 20201203.7

**Repro Steps:**
1. Copy connection string of one MongoDB account.
2. F1 -> Cosmos DB: Attach Database Account... -> Select MongoDB.

**Expect:**
The "Enter the connection string..." textbox pops up in 2 seconds.

**Actual:**
It will takes more than 15 seconds to pop up the "Enter the connection string..." textbox.

**More Info:**
This issue does not reproduce when attaching SQL/Table/Gremlin/PostgreSQL accounts.

neelip:
Fixes #1771

### DIFF
--- a/src/tree/AttachedAccountsTreeItem.ts
+++ b/src/tree/AttachedAccountsTreeItem.ts
@@ -6,7 +6,7 @@
 import { Environment } from '@azure/ms-rest-azure-env';
 import { TokenCredentialsBase } from '@azure/ms-rest-nodeauth';
 import * as vscode from 'vscode';
-import { appendExtensionUserAgent, AzExtParentTreeItem, AzExtTreeItem, AzureParentTreeItem, AzureTreeItem, GenericTreeItem, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { AzExtParentTreeItem, AzExtTreeItem, AzureParentTreeItem, AzureTreeItem, GenericTreeItem, ISubscriptionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { API, getExperienceFromApi, getExperienceQuickPick, getExperienceQuickPicks } from '../AzureDBExperiences';
 import { removeTreeItemFromCache } from '../commands/api/apiCache';
 import { emulatorPassword, getThemedIconPath } from '../constants';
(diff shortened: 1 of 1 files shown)


src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -157,7 +150,7 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
                 prompt: 'Enter the connection string for your database account',
                 validateInput: validateInput,
                 ignoreFocusOut: true,
-                value: defaultValue
+                value: placeholder
```
ejizba, line 153: I feel like this should be either `mongodb://127.0.0.1:27017` or undefined, not the `placeHolder`

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -250,18 +243,6 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
         return this._attachedAccounts;
     }
 
-    private async canConnectToLocalMongoDB(): Promise<boolean> {
```
ejizba, line 253: Another option is to keep this in, but add a timeout of like 1 second. You can leverage `Promise.race` for this

```typescript
private async canConnectToLocalMongoDB(): Promise<boolean> {
    async function timeout(): Promise<boolean> {
        await delay(1000);
        return false;
    }
    async function connect(): Promise<boolean> {
        try {
            const db: MongoClient | void = await connectToMongoClient(localMongoConnectionString, appendExtensionUserAgent());
            // grandfathered in
            // tslint:disable-next-line: no-floating-promises
            db.close();
            return true;
        } catch {
            return false;
        }
    }
    return await Promise.race([timeout(), connect()]);
}
```

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -251,15 +252,22 @@ export class AttachedAccountsTreeItem extends AzureParentTreeItem {
...
-            db.close();
-            return true;
-        } catch (error) {
+        async function timeout(): Promise<boolean> {
+            await delay(1000);
             return false;
         }
+        async function connect(): Promise<boolean> {
+            try {
+                const db: MongoClient | void = await connectToMongoClient(localMongoConnectionString, appendExtensionUserAgent());
```
ejizba, line 261: Haha I hope you tested the code because I didn't - also I messed around with a few options and must've copied something unintentionally because this `| void` thing doesn't need to be there

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -403,3 +411,10 @@ class AttachedAccountRoot implements ISubscriptionContext {
         throw this._error;
     }
 }
+
+async function delay(milliseconds: number): Promise<void> {
```
ejizba, line 415: At some point we want to put this in the utils folder because it's in like 3 files in this repo now

ejizba (approved):

src/tree/AttachedAccountsTreeItem.ts
neelip: Lol I didn't notice the void. No worries, I tested it, it's working.

src/tree/AttachedAccountsTreeItem.ts
neelip: That's a good idea, I did notice it was mentioned in couple places.

PR #1793, opened 2021-01-07
Grouped Context Menus

neelip:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -532,332 +532,412 @@
                 {
                     "command": "azureDatabases.openInPortal",
                     "when": "view == azureDatabasesExplorer && viewItem == azureextensionui.azureSubscription",
-                    "group": "1@2"
+                    "group": "2@1"
                 },
                 {
                     "command": "azureDatabases.refresh",
                     "when": "view == azureDatabasesExplorer && viewItem == azureextensionui.azureSubscription",
-                    "group": "2@1"
+                    "group": "3@1"
                 },
                 {
                     "command": "cosmosDB.deleteAccount",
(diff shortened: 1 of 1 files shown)


nturinski:
Could you provide screenshots of what the new menus look like?  I think that'll make this a lot easier to review.

package.json
```diff
@@ -532,332 +532,412 @@
                 {
                     "command": "azureDatabases.openInPortal",
                     "when": "view == azureDatabasesExplorer && viewItem == azureextensionui.azureSubscription",
-                    "group": "1@2"
+                    "group": "2@1"
```
ejizba, line 535: I like having "Open in Portal" and "Refresh" as a part of the same group, like [I did](https://github.com/microsoft/vscode-azureresourcegroups/blob/main/package.json#L144) in the rg extension. I consider the group to be "common commands across multiple nodes/extensions that are simple but not _that_ useful"

ejizba (approved):
FYI @neelip you'll get a faster response from us if you post a new comment rather than edit the description because it sends another email. I happened to see the edit this time, but won't always check

package.json
nturinski: I agree with Eric.  Plus having a totally separate group for just "Open in Portal" looks a bit clunky to me.

nturinski:
Not a requirement, but I find that this helps keep the `package.json` cleaner.  https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/package.json#L310-L323

By using a Regex, you don't need multiple declarations of .refresh commands.  By grouping it as 9, you're guaranteeing that it'll be at the bottom of the list (unless for some reason we ever hit more than 9 groups 🥴)

PR #1799, opened 2021-01-14
Refresh Postgres Functions tree after executing query

### LINKED ISSUE #1735: The function query doesn't appear automatically under the "Functions" node after executing the query
**OS:** Win10
**Build Version:** 20201102.9

**Repro Steps:**

1. Drill down a PostgreSQL server -> Database -> Right click the "Functions" node -> Click "Create Function Query..." -> Create a function query with name "test" -> Execute the query.
2. Check the result.

**Expect:**
The function query "test" appears automatically under the "Functions" node.

**Actual:**
The function query "test" doesn't appear automatically under the "Functions" node.

**More Info:**

1. This issue is fixed after refreshing.
2. This issue also reproduces for GitHub codespaces.

neelip:
Fixes #1735

### DIFF
--- a/src/postgres/commands/executePostgresQuery.ts
+++ b/src/postgres/commands/executePostgresQuery.ts
@@ -56,4 +56,5 @@ export async function executePostgresQuery(context: IActionContext): Promise<voi
 
         await vscodeUtil.showNewFile(csvData, outputFileName, '.csv');
     }
+    await treeItem.refresh();
 }


src/postgres/commands/executePostgresQuery.ts
```diff
@@ -56,4 +56,5 @@ export async function executePostgresQuery(context: IActionContext): Promise<voi
 
         await vscodeUtil.showNewFile(csvData, outputFileName, '.csv');
     }
+    await treeItem.refresh();
```
wwlorey, line 59: Some words of wisdom given to myself to be aware of: https://github.com/microsoft/vscode-azurestorage/pull/840#discussion_r499943538.

I believe using a refresh in this case is OK though because it doesn't interfere with the file system (we open Postgres queries as one-off new files)

wwlorey (approved):

PR #1810, opened 2021-02-01
Added VerifyProviderStep to "Create Postgres Server"

### LINKED ISSUE #1795: Add `verifyRegisterProvider` step to Wizard's executeSteps
Does this occur consistently? 
Repro steps:

1.
2.

Action: azureDatabases.createServer
Error type: MissingSubscriptionRegistration
Error Message: The subscription is not registered to use namespace 'Microsoft.DocumentDB'. See https://aka.ms/rps-not-found for how to register subscriptions.

Version: 0.15.1
OS: darwin
OS Release: 19.3.0
Product: Visual Studio Code
Product Version: 1.52.1
Language: en

<details>
<summary>Call Stack</summary>

```
new RestError extension.bundle.js:56:970
extension.bundle.js:260:2896
processTicksAndRejections task_queues.js:94:5
```

</details>

neelip:
Fixes #1795

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -73,7 +73,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         LocationListStep.addStep(wizardContext, promptSteps);
 
         const executeSteps: AzureWizardExecuteStep<IPostgresServerWizardContext | ICosmosDBWizardContext>[] = [];
-        executeSteps.push(new VerifyProvidersStep(['Microsoft.Web', 'Microsoft.Storage', 'Microsoft.Insights']));
+        executeSteps.push(new VerifyProvidersStep(['Microsoft.DocumentDB', 'Microsoft.DBforPostgreSQL']));
 
         const wizard = new AzureWizard(wizardContext, {
             promptSteps,
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -8,7 +8,7 @@ import { DatabaseAccountGetResults, DatabaseAccountListKeysResult, DatabaseAccou
 import { PostgreSQLManagementClient } from '@azure/arm-postgresql';
 import { Server, ServerListResult } from '@azure/arm-postgresql/src/models';
 import * as vscode from 'vscode';
-import { AzExtTreeItem, AzureTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, createAzureClient, ICreateChildImplContext, ILocationWizardContext, LocationListStep, ResourceGroupListStep, SubscriptionTreeItemBase, VerifyProvidersStep } from 'vscode-azureextensionui';
+import { AzExtTreeItem, AzureTreeItem, AzureWizard, AzureWizardPromptStep, createAzureClient, ICreateChildImplContext, ILocationWizardContext, LocationListStep, ResourceGroupListStep, SubscriptionTreeItemBase } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -72,9 +72,12 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         ];
         LocationListStep.addStep(wizardContext, promptSteps);
 
+        const executeSteps: AzureWizardExecuteStep<IPostgresServerWizardContext | ICosmosDBWizardContext>[] = [];
+        executeSteps.push(new VerifyProvidersStep(['Microsoft.Web', 'Microsoft.Storage', 'Microsoft.Insights']));
```
ejizba, line 76: Those providers are for App Service. You have to pick the providers for your scenario, and you will have to handle the fact that Cosmos and Postgres probably have different providers that they need

neelip: Didn't realize that 😅  Thanks for catching it!

nturinski (approved):

src/tree/SubscriptionTreeItem.ts
```diff
@@ -72,9 +72,12 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         ];
         LocationListStep.addStep(wizardContext, promptSteps);
 
+        const executeSteps: AzureWizardExecuteStep<IPostgresServerWizardContext | ICosmosDBWizardContext>[] = [];
+        executeSteps.push(new VerifyProvidersStep(['Microsoft.DocumentDB', 'Microsoft.DBforPostgreSQL']));
```
ejizba, line 76: We don't want to register both providers, because a user can only use one at a time. You should add this step in the subwizard here, after you know which type they choose:
https://github.com/microsoft/vscode-cosmosdb/blob/main/src/tree/AzureDBAPIStep.ts#L35

ejizba (changes requested):

src/tree/SubscriptionTreeItem.ts
nturinski: Fair enough.  Registering a provider can be a pretty expensive/long task, so we probably shouldn't register both if they don't need it.

ejizba: Could also be problems where someone only has permissions for one provider or a sovereign cloud only supports one provider, but not the other

neelip: I realized that this code format only verifies providers in the createServer() scenario and it doesn't cover with first time users with the loadMore...() scenario. 

I am not sure how to handle this situation, as we don't want to verify provider every time the servers are loaded. Any thoughts?

ejizba (approved):

src/tree/SubscriptionTreeItem.ts
ejizba: This is what I do for loadMore in functions: https://github.com/microsoft/vscode-azurefunctions/blob/main/src/tree/SubscriptionTreeItem.ts#L49

If the provider isn't registered, by definition that means they don't have any resources. So all that code does is show no resources instead of the error. I'm not sure if cosmos/postgres behaves the same way and honestly that code might not work anymore after we switched to the new sdk (I should test it). In any case, the most important piece is what you've already done, which is to register when they create the first resource.

You can mess around with providers by going here in the portal: "Subscriptions" -> "Your subscription" -> "Resource providers"

neelip: That makes sense, for the resources to not load if the provider is not registered in the first place. 
I will test it out and see how it works on databases.

PR #1815, opened 2021-02-08
Update SQL Stored Procedure Name Validation

### LINKED ISSUE #1719: An error "EntryNotFound..." occurs when creating a new stored procedure with name as "&&&"
**OS:** Win10
**Build Version:** 20200922.5

**Repro Steps:**
1. Drill down the SQL account.
2. Right click the "Stored Procedures" node --> Click "Create Stored Procedure..." --> Enter a name with "&&&".
3. Check the result.

**Expect:**
Succeed to create a new stored procedure without any error.

**Actual:**
An error "EntryNotFound..." occurs when creating a new stored procedure with name as "&&&".
image

**More Info:**
1. Succeed to create a new stored procedure with name as "&&&" without any error in portal.
2. This issue also reproduces for Graph account.

### LINKED ISSUE #1773: There is no validation for the empty Stored Procedure name in SQL account
**OS:** Win10
**Build Version:** 20201203.7

**Repro Steps:**
1. Drill down the SQL account.
2. Right click the "Stored Procedures" node --> Click "Create Stored Procedure..." --> Keep the name with empty.

**Expect:**
There is an error for the empty Stored Procedure name, like below screen.
image

**Actual:**
There is no validation for the empty Stored Procedure name.

**More Info:**
1. There is an error when creating a Stored Procedure with empty name.
image
2. This issue does not reproduce for PostgreSQL server.

### LINKED ISSUE #1789: There is no validation when inputting the duplicate stored procedure name
**OS:** Win10
**Build Version:** 20210105.1

**Repro Steps:**
1. Drill down the SQL account.
2. Create a stored procedure with name "test".
3. Right click the "Stored Procedures" node --> Click "Create Stored Procedure..." --> Input name "test" again.
4. Check if there is a validation for the duplicate name.

**Expect:**
There is a validation when inputting the duplicate stored procedure name.

**Actual:**
There is no validation when inputting the duplicate stored procedure name.
image

**More Info:**
There is an error when creating a stored procedure with the duplicate name.
image

neelip:
Fixes #1789, Fixes #1773 and Fixes #1719

### DIFF
--- a/src/docdb/tree/DocDBStoredProceduresTreeItem.ts
+++ b/src/docdb/tree/DocDBStoredProceduresTreeItem.ts
@@ -5,7 +5,7 @@
 
 import { Container, CosmosClient, FeedOptions, QueryIterator, Resource, StoredProcedureDefinition } from '@azure/cosmos';
 import * as vscode from "vscode";
-import { ICreateChildImplContext } from 'vscode-azureextensionui';
+import { AzExtTreeItem, ICreateChildImplContext } from 'vscode-azureextensionui';
 import { defaultStoredProcedure, getThemeAgnosticIconPath } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { GraphCollectionTreeItem } from '../../graph/tree/GraphCollectionTreeItem';
@@ -39,12 +39,10 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<DocDBStoredProcedureTreeItem> {
         const client = this.root.getCosmosClient();
-        const currStoredProcedureList = await this.getCachedChildren(context);
+        const currStoredProcedureList: AzExtTreeItem[] = await this.getCachedChildren(context);
         const currStoredProcedureNames: string[] = [];
         for (const sp of currStoredProcedureList) {
-            if (sp instanceof DocDBStoredProcedureTreeItem) {
-                currStoredProcedureNames.push(sp.id);
-            }
+            currStoredProcedureNames.push(nonNullProp(sp, "id"));
         }
         const spID = (await ext.ui.showInputBox({
             prompt: "Enter a unique stored procedure ID",
@@ -77,23 +75,21 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
         return this.parent.getContainerClient(client);
     }
 
-    private async validateStoredProcedureName(name: string, currStoredProcedureNames: String[]): Promise<string | null | undefined> {
-        if (name) {
-            if (name.length < 1 || name.length > 255) {
-                return "Name has to be between 1 and 255 chars long";
-            }
-            if (/[/\\?#&]/.test(name)) {
-                return localize("illegalChars", "Name contains illegal chars: /,\\,?,#,&");
-            }
-            if (name[name.length - 1] === " ") {
-                return localize("endsWithSpace", "Name cannot end with a space.");
-            }
-            if (currStoredProcedureNames.includes(name)) {
-                return localize('NameExists', 'Stored Procedure "{0}" already exists.', name);
-            }
-        } else {
-            return localize("noEmptyValue", "Name cannot be empty");
+    private validateStoredProcedureName(name: string, currStoredProcedureNames: String[]): string | undefined {
+        if (!name || name.length < 1 || name.length > 255) {
+            return localize("nameLength", "Name has to be between 1 and 255 chars long");
         }
-        return null;
+
+        if (/[/\\?#&]/.test(name)) {
+            return localize("illegalChars", "Name contains illegal chars: /, \\, ?, #, &");
+        }
+        if (name[name.length - 1] === " ") {
+            return localize("endsWithSpace", "Name cannot end with a space.");
+        }
+        if (currStoredProcedureNames.includes(name)) {
+            return localize('nameExists', 'Stored Procedure "{0}" already exists.', name);
+        }
+
+        return undefined;
     }
 }
--- a/src/docdb/tree/DocDBStoredProceduresTreeItem.ts
+++ b/src/docdb/tree/DocDBStoredProceduresTreeItem.ts
@@ -5,7 +5,7 @@
 
 import { Container, CosmosClient, FeedOptions, QueryIterator, Resource, StoredProcedureDefinition } from '@azure/cosmos';
 import * as vscode from "vscode";
-import { AzExtTreeItem, ICreateChildImplContext } from 'vscode-azureextensionui';
+import { ICreateChildImplContext } from 'vscode-azureextensionui';
 import { defaultStoredProcedure, getThemeAgnosticIconPath } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { GraphCollectionTreeItem } from '../../graph/tree/GraphCollectionTreeItem';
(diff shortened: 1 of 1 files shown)


src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,14 +71,29 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
 
-    private validateName(name: string): string | null | undefined {
+    private async validateStoredProcedureName(name: string, getChildrenTask: Promise<AzExtTreeItem[]>): Promise<string | null | undefined> {
+        const currStoredProcedureList = await getChildrenTask;
+        const currStoredProcedureNames: string[] = [];
+        for (const sp of currStoredProcedureList) {
+            if (sp instanceof DocDBStoredProcedureTreeItem) {
+                currStoredProcedureNames.push(sp.id);
+            }
+        }
```
wwlorey, line 81: I think it would be more efficient to do this work in `createChildImpl`. That way we only have to get the existing stored procedures once instead of every time `validateStoredProcedureName` is called

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,14 +71,29 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
         if (name) {
-            if (name.indexOf("/") !== -1 || name.indexOf("\\") !== -1 || name.indexOf("?") !== -1 || name.indexOf("#") !== -1) {
-                return "Id contains illegal chars: /,\\,?,#";
+            if (name.length < 1 || name.length > 255) {
+                return "Name has to be between 1 and 255 chars long";
+            }
+            if (/[/\\?#&]/.test(name)) {
+                return localize("illegalChars", "Name contains illegal chars: /,\\,?,#,&");
```
wwlorey, line 87: From https://github.com/microsoft/vscode-cosmosdb/issues/1719:

> Succeed to create a new stored procedure with name as "&&&" without any error in portal.

It sounds like we should allow `&` if the portal does

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,14 +71,29 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
             }
             if (name[name.length - 1] === " ") {
-                return "Id ends with a space.";
+                return localize("endsWithSpace", "Name ends with a space.");
```
wwlorey, line 90: nit: I think it makes more sense to add a "cannot" here
```suggestion
                return localize("endsWithSpace", "Name cannot end with a space.");
```

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
neelip: The portal throws an error while creating a sp containing "&" in the name. Even if the extension allows creating sp with "&" in the name, there is an error while deleting it in the portal. 
image

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,14 +77,22 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
-    private validateName(name: string): string | null | undefined {
+    private async validateStoredProcedureName(name: string, currStoredProcedureNames: String[]): Promise<string | null | undefined> {
         if (name) {
-            if (name.indexOf("/") !== -1 || name.indexOf("\\") !== -1 || name.indexOf("?") !== -1 || name.indexOf("#") !== -1) {
-                return "Id contains illegal chars: /,\\,?,#";
+            if (name.length < 1 || name.length > 255) {
+                return "Name has to be between 1 and 255 chars long";
+            }
+            if (/[/\\?#&]/.test(name)) {
+                return localize("illegalChars", "Name contains illegal chars: /,\\,?,#,&");
```
wwlorey, line 86: nit: I think adding a space between each invalid character makes this look a bit more approachable

| Current | Suggested |
|-|-|
| Screen Shot 2021-02-09 at 5 02 15 PM | Screen Shot 2021-02-09 at 5 00 29 PM |

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,14 +77,22 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
             }
             if (name[name.length - 1] === " ") {
-                return "Id ends with a space.";
+                return localize("endsWithSpace", "Name cannot end with a space.");
+            }
+            if (currStoredProcedureNames.includes(name)) {
+                return localize('NameExists', 'Stored Procedure "{0}" already exists.', name);
             }
+        } else {
+            return localize("noEmptyValue", "Name cannot be empty");
```
wwlorey, line 95: We can get rid of this because it's already covered by the `if (name.length < 1 || name.length > 255)` check

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -38,9 +39,16 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
 
     public async createChildImpl(context: ICreateChildImplContext): Promise<DocDBStoredProcedureTreeItem> {
         const client = this.root.getCosmosClient();
+        const currStoredProcedureList = await this.getCachedChildren(context);
+        const currStoredProcedureNames: string[] = [];
+        for (const sp of currStoredProcedureList) {
+            if (sp instanceof DocDBStoredProcedureTreeItem) {
```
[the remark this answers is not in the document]
neelip: You right, it is not necessary here. The list should only contain `DocDBStoredProcedureTreeItem`.

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
neelip: I think we can remove `name.length < 1` from the first condition as "noEmptyValue" covers it. But checking for length will require `name` to not be null.

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
neelip: Should we keep the commas?

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -77,23 +75,21 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
-            if (name[name.length - 1] === " ") {
-                return localize("endsWithSpace", "Name cannot end with a space.");
-            }
-            if (currStoredProcedureNames.includes(name)) {
-                return localize('NameExists', 'Stored Procedure "{0}" already exists.', name);
-            }
-        } else {
-            return localize("noEmptyValue", "Name cannot be empty");
+    private validateStoredProcedureName(name: string, currStoredProcedureNames: String[]): string | undefined {
+        if (!name || name.length < 1 || name.length > 255) {
```
wwlorey, line 79: You don't have to check `!name` here since it's type `string` and not something like `string | undefined`

src/docdb/tree/DocDBStoredProceduresTreeItem.ts
```diff
@@ -69,15 +75,21 @@ export class DocDBStoredProceduresTreeItem extends DocDBTreeItemBase<StoredProce
...
 
-    private validateName(name: string): string | null | undefined {
-        if (name) {
-            if (name.indexOf("/") !== -1 || name.indexOf("\\") !== -1 || name.indexOf("?") !== -1 || name.indexOf("#") !== -1) {
-                return "Id contains illegal chars: /,\\,?,#";
-            }
-            if (name[name.length - 1] === " ") {
-                return "Id ends with a space.";
-            }
+    private validateStoredProcedureName(name: string, currStoredProcedureNames: String[]): string | undefined {
```
wwlorey, line 78: nit:
```suggestion
    private validateStoredProcedureName(name: string, currStoredProcedureNames: string[]): string | undefined {
```

wwlorey (approved):

PR #1816, opened 2021-02-09
Turn off "autoSave" for Workspace to fix issue with Codespaces

### LINKED ISSUE #1733: [Codespaces] A notification "Saving '***' will update the entity..." pops up when not saving for 'cosmos-collection.json' file
**OS:** Win10
**Build Version:** 0.15.0 (latest release build)

**Repro Steps:**

1. Drill down from MongoDB Account -> Database -> Collection.
2. Right-click a collection -> Click 'Open Collection' -> Make changes in the opend 'cosmos-collection.json' file.
3. Don't save and check the result.

**Expect:**
No notification pops up.

**Actual:**
A notification "Saving '***' will update the entity..." pops up.
coll

**More Info:**
This issue also reproduces when making changes for the opened "cosmos-stored procedure/document.json" file.
coll1

neelip:
Fixes #1733

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -66,6 +66,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
         ext.fileSystem = new DatabasesFileSystem(ext.tree);
         context.subscriptions.push(vscode.workspace.registerFileSystemProvider(DatabasesFileSystem.scheme, ext.fileSystem));
+        vscode.workspace.getConfiguration().update(ext.settingsKeys.vsCode.autoSave, "off");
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -66,6 +66,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
         ext.fileSystem = new DatabasesFileSystem(ext.tree);
         context.subscriptions.push(vscode.workspace.registerFileSystemProvider(DatabasesFileSystem.scheme, ext.fileSystem));
+        vscode.workspace.getConfiguration().update(ext.settingsKeys.vsCode.autoSave, "off");
```
nturinski, line 69: We don't want to just turn this setting off without alerting a user.  Also, I believe we only want to change the setting if the user is running in Codespaces.

I think we should check to see the platform, and then alert the user that we changed the setting because it will cause problems to leave it on in Codespaces.

PR #1821, opened 2021-02-16
Use "Report Issue" command

ejizba:
instead of button


nturinski (approved):
😇 I'm sure you know what my nit would be

PR #1823, opened 2021-02-17
Depend on Resource Groups extension

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/852


nturinski (approved):

neelip:
Should we need to update the README as well?

ejizba:
> Should we need to update the README as well?

Did you have something in mind?

neelip:
No, I don't see a need for any changes in the README for Databases, but just wanted check.

PR #1826, opened 2021-02-18
RC for 0.16.0

neelip:


CHANGELOG.md
```diff
@@ -2,12 +2,17 @@
 
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
-## Unreleased
+## 0.16.0 - 2021-02-23
 ### Added
 - Now depends on the "Azure Resources" extension, which provides a "Resource Groups" and "Help and Feedback" view
+- "Attached Emulator..." option added to Attach Database Accounts contest menu
```
nturinski, line 8: context* menu

But I'm not sure if this is something that we need to include in the "Added" section.

package-lock.json
```diff
@@ -1,13134 +1,8 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.15.2-alpha",
-    "lockfileVersion": 2,
+    "version": "0.16.0",
+    "lockfileVersion": 1,
```
wwlorey, line 4: It looks like you need to update npm to v7 so this lockfile stays v2

CHANGELOG.md
```diff
@@ -5,11 +5,11 @@ All notable changes to the "azure-cosmosdb" extension will be documented in this
 ## 0.16.0 - 2021-02-23
 ### Added
 - Now depends on the "Azure Resources" extension, which provides a "Resource Groups" and "Help and Feedback" view
```
neelip, line 7: @ejizba I just noticed this, is this supposed to be a continuation of this line?
""Report an Issue" button was removed from errors. Use the "Help and Feedback" view or command palette instead"

ejizba: I'm not entirely sure what you're asking, but I'll just say the way I did it was on purpose. If you give me a "suggested change", then I might be able to explain more

CHANGELOG.md
```diff
@@ -2,12 +2,17 @@
...
 All notable changes to the "azure-cosmosdb" extension will be documented in this file.
 
-## Unreleased
+## 0.16.0 - 2021-02-23
 ### Added
 - Now depends on the "Azure Resources" extension, which provides a "Resource Groups" and "Help and Feedback" view
 
 ### Changed
 - "Report an Issue" button was removed from errors. Use the "Help and Feedback" view or command palette instead
+- Grouped items in all context menus to improve access and readability
```
ejizba, line 11: Here is how I view each section of the changelog:

- Added: new features or enhancements
- Changed: Removed an old feature/enhancement, but replaced it with something else
- Removed: Removed an old feature/enhancement with no replacement

I think both of your bullets would fit better in "Added"

package-lock.json
```diff
@@ -4828,7 +4823,6 @@
         },
         "node_modules/fsevents/node_modules/balanced-match": {
             "version": "1.0.0",
-            "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
```
ejizba, line 4831: What version of npm are you on? I'm on 7.5.3. We shouldn't see a bunch of changes to this file

neelip: I am on 7.5.5

CHANGELOG.md
neelip: I mean the starting of the sentence "Now depends..." didn't make sense. I was thinking something in the lines of "This extension now depends..." But I guess it is implied.

package-lock.json
ejizba: Okay I updated to the latest (7.5.6), checked out this branch, and it will add back all the "resolved" lines when I run "npm install"

@wwlorey @nturinski can y'all try what I tried and see if you see the same behavior?

@neelip you can also try deleting the "node_modules" folder and _then_ running "npm install". Maybe you have something cached causing problems

wwlorey: I'm on npm version 7.5.6 and I deleted node_modules before npm installing. I didn't see the resolved lines get added back in but it did remove some integrity lines: [diff.txt](https://github.com/microsoft/vscode-cosmosdb/files/6025503/diff.txt)

CHANGELOG.md
nturinski: Rather than putting it in the "Changed" section, I meant I'm not sure this constitutes as a feature/enhancement that belongs in the `CHANGELOG`.  It's a pretty low impact change and we usually reserve the CHANGELOG for bigger features/enhancements we want people to discover.

package-lock.json
neelip: I am noticing that too. I see all the resolved line added back.

ejizba (approved):

PR #1827, opened 2021-02-18
Add resource groups extension to recommended for workspace

ejizba:
Already did it in most repos yesterday, just not the first two.


nturinski (approved):

nturinski (commented):
Does it need to be recommended if it's a dependency though?

ejizba:
> Does it need to be recommended if it's a dependency though?

When we F5 as devs, yes. For our customers, this has no effect

PR #1841, opened 2021-03-31
Bump extension version to 0.16.1-alpha

neelip:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.16.0",
+    "version": "0.16.1-alpha",
```
ejizba, line 3: There's another one on line 8. You can manually update both, but often I prefer to just run `npm install` and let it do it for me

ejizba (approved):

nturinski:
@neelip I think this fell off the radar a bit!  Could you run the npm command and merge this?  We don't want to keep getting telemetry from our testing to come in as 0.16.0.

PR #1844, opened 2021-04-05
Switch to new bots 🤖

ejizba:


.github/workflows/needs-more-info-closer.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: npm install --production --prefix ./azuretools/triage-actions
+      - name: Run Needs More Info Closer
+        uses: ./azuretools/triage-actions/needs-more-info-closer
+        with:
+          token: ${{secrets.AZCODE_BOT_PAT}}
+          label: needs more info
+          closeDays: 14
+          closeComment: "This issue has been closed automatically because it needs more information and has not had recent activity. See also our [issue reporting](https://aka.ms/azcodeissuereporting) guidelines.\n\nHappy Coding!"
+          pingDays: 80
+          pingComment: "Hey @${assignee}, this issue might need further attention.\n\n@${author}, you can help us out by closing this issue if the problem no longer exists, or adding more information."
```
nturinski, line 27: What happens if there is no assignee anyway?

nturinski (approved):

.github/workflows/needs-more-info-closer.yml
ejizba: It just doesn't ping anyone

https://github.com/microsoft/vscode-azuretools/blob/main/triage-actions/needs-more-info-closer/NeedsMoreInfoCloser.ts#L57

PR #1847, opened 2021-04-15
The execution order of the Mocha

v-wuzhai:


test/nightly/deleteAccount.ts
```diff
@@ -0,0 +1,61 @@
...
+    async function deleteAccount(accountType: string): Promise<void> {
+        const getAccount: CosmosDBManagementModels.DatabaseAccountGetResults = await client.databaseAccounts.get(resourceGroupList[accountType], accountList[accountType]);
+        accountItem[accountType] = getAccount;
+        const testInputs: string[] = [`${accountList[accountType]} (${accountType})`, DialogResponses.deleteResponse.title];
+        await testUserInput.runWithInputs(testInputs, async () => {
+            await vscode.commands.executeCommand('cosmosDB.deleteAccount');
+        });
+    }
+
+    async function delayDeleteAccount(ms: number, accountType: string): Promise<void> {
```
nturinski, line 48: I think you could combine this and `createAccount` to be a generic function.  Just pass in the callback (`deleteAccount` or `createAccount`)

nturinski (commented):
I believe the main reason for the amazing performance increase you've made is due to parallel account deletes.  

Rather than doing all of the requires in the `global.test` file, I think if you named the `deleteAccount.test.ts` file to something like `zzz_deleteAccount.test.ts`, it will always execute last.   Then you could run all the delete account actions in parallel.

I did a little experiment here: https://github.com/microsoft/vscode-cosmosdb/commit/9bc97588a6135194af1338247909f644a01f68df

nturinski:
I discussed this offline with Eric.  We both agree that we don't really want to use the "require" statements, but we found a pretty good workaround.

If you do all of the `deleteAccount` actions in the `suiteTeardown` of `global.resource.test.ts `, you can guarantee that it will be run last.  Then you can run the `accountDoesExist` command for all of the accounts that were supposed to be deleted.  I confirmed that if an assert fails in the `suiteTeardown`, the build will fail like so:

image

So in conclusion, please make the following changes:
- Please revert all the file names back to what they were 
- Move the `createAccount` promises back to the `suiteSetup` of `global.resource.test.ts`
- Add the `deleteAccount` promises to the `suiteTeardown` of `global.resource.test.ts`
- Change the delay command back to what it was, but change the `ms` parameter to `s`

test/nightly/global.resource.test.ts
```diff
@@ -31,15 +33,27 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
     if (longRunningTestsEnabled) {
         this.timeout(10 * 60 * 1000);
-        await deleteResourceGroups();
-        ext.azureAccountTreeItem.dispose();
+
+        // Delete account
+        await Promise.all([delayOpAccount(5, /MongoDB/, deleteAccount), delayOpAccount(10, /SQL/, deleteAccount)]);
+        try {
+            // If two or more of the following asserts fail, only one error will be thrown as a result.
+            await doesAccountExsit(AccountApi.MongoDB);
```
nturinski, line 48:
```suggestion
            await doesAccountExist(AccountApi.MongoDB);
```

test/nightly/global.resource.test.ts
```diff
@@ -31,15 +33,27 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
+            await doesAccountExsit(AccountApi.Core);
```
nturinski, line 49:
```suggestion
            await doesAccountExist(AccountApi.Core);
```

test/nightly/global.resource.test.ts
```diff
@@ -70,16 +84,32 @@ async function createAccount(accountType: RegExp): Promise<void> {
     });
 }
 
-async function delayCreateAccount(ms: number, accountType: RegExp): Promise<void> {
+async function doesAccountExsit(accountType: string): Promise<void> {
```
nturinski, line 87:
```suggestion
async function doesAccountExist(accountType: string): Promise<void> {
```

test/nightly/global.resource.test.ts
```diff
@@ -31,15 +33,27 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
 });
 
 suiteTeardown(async function (this: Mocha.Context): Promise<void> {
     if (longRunningTestsEnabled) {
         this.timeout(10 * 60 * 1000);
-        await deleteResourceGroups();
-        ext.azureAccountTreeItem.dispose();
+
+        // Delete account
+        await Promise.all([delayOpAccount(5, /MongoDB/, deleteAccount), delayOpAccount(10, /SQL/, deleteAccount)]);
```
nturinski, line 45: We should delete the Graph account in here as well.

test/nightly/global.resource.test.ts
nturinski, line 48: Rather than calling `doesAccountExist` on every type of AccountApi, you should be able to iterate over the keys of `AccountList`.  I don't think that the `accountType` should matter when deleting the accounts and should be able to just pass the name of the account as a string.

nturinski (commented):
I have no idea why, but this test is failing with your changes:
https://github.com/microsoft/vscode-cosmosdb/blob/328a2cb675cedb012869411cdd739587363a55fa/test/mongoShell.test.ts#L276

https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=35151&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=3245d04e-8e5b-52c0-c467-eb348b13e2ce

I'll try to help investigate why, but thought I'd let you know too.

nturinski (approved):
🎉🥳

Issue #1682, opened 2020-09-04
Support for serverless


markjbrown:
Cosmos now has support for serverless. Can you change the create workflow to allow a customer to select serverless for throughput? Also would need to be able to detect this during database or collection creation to not show throughput options.

ThomasWeiss:
We would need to:

- Add a "capacity model" choice when creating Core, Mongo, Table or Gremlin account types; options are "provisioned throughput" or "serverless"
  ![image](https://user-images.githubusercontent.com/4797533/108789899-33086400-7530-11eb-9637-ea59b59a8745.png)
- Modify the container creation flow to make throughput optional - and prompt that no throughput should be passed in a serverless account.
  ![image](https://user-images.githubusercontent.com/4797533/108789951-58956d80-7530-11eb-9143-cdaaeb17efa7.png)

PR #1852, opened 2021-04-24
feat: add CosmosDB serverless

christopheranderson:
Addresses #1682 

## Changes

1. Add capacity model to Cosmos DB API options w/ Provisioned and Serverless
2. Only prompt for RUs on account create for non-serverless accounts

## Contributors
Cosmos DB - @ThomasWeiss, @christopheranderson

### DIFF
--- a/src/AzureDBExperiences.ts
+++ b/src/AzureDBExperiences.ts
@@ -22,6 +22,10 @@ export enum DBAccountKind {
 
 export type CapabilityName = 'EnableGremlin' | 'EnableTable';
 
+export type CapacityModelName = 'Provisioned' | 'Serverless';
+
+export const SERVERLESS_CAPABILITY_NAME = 'EnableServerless';
+
 export function getExperienceFromApi(api: API): Experience {
     let info = experiencesMap.get(api);
     if (!info) {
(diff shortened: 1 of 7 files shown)


src/tree/AzureDBAPIStep.ts
```diff
@@ -29,6 +29,18 @@ export class AzureDBAPIStep extends AzureWizardPromptStep<IPostgresServerWizardC
             placeHolder: localize('selectDBServerMsg', 'Select an Azure Database Server.')
         });
 
+        if (result.data?.api !== API.Postgres) {
+            const modelPicks: IAzureQuickPickItem<CapacityModelName>[] = [
+                {label: "Provisioned Throughput", data: "Provisioned"},
+                {label: "Serverless", data: "Serverless"}
+            ];
+            const modelResult: IAzureQuickPickItem<CapacityModelName> = await ext.ui.showQuickPick(modelPicks, {
```
ejizba, line 37: You need to create a separate `AzureWizardPromptStep` for this otherwise the back button won't work properly. Each call to `showQuickPick` or `showInputBox` should be it's own step

test/nightly/global.resource.test.ts
```diff
@@ -31,7 +33,12 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
         client = createAzureClient(testAccount.getSubscriptionContext(), CosmosDBManagementClient);
 
         // Create account
-        await Promise.all([delayCreateAccount(5, /graph/), delayCreateAccount(10, /MongoDB/), delayCreateAccount(15, /SQL/)]);
+        await Promise.all([
```
ejizba, line 36: We're already having problems with these nightly tests taking way too long. Maybe just keep the same number of accounts and only make one (Mongo or SQL) serverless instead of provisioned.

That being said, don't worry too much about the nightly tests. They're annoying to test locally and I think you're blocked from running them in our Azure Dev Ops instance. I'd be fine if you left this up to us.

src/AzureDBExperiences.ts
```diff
@@ -45,15 +49,24 @@ export function getExperienceLabel(databaseAccount: DatabaseAccountGetResults):
...
+    }
+
+    // Check for serverless capability
+    if (!!experience && resource.capabilities?.find(cap => cap.name === SERVERLESS_CAPABILITY_NAME)) {
+        experience.capacityModel = "Serverless";
+    } else if (!!experience) {
+        experience.capacityModel = "Provisioned";
     }
 
     return undefined;
```
ejizba, line 72: This should be `return experience;`, right?

neelip:
@christopheranderson We have directed this update to new PR #1868. Please check there for all updates.

Issue #1857, opened 2021-05-06
CosmosDB: enableEndpointDiscovery must be configurable


cmendible:
Hi team!

With the current implementation of the getCosmosClient function: 

https://github.com/microsoft/vscode-cosmosdb/blob/252441e1663f3a638bd0eeb75e46b3dc7ddfe710/src/docdb/getCosmosClient.ts#L12

It´s not possible to connect with a CosmosDB database that is protected via private endpoint and also behind an Azure Application Gateway because the default connection policy enables endpoint discovery: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.connectionpolicy.enableendpointdiscovery?view=azure-dotnet

The extension should provide a way to disable endpoint discovery, using code similar to the one shown in the following snippet:

``` node
var enableEndpointDiscovery = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.enableEndpointDiscovery);
const connectionPolicy = { enableEndpointDiscovery: (enableEndpointDiscovery === undefined) ? true : enableEndpointDiscovery };
return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent(), agent: new https.Agent({ rejectUnauthorized: isEmulator ? !isEmulator : vscodeStrictSSL }), connectionPolicy: connectionPolicy });
```

cmendible:
@nturinski can we close the issue since #1858 was already merged?

nturinski:
Yeah, thanks for reminding me and for your contribution!

PR #1858, opened 2021-05-06
Support for CosmosDB's enableEndpointDiscovery #1857

cmendible:
Support for CosmosDB's enableEndpointDiscovery as requested in: #1857

### DIFF
--- a/src/docdb/getCosmosClient.ts
+++ b/src/docdb/getCosmosClient.ts
@@ -12,8 +12,8 @@ import { ext } from "../extensionVariables";
 export function getCosmosClient(endpoint: string, key: string, isEmulator: boolean | undefined): CosmosClient {
 
     const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
-    var enableEndpointDiscovery = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.enableEndpointDiscovery);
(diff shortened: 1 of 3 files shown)


src/docdb/getCosmosClient.ts
```diff
@@ -12,6 +12,8 @@ import { ext } from "../extensionVariables";
 export function getCosmosClient(endpoint: string, key: string, isEmulator: boolean | undefined): CosmosClient {
 
     const vscodeStrictSSL: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.vsCode.proxyStrictSSL);
-    return new CosmosClient({ endpoint, key, userAgentSuffix: appendExtensionUserAgent(), agent: new https.Agent({ rejectUnauthorized: isEmulator ? !isEmulator : vscodeStrictSSL }) });
+    const enableEndpointDiscovery: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>(ext.settingsKeys.enableEndpointDiscovery);
+    const connectionPolicy = { enableEndpointDiscovery: (enableEndpointDiscovery === undefined) ? true : enableEndpointDiscovery };
```
nturinski, line 16: Sorry, I'm not super familiar with this, but from this issue https://github.com/microsoft/vscode-cosmosdb/issues/1857, my understanding is that we want the user to be able to set this to be `false`, correct?  And the default behavior is that it is usually set to `true`?  

In the documentation linked in the issue and in the SDK index.d.ts, `enableEndpointDiscovery` is an optional property, so I would have assumed that it would default to false.  But I just want to make sure that the point of this setting is to allow users to set this to `false` if their private endpoint requires them to.

nturinski (approved):
Discussed offline; good to merge.

PR #1867, opened 2021-06-09
Updated Postgres connection certificates

### LINKED ISSUE #1766: Root certificate "Baltimore Root CA" for Postgres connection expires in Feb 2021
image
https://docs.microsoft.com/en-us/azure/postgresql/concepts-certificate-rotation

neelip:
Fixes #1766 

There is no immediate change to the root certificate but eventually BaltimoreCyberTrustRoot will be replaced with DigiCertGlobalRootG2. 

There is an option to add both the certs to the connection config. We can do this now and avoid sudden outages if and when the BaltimoreCyberTrustRoot will be replaced.


nturinski (approved):
I did a brief run-through of how the CAs are used and to make sure it's valid to pass multiple certificates into the [ConnectionOptions](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_tls_d_._tls_.securecontextoptions.html).  Seems good to me.

PR #1868, opened 2021-06-10
Added serverless capacity model to CosmosDB create

### LINKED ISSUE #1682: Support for serverless

neelip:
Fixes #1682

### DIFF
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ContainerDefinition, ContainerResponse, CosmosClient, DatabaseDefinition, FeedOptions, QueryIterator, Resource } from '@azure/cosmos';
+import { ContainerDefinition, ContainerResponse, CosmosClient, DatabaseDefinition, FeedOptions, QueryIterator, RequestOptions, Resource } from '@azure/cosmos';
 import * as vscode from 'vscode';
 import { AzureTreeItem, DialogResponses, ICreateChildImplContext, TreeItemIconPath, UserCancelledError } from 'vscode-azureextensionui';
 import { ext } from '../../extensionVariables';
@@ -96,16 +96,20 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Contai
                 paths: [partitionKey]
             };
         }
-        const isFixed: boolean = !(containerDefinition.partitionKey);
-        const minThroughput = isFixed ? minThroughputFixed : minThroughputPartitioned;
-        const throughput: number = Number(await ext.ui.showInputBox({
-            value: minThroughput.toString(),
-            ignoreFocusOut: true,
-            prompt: `Initial throughput capacity, between ${minThroughput} and ${maxThroughput}`,
-            validateInput: (input: string) => validateThroughput(isFixed, input)
-        }));
+        const options: RequestOptions = {};
 
-        const options = { offerThroughput: throughput };
+        if (!this.parent.isServerless) {
+            const isFixed: boolean = !(containerDefinition.partitionKey);
+            const minThroughput = isFixed ? minThroughputFixed : minThroughputPartitioned;
+            const throughput: number = Number(await ext.ui.showInputBox({
+                value: minThroughput.toString(),
+                ignoreFocusOut: true,
+                prompt: `Initial throughput capacity, between ${minThroughput} and ${maxThroughput}`,
+                validateInput: (input: string) => validateThroughput(isFixed, input)
+            }));
+
+            options.offerThroughput = throughput;
+        }
 
         context.showCreatingTreeItem(containerName);
         const client = this.root.getCosmosClient();
--- a/src/tree/AzureDBAPIStep.ts
+++ b/src/tree/AzureDBAPIStep.ts
@@ -16,6 +16,7 @@ import { PostgresServerNameStep } from '../postgres/commands/createPostgresServe
 import { PostgresServerSetCredentialsStep } from '../postgres/commands/createPostgresServer/steps/PostgresServerSetCredentialsStep';
 import { PostgresServerSetFirewallStep } from '../postgres/commands/createPostgresServer/steps/PostgresServerSetFirewallStep';
 import { localize } from '../utils/localize';
+import { CosmosDBAccountCapacityStep } from './CosmosDBAccountWizard/CosmosDBAccountCapacityStep';
 import { CosmosDBAccountCreateStep } from './CosmosDBAccountWizard/CosmosDBAccountCreateStep';
 import { CosmosDBAccountNameStep } from './CosmosDBAccountWizard/CosmosDBAccountNameStep';
 import { ICosmosDBWizardContext } from './CosmosDBAccountWizard/ICosmosDBWizardContext';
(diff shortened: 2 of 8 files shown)


src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    public async prompt(wizardContext: ICosmosDBWizardContext): Promise<void> {
+
+        const placeHolder: string = localize('selectDBServerMsg', 'Select a capacity model.')
+
+        wizardContext.isServerless = (await ext.ui.showQuickPick(
+            [
+                { label: localize('provisionedOption', 'Provisioned Throughput'), data: false },
+                { label: localize('serverlessOption', 'Serverless'), data: true }
+            ], { placeHolder })).data;
```
nturinski, line 21: It might be worth adding a "Learn More" link here.  I know the portal uses the same verbage, but provisioned throughput and serverless feel like they don't mean a whole lot to me.

The portal uses this one: https://aka.ms/cosmos-models

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,30 @@
...
+import { AzureWizardPromptStep } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../utils/localize';
+import { ICosmosDBWizardContext } from './ICosmosDBWizardContext';
+
+export class CosmosDBAccountCapacityStep extends AzureWizardPromptStep<ICosmosDBWizardContext> {
+
+    public async prompt(wizardContext: ICosmosDBWizardContext): Promise<void> {
+
+        const placeHolder: string = localize('selectDBServerMsg', 'Select a capacity model.')
```
nturinski, line 15: Not sure if we have to go with what the portal says, but FWIW

> Choose the capacity mode for your database operations.

src/tree/AzureDBAPIStep.ts
```diff
@@ -51,7 +52,8 @@ export class AzureDBAPIStep extends AzureWizardPromptStep<IPostgresServerWizardC
             ];
         } else {
             promptSteps = [
-                new CosmosDBAccountNameStep()
+                new CosmosDBAccountCapacityStep(),
```
nturinski, line 55: It's odd that the capacity mode step comes before the accountNameStep.

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,30 @@
...
+            [
+                { label: localize('provisionedOption', 'Provisioned Throughput'), data: false },
+                { label: localize('serverlessOption', 'Serverless'), data: true }
+            ], { placeHolder })).data;
+
+    }
+
+
+    public shouldPrompt(wizardContext: ICosmosDBWizardContext): boolean {
+        return !wizardContext.isServerless;
```
nturinski, line 27:
```suggestion
        return wizardContext.isServerless === undefined
```

Otherwise false will be read as undefined as well.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
```diff
@@ -96,16 +96,20 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Contai
...
-        const minThroughput = isFixed ? minThroughputFixed : minThroughputPartitioned;
-        const throughput: number = Number(await ext.ui.showInputBox({
-            value: minThroughput.toString(),
-            ignoreFocusOut: true,
-            prompt: `Initial throughput capacity, between ${minThroughput} and ${maxThroughput}`,
-            validateInput: (input: string) => validateThroughput(isFixed, input)
-        }));
-
-        const options = { offerThroughput: throughput };
+        const options: RequestOptions = {};
```
nturinski, line 99: Could you explain why this change was made?

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
nturinski: You can do it like this: https://github.com/microsoft/vscode-azurestaticwebapps/blob/d09f669ba0208a66b2b56a1f20c13b4c7eab2dba/src/commands/createStaticWebApp/BuildPresetListStep.ts#L20

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
neelip: I think I would prefer "Select a capacity model." as it would consistent with other placeholders. 
The "Choose the capacity mode..." might confuse the user as capacity model is selected at the server even tho it is meant for database operations. 

What if we add a "Learn More..." option to this prompt instead?

src/tree/AzureDBAPIStep.ts
neelip: accountNameStep can go before CapacityStep, it shouldn't cause a problem. 

I was going for this setup:

1. "Select an Azure Database Server."
2. and pick the next categorization which is capacity model with "Select a capacity model" prompt.
3. and then start creating the server starting with the nameStep.

src/docdb/tree/DocDBDatabaseTreeItemBase.ts
neelip: Options need to be defined for `throughput` when working with non-serverless database instance, and options need to be empty when working with serverless.

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,46 @@
...
+export class CosmosDBAccountCapacityStep extends AzureWizardPromptStep<ICosmosDBWizardContext> {
+
+    public async prompt(wizardContext: ICosmosDBWizardContext): Promise<void> {
+
+        const placeHolder: string = localize('selectDBServerMsg', 'Select a capacity model.')
+        const picks: IAzureQuickPickItem<boolean | undefined>[] = [
+            { label: localize('provisionedOption', 'Provisioned Throughput'), data: false },
+            { label: localize('serverlessOption', 'Serverless'), data: true },
+        ];
+        const learnMore: IAzureQuickPickItem = { label: localize('learnMore', '$(link-external) Learn more...'), description: '', data: undefined };
```
nturinski, line 21: nit: You can just omit providing a description.

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,46 @@
...
+            if (pick === learnMore) {
+                await openUrl('https://aka.ms/cosmos-models');
+            }
+        } while (pick === learnMore);
+
+        if (pick.data) {
+            wizardContext.isServerless = pick.data;
+            wizardContext.telemetry.properties.isServerless = pick.data ? 'true' : 'false';
+        }
+
```
nturinski, line 36: nit: Lot of weird empty space here.

src/tree/CosmosDBAccountWizard/CosmosDBAccountCapacityStep.ts
```diff
@@ -0,0 +1,46 @@
...
+
+
+    }
+
+
```
nturinski, line 41: nit: Here as well

nturinski (approved):

PR #1872, opened 2021-06-18
Abstract the Postgres management client to allow both single and flexible server

tonybaloney:
- [x] Implement server listing
- [x] Implement connection string
- [x] Implement database list
- [x] Implement provisioning


tonybaloney:
screenshot 2021-06-18 at 11 05 04
server listing (both single and flexible)

tonybaloney:
The provisioning for flexible creates the server but then hits this issue-- 
```
3:34:09 PM: Creating PostgreSQL Server "ant-flexi-2"... It should be ready in several minutes.
3:34:44 PM: Error: The resource type 'locations/azureAsyncOperation' could not be found in the namespace 'Microsoft.DBforPostgreSQL' for api version '2021-04-10-privatepreview'. The supported api-versions are '2017-12-01-preview,2017-12-01,2018-03-29-privatepreview,2020-10-05-privatepreview,2020-02-14-privatepreview,2020-02-14-preview,2020-11-05-preview'.
```

tonybaloney:
Provisioning for single server always fails saying the password is too weak so I think the call mask is causing it to be replaced with *****

tonybaloney:
Again for deleting single servers

```
The resource type 'locations/azureAsyncOperation' could not be found in the namespace 'Microsoft.DBforPostgreSQL' for api version '2021-04-10-privatepreview'. The supported api-versions are '2017-12-01-preview,2017-12-01,2018-03-29-privatepreview,2020-10-05-privatepreview,2020-02-14-privatepreview,2020-02-14-preview,2020-11-05-preview'.
```

src/postgres/commands/createPostgresServer/steps/PostgresServerSkuStep.ts
```diff
@@ -150,3 +152,28 @@ const availableSkus: ISkuOption[] = [
         },
     },
 ];
+
+// Official storage sizes are 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216
+const flexibleServerSkus: ISkuOption[] = [
+    {
+        label: "B1ms",
```
wwlorey, line 159: Is there any info you could add to the description that would account for why "ms" and "s" are included in the labels? Or is this common SKU naming that people will be familiar with? It just looks a little strange to me
Screen Shot 2021-06-22 at 12 06 46 PM

src/postgres/abstract/AbstractPostgresClient.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import { PostgreSQLManagementClient } from "@azure/arm-postgresql";
+import { PostgreSQLFlexibleManagementClient } from "@azure/arm-postgresql-flexible";
+import * as msRest from "@azure/ms-rest-js";
+import { IMinimumServiceClientOptions } from "vscode-azureextensionui";
+import { IAbstractPostgresClient } from "./IAbstractPostgresClient";
+import { asAbstractDatabase, asFlexibleParameters, asSingleParameters, flexibleAsAbstractServer, singleAsAbstractServer } from "./maps";
+import * as Models from "./models";
+
+export class AbstractPostgresClient implements IAbstractPostgresClient  {
```
wwlorey, line 14: I'd be interested to hear what @microsoft/vscodeazuretoolsdev thinks about this but I think we can get rid of `IAbstractPostgresClient` and let this class stand on its own. The interface is useful if we're creating multiple classes that implement it but here it's only used with a single class. And the only other places where it's used are typings for `createAzureClient` calls which can all be replaced with type `AbstractPostgresClient`

src/tree/SubscriptionTreeItem.ts
```diff
@@ -35,13 +36,13 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
 
         //Postgres
-        const postgresClient: PostgreSQLManagementClient = createAzureClient(this.root, PostgreSQLManagementClient);
-        const postgresServers: ServerListResult = await postgresClient.servers.list();
+        const postgresClient: IAbstractPostgresClient = createAzureClient(this.root, AbstractPostgresClient);
+        const postgresServers: PostgresAbstractServerList = await postgresClient.listServers() as PostgresAbstractServerList;
```
wwlorey, line 40:
```suggestion
        const postgresServers: PostgresAbstractServerList = await postgresClient.listServers();
```

src/postgres/abstract/AbstractPostgresClient.ts
```diff
@@ -0,0 +1,83 @@
...
+            case Models.PostgresServerType.Flexible:
+                return this.postgresFlexibleClient.servers.deleteMethod(resourceGroup, name);
+            case Models.PostgresServerType.Single:
+                return this.postgresSingleClient.servers.deleteMethod(resourceGroup, name);
+            default:
+                throw new Error("Service not implemented.");
+        }
+    }
+
+    async listServers(): Promise<Models.PostgresAbstractServerList> {
```
wwlorey, line 67:
```suggestion
    async listAllServers(): Promise<Models.PostgresAbstractServerList> {
```

It doesn't have to be this exact name but I want to convey that it's listing flexible and single servers. And it's the only function in this class that doesn't just do one or the other

src/postgres/abstract/AbstractPostgresClient.ts
```diff
@@ -0,0 +1,83 @@
...
+            case Models.PostgresServerType.Flexible:
+                return this.postgresFlexibleClient.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
+            case Models.PostgresServerType.Single:
+                return this.postgresSingleClient.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
+            default:
+                throw new Error("Service not implemented.");
+        }
+    }
+
+    async createFirewallRule(serverType: Models.PostgresServerType, resourceGroup: string, name: string, ruleName: string, rule: Models.AbstractFirewallRule): Promise<Models.AbstractFirewallRule> {
```
wwlorey, line 34:
```suggestion
    async createOrUpdateFirewallRule(serverType: Models.PostgresServerType, resourceGroup: string, name: string, ruleName: string, rule: Models.AbstractFirewallRule): Promise<Models.AbstractFirewallRule> {
```

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -46,8 +47,7 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
         return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
     }
 
-    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
-    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
+    const availability: AbstractNameAvailability = (await client.checkNameAvailability(serverType, name));
```
wwlorey, line 50: nit: not sure why those parentheses were there before
```suggestion
    const availability: AbstractNameAvailability = await client.checkNameAvailability(serverType, name);
```

src/postgres/commands/createPostgresServer/steps/PostgresServerSkuStep.ts
```diff
@@ -150,3 +152,28 @@ const availableSkus: ISkuOption[] = [
         },
     },
 ];
+
+// Official storage sizes are 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216
```
wwlorey, line 156: Could you include a commented link to the source for this info?

src/tree/CosmosDBAccountWizard/CosmosDBAccountCreateStep.ts
```diff
@@ -30,7 +30,7 @@ export class CosmosDBAccountCreateStep extends AzureWizardExecuteStep<ICosmosDBW
         const options: DatabaseAccountCreateUpdateParameters = {
             location: locationName,
             locations: [{ locationName: locationName }],
-            kind: defaultExperience.kind,
+            kind: defaultExperience.kind as DatabaseAccountKind,
```
wwlorey, line 33:
```suggestion
            kind: defaultExperience.kind,
```

I don't think this is needed either. Was typescript complaining about this?

wwlorey (commented):
> The provisioning for flexible creates the server but then hits this issue--
> 
> ```
> 3:34:09 PM: Creating PostgreSQL Server "ant-flexi-2"... It should be ready in several minutes.
> 3:34:44 PM: Error: The resource type 'locations/azureAsyncOperation' could not be found in the namespace 'Microsoft.DBforPostgreSQL' for api version '2021-04-10-privatepreview'. The supported api-versions are '2017-12-01-preview,2017-12-01,2018-03-29-privatepreview,2020-10-05-privatepreview,2020-02-14-privatepreview,2020-02-14-preview,2020-11-05-preview'.
> ```

I'm running into this as well. Do you know if/when that api version will be supported?

src/postgres/abstract/AbstractPostgresClient.ts
neelip: I agree with Will on this. Right now, IAbstractPostgresClient is only used for declaring type. I think it is a good idea to remove it and just use AbstractPostgresClient, I don't see any harm in it.

src/postgres/commands/createPostgresServer/steps/PostgresServerSkuStep.ts
tonybaloney: This is how they're documented in the official docs https://docs.microsoft.com/en-au/azure/postgresql/flexible-server/concepts-compute-storage#compute-tiers-vcores-and-server-types

I don't know what m or s means.

tonybaloney:
@wwlorey updates completed from the review

wwlorey:
This looks good 👍👍  I'll wait until we hear back about the API problem before testing again & getting this merged

tonybaloney:
> This looks good 👍👍  I'll wait until we hear back about the API problem before testing again & getting this merged

There is an issue with the SDK release. I'm waiting for a patch and will update packages.json once it's done

src/postgres/abstract/AbstractPostgresClient.ts
```diff
@@ -0,0 +1,82 @@
...
+            case Models.PostgresServerType.Flexible:
+                return this.postgresFlexibleClient.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
+            case Models.PostgresServerType.Single:
+                return this.postgresSingleClient.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
+            default:
+                throw new Error("Service not implemented.");
+        }
+    }
+
+    async createOrUpdateFirewallRule(serverType: Models.PostgresServerType, resourceGroup: string, name: string, ruleName: string, rule: Models.AbstractFirewallRule): Promise<Models.AbstractFirewallRule> {
```
ejizba, line 33: The vast majority of these methods are just copy/pasted for both clients. Rather than duplicating the code, how about we do something like this:

```typescript
export type AbstractPostgresClient = PostgreSQLFlexibleManagementClient | PostgreSQLManagementClient;

export function createAbstractPostgresClient(serverType: PostgresServerType, clientInfo: ISubscriptionContext): AbstractPostgresClient {
    switch (serverType) {
        case PostgresServerType.Flexible:
            return createAzureClient(clientInfo, PostgreSQLFlexibleManagementClient);
        case PostgresServerType.Single:
            return createAzureClient(clientInfo, PostgreSQLManagementClient);
        default:
            throw new Error("Service not implemented.");
    }
}
```

As an example, `configurePostgresFirewall.ts` would then look like this:
```typescript
const client = createAbstractPostgresClient(treeItem.serverType, treeItem.root);
await client.firewallRules.createOrUpdate(resourceGroup, serverName, firewallRuleName, {
    startIpAddress: ip,
    endIpAddress: ip
});
```

src/postgres/abstract/AbstractPostgresClient.ts
```diff
@@ -0,0 +1,82 @@
...
+            case Models.PostgresServerType.Flexible:
+                return this.postgresFlexibleClient.firewallRules.createOrUpdate(resourceGroup, name, ruleName, rule);
+            case Models.PostgresServerType.Single:
+                return this.postgresSingleClient.firewallRules.createOrUpdate(resourceGroup, name, ruleName, rule);
+            default:
+                throw new Error("Service not implemented.");
+        }
+    }
+
+    async createServer(serverType: Models.PostgresServerType, resourceGroup: string, name: string, parameters: Models.AbstractServerCreate): Promise<Models.PostgresAbstractServer> {
```
ejizba, line 44: `createServer` and `listAllServers` are the only two methods on this class that aren't simple copy/paste and thus the only reason we _might_ need this class still. However, since each method is only used once, I don't think there's much value to putting them in a separate class like this. The true value of abstracting it away would be if we were calling the methods multiple times, which I don't see happening. I say we go with my `createAbstractPostgresClient` function above and move this code directly into `PostgresServerCreateStep`

src/postgres/abstract/maps.ts
```diff
@@ -0,0 +1,85 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as FlexibleModels from "@azure/arm-postgresql-flexible/esm/models";
+import * as SingleModels from "@azure/arm-postgresql/esm/models";
+import { AbstractServerCreate, PostgresAbstractDatabase, PostgresAbstractServer, PostgresServerType } from "./models";
+
+
+
+export function singleAsAbstractServer(server: SingleModels.Server) : PostgresAbstractServer {
```
ejizba, line 12: Rather than copy/pasting the properties like this, we can directly use the original object and a `type` instead of an `interface`.

The only complication with `PostgresAbstractServer` is `serverType`. Do we actually need that property or is there already some property that identifies if it's flexible/single? If we _do_ need the property, you can still use the original object, but just add `serverType` on top like this:
```typescript
export type PostgresAbstractServer = (SingleModels.Server | FlexibleModels.Server) & { serverType: PostgresServerType; }
```

```typescript
export function singleAsAbstractServer(server: SingleModels.Server): PostgresAbstractServer {
    return Object.assign(server, { serverType: PostgresServerType.Single });
}
```

src/postgres/abstract/models.ts
```diff
@@ -0,0 +1,135 @@
...
+    serverType: PostgresServerType;
+    /**
+     * The Azure Active Directory identity of the server.
+     */
+    identity?: ResourceIdentity;
+}
+
+export type PostgresAbstractServerList = Array<PostgresAbstractServer>;
+
+export interface PostgresAbstractDatabase {
```
ejizba, line 59:
```
export type PostgresAbstractDatabase = FlexibleModels.Database | SingleModels.Database;
```

src/postgres/abstract/models.ts
```diff
@@ -0,0 +1,135 @@
...
+ export interface AbstractSku {
+    /**
+     * The name of the sku, typically, tier + family + cores, e.g. B_Gen4_1, GP_Gen5_8.
+     */
+    name: string;
+    /**
+     * The tier of the particular SKU, e.g. Basic. Possible values include: 'Basic',
+     * 'GeneralPurpose', 'MemoryOptimized'
+     */
+    tier?: string;
```
ejizba, line 87:
```
tier?: FlexibleModels.SkuTier | SingleModels.SkuTier;
```

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -46,8 +47,7 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
         return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
     }
 
-    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
-    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
+    const availability: AbstractNameAvailability = await client.checkNameAvailability(serverType, name);
 
     if (!availability.nameAvailable) {
         if (availability.reason === 'AlreadyExists') {
```
ejizba, line 53: If I follow my suggested pattern for `type` vs. `interface`, it'll look something like this:
```typescript
export type AbstractNameAvailability = SingleModels.NameAvailability | FlexibleModels.NameAvailability;
```

However, all of the sudden this line starts complaining because `reason` doesn't exist on `FlexibleModels.NameAvailability`. That implies to me that either:
1. The types on `FlexibleModels.NameAvailability` are wrong
2. The types are correct and this logic doesn't actually work for flexible servers

If 2 is true, we need to update this logic. If 1 is true, we can update the type like so:
```typescript
export type AbstractNameAvailability = SingleModels.NameAvailability | (FlexibleModels.NameAvailability & {
    /**
    * NOTE: The types on `FlexibleModels.NameAvailability` are incorrect, this property does exist
    */
    reason?: string;
})
```

src/postgres/abstract/maps.ts
```diff
@@ -0,0 +1,85 @@
...
+            storageMB: parameters.storageMB
+        },
+        sku: {
+            name: parameters.sku.name,
+            tier: parameters.sku.tier as FlexibleModels.SkuTier
+        },
+    }
+}
+
+export function asSingleParameters(parameters: AbstractServerCreate) : SingleModels.ServerForCreate {
```
ejizba, line 64: If a function is only used once, we normally put it in the file where it's used, rather than in a generic file like "maps.ts" (unless we predict it being used multiple times in the future, but I don't see that happening here). I think `asFlexibleParameters` and `asSingleParameters` would fit better in `PostgresServerCreateStep.ts`

ejizba (commented):
Sorry for jumping in here a little late, but I think there's too much duplicated code inside `AbstractPostgresClient` and `models.ts`. We lose a bit of type safety and risk some silly copy/paste bugs. We should be able to leverage TypeScript features like union types more heavily to clean it up

tonybaloney:
> Sorry for jumping in here a little late, but I think there's too much duplicated code inside `AbstractPostgresClient` and `models.ts`. We lose a bit of type safety and risk some silly copy/paste bugs. We should be able to leverage TypeScript features like union types more heavily to clean it up

@ejizba thanks for the feedback. My TS knowledge is still pretty basic. This PR is still blocked on an SDK patch so I'll work through your feedback and update this PR.

src/postgres/abstract/maps.ts
tonybaloney: The models are very different. Doing this is throwing a range of compiler errors

```
Argument of type '(server: Server) => PostgresAbstractServer' is not assignable to parameter of type '((value: Server, index: number, array: Server[]) => PostgresAbstractServer) & ((value: Server, index: number, array: Server[]) => PostgresAbstractServer)'.
  Type '(server: Server) => PostgresAbstractServer' is not assignable to type '(value: Server, index: number, array: Server[]) => PostgresAbstractServer'.
    Types of parameters 'server' and 'value' are incompatible.
      Type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql/esm/models/index").Server' is not assignable to type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql-flexible/esm/models/index").Server'.
        Types of property 'sku' are incompatible.
          Type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql/esm/models/index").Sku | undefined' is not assignable to type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql-flexible/esm/models/index").Sku | undefined'.
            Type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql/esm/models/index").Sku' is not assignable to type 'import("/Users/anthonyshaw/projects/vscode-cosmosdb/node_modules/@azure/arm-postgresql-flexible/esm/models/index").Sku'.
              Types of property 'tier' are incompatible.
                Type 'SkuTier | undefined' is not assignable to type 'SkuTier'.
                  Type 'undefined' is not assignable to type 'SkuTier'.ts(2345)
```

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
tonybaloney: thats just because I referred to the reason instead of the message, this can be unionised.

tonybaloney:
@ejizba refactored as per your feedback in https://github.com/microsoft/vscode-cosmosdb/pull/1872/commits/bb01e7a8c287c2bc02309a8a423db06cb83572f9

src/postgres/abstract/models.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as SingleModels from "@azure/arm-postgresql/esm/models";
+
+export enum PostgresServerType {
+    Flexible,
+    Single
+}
+
+export type PostgresAbstractServer = (SingleModels.Server | FlexibleModels.Server) & { serverType?: PostgresServerType; }
+
+export type PostgresAbstractServerList = Array<PostgresAbstractServer>;
```
ejizba, line 16: Remove these "List" types, they're very simple and I don't think they're used

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -46,11 +47,10 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
         return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
     }
 
-    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
-    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
+    const availability: AbstractNameAvailability = await client.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
 
     if (!availability.nameAvailable) {
-        if (availability.reason === 'AlreadyExists') {
+        if (availability.message === 'AlreadyExists') {
```
ejizba, line 53: I don't think `message` is right - this is what I'm seeing
Screen Shot 2021-06-30 at 1 12 03 PM

That's for single server. Not sure what flexible looks like because of the sdk issue.

src/postgres/commands/createPostgresServer/steps/PostgresServerSkuStep.ts
```diff
@@ -40,9 +40,11 @@ export class PostgresServerSkuStep extends AzureWizardPromptStep<IPostgresServer
         return wizardContext.sku === undefined;
     }
 
-    public async getPicks(): Promise<IAzureQuickPickItem<Sku | undefined>[]> {
-        const options: IAzureQuickPickItem<Sku | undefined>[] = [];
-        availableSkus.forEach((option) => {
+    public async getPicks(serverType: PostgresServerType): Promise<IAzureQuickPickItem<AbstractSku | undefined>[]> {
+        const options: IAzureQuickPickItem<AbstractSku | undefined>[] = [];
+        const skuOptions: ISkuOption[] = serverType == PostgresServerType.Single ? singleServerSkus : flexibleServerSkus ;
```
ejizba, line 45: Always use `===` instead of `==`

I'm kind of surprised our linting doesn't catch this - thoughts @wwlorey
https://eslint.org/docs/rules/eqeqeq

src/postgres/abstract/models.ts
```diff
@@ -0,0 +1,60 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as FlexibleModels from "@azure/arm-postgresql-flexible/esm/models";
+import * as SingleModels from "@azure/arm-postgresql/esm/models";
```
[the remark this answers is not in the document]
tonybaloney: I've seen this pattern a few times in this module, I'll raise another PR with other occurrences.

tonybaloney:
Blocked by pending update to https://github.com/Azure/azure-sdk-for-js/issues/15979

tonybaloney:
Related https://github.com/Azure/azure-sdk-for-js/pull/16136

tonybaloney:
@wwlorey @ejizba this PR is now finished/ready for review. the backend dependency on the JS SDK has been fixed and I've updated package.json with the new version

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -2,17 +2,19 @@
...
 import { AzureWizardExecuteStep, callWithMaskHandling, createAzureClient, LocationListStep } from 'vscode-azureextensionui';
 import { ext } from '../../../../extensionVariables';
 import { localize } from '../../../../utils/localize';
 import { nonNullProp } from '../../../../utils/nonNull';
+import { AbstractServerCreate, PostgresServerType } from '../../../abstract/models';
 import { IPostgresServerWizardContext } from '../IPostgresServerWizardContext';
 
 export class PostgresServerCreateStep extends AzureWizardExecuteStep<IPostgresServerWizardContext> {
     public priority: number = 150;
+    public defaultVersion: string = "11";
```
ejizba, line 17: Any specific reason this default changed from 10 to 11? If we're going to update the default, I'd prefer that as a separate PR instead of hidden in this larger refactoring. I'm not super versed on postgres versions, so I'd wanna know what the portal uses as a default, what the support level is for 10 vs 11, etc.

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -46,13 +47,10 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
         return localize('serverNamePrefixSuffixCheck', 'Server name must not start or end in a hyphen.');
     }
 
-    const availabilityRequest: NameAvailabilityRequest = { name: name, type: "Microsoft.DBforPostgreSQL" };
-    const availability: NameAvailability = (await client.checkNameAvailability.execute(availabilityRequest));
+    const availability: AbstractNameAvailability = await client.checkNameAvailability.execute({name: name, type: "Microsoft.DBforPostgreSQL"});
```
ejizba, line 50: This isn't workin for me for flexible servers. It's silently failing under the covers because of this error:

> Unknown resource type specified for a subscription level name availability check. Supported resource types are: Microsoft.DBforPostgreSQL/flexibleServers.

src/postgres/commands/createPostgresServer/steps/PostgresServerNameStep.ts
```diff
@@ -46,13 +47,10 @@ async function validatePostgresServerName(name: string, client: PostgreSQLManage
...
 
     if (!availability.nameAvailable) {
-        if (availability.reason === 'AlreadyExists') {
-            return localize('serverNameAvailabilityCheck', 'A server named "{0}" already exists.', name);
-        }
+        return localize('serverNameAvailabilityCheck', 'Server name "{0}" is not available.', name);
```
ejizba, line 53: Let's just return `availability.message` - not sure why we weren't using that already

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
```diff
@@ -2,17 +2,19 @@
...
 
     public async execute(wizardContext: IPostgresServerWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
```
ejizba, line 19: I did a bit of a refactor and caused some merge conflicts for ya - sorry bout that, but please fix them :)

src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts
tonybaloney: The portal default is 11 for single server and 12 for flexible server. 

Single server supports 9.6-11
Flexible server supports 11-13

10 is no longer supported in flexible server, but still in single server (but no longer the default as it was when this was originally written).

I can make it a separate PR but would have to replace this with a switch to default to 10 for single and 11 for flexible

ejizba: Matching the portal would be best, aka 11 for single and 12 for flexible. I think I'm okay including this in the current PR.

src/AzureDBExperiences.ts
```diff
@@ -109,8 +110,9 @@ const CoreExperience: Experience = { api: API.Core, longName: "Core", descriptio
 export const MongoExperience: Experience = { api: API.MongoDB, longName: "Azure Cosmos DB for MongoDB API", shortName: "MongoDB", kind: DBAccountKind.MongoDB, tag: "Azure Cosmos DB for MongoDB API" };
 const TableExperience: Experience = { api: API.Table, longName: "Azure Table", shortName: "Table", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableTable', tag: "Azure Table" };
 const GremlinExperience: Experience = { api: API.Graph, longName: "Gremlin", description: "(graph)", shortName: "Gremlin", kind: DBAccountKind.GlobalDocumentDB, capability: 'EnableGremlin', tag: "Gremlin (graph)" };
-const PostgresExperience: Experience = { api: API.Postgres, longName: "PostgreSQL", shortName: "PostgreSQL" };
+const PostgresSingleExperience: Experience = { api: API.PostgresSingle, longName: "PostgreSQL Single Server", shortName: "PostgreSQLSingle"};
+const PostgresFlexibleExperience: Experience = { api: API.PostgresFlexible, longName: "PostgreSQL Flexible Server", shortName: "PostgreSQLFlexible"};
```
ejizba, line 114: When I was trying out the portal experience to check the default versions I noticed they label Flexible as "Preview". We should do the same

tonybaloney:
@ejizba amendments made in https://github.com/microsoft/vscode-cosmosdb/pull/1872/commits/cb353d473788089b30b04e7c23510df865c8eca8

ejizba (approved):
Looks good to me! @wwlorey were you still going to double check?

wwlorey (approved):
Looks good 👍

PR #1887, opened 2021-07-13
Add the "Delete graph database" test case

v-wuzhai:


neelip:
What is the reason for deleting this test case?

v-wuzhai:
@neelip I added an automation test case "Delete graph database" in this PR. I have changed the confused PR title.

nturinski:
@v-wuzhai I really appreciate the work you put into this PR, but as it is, Databases is already one of our longer running nightly tests (at ~32 minutes).    I think we want to avoid adding more tests that are not core scenario because while the tests are running, those resources are tied up.

We may consider adding automated Postgres create/delete tests, but we should discuss that as a team since as mentioned, Databases already takes a long time to run.

PR #1888, opened 2021-07-15
Funnel telemetry - updates

neelip:
Bump shared packages and add 'stepName'.

### DIFF
--- a/src/graph/tree/GraphTreeItem.ts
+++ b/src/graph/tree/GraphTreeItem.ts
@@ -75,11 +75,11 @@ export class GraphTreeItem extends AzureTreeItem<IDocDBTreeRoot> {
             } else {
                 const viewExt: vscode.MessageItem = { title: localize('viewExt', 'View Extension') };
                 const message: string = localize('mustInstallGraph', 'You must install the Cosmos DB Graph extension to view a graph.');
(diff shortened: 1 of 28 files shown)


src/graph/tree/GraphTreeItem.ts
```diff
@@ -75,11 +75,11 @@ export class GraphTreeItem extends AzureTreeItem<IDocDBTreeRoot> {
             } else {
                 const viewExt: vscode.MessageItem = { title: localize('viewExt', 'View Extension') };
                 const message: string = localize('mustInstallGraph', 'You must install the Cosmos DB Graph extension to view a graph.');
-                context.telemetry.properties.cancelStep = 'installGraphExtPreMessage';
-                await context.ui.showWarningMessage(message, { modal: true }, viewExt);
-                context.telemetry.properties.cancelStep = 'installGraphExtPostMessage';
+                context.telemetry.properties.lastStep = 'installGraphExtPreMessage';
```
ejizba, line 78: We don't have to explicitly set `lastStep` like we did for `cancelStep`. You can delete this line and the other similar line since you added `stepName` to `showWarningMessage` and `UserCancelledError`

ejizba (approved):
See here to fix build: https://github.com/microsoft/vscode-azureappservice/pull/2092

PR #1890, opened 2021-07-17
Update authentication related code to support Flexible Server

neelip:
Flexible Servers don't have usernames in this format --> username@servername, like Single Servers do: 
image

Flexible Servers support DigiCertGlobalRootCA root certificates, which is different from Single Servers. 
https://aka.ms/AAd75x5


src/postgres/commands/checkAuthentication.ts
```diff
@@ -22,7 +22,7 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
                 await enterPostgresCredentials(context, treeItem.parent);
 
                 // Need to configure firewall only for Azure Subscritption accounts
-            } else if (treeItem.parent.resourceGroup && parsedError.errorType === firewallNotConfiguredErrorType) {
+            } else if (treeItem.parent.resourceGroup && (parsedError.errorType === firewallNotConfiguredErrorType || parsedError.errorType === 'ETIMEDOUT')) {
```
neelip, line 25: Connecting to Flexible Servers is throwing a 'ETIMEDOUT' ERROR.

src/postgres/getClientConfig.ts
```diff
@@ -48,51 +50,6 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
         await client.connect();
         return clientConfig;
     } finally {
-        await client.end();
+        client.end;
     }
 }
-
-export const BaltimoreCyberTrustRoot: string = `-----BEGIN CERTIFICATE-----
```
ejizba, line 55: I liked it better when these were here. They're only used in this file and they're kind of long/ugly constants, so I don't think I wanna clutter the `constants.ts` file

src/postgres/commands/checkAuthentication.ts
```diff
@@ -24,10 +29,21 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
...
     }
     return clientConfig;
 }
+
+export async function isFirewallRuleSet(treeItem: PostgresServerTreeItem): Promise<boolean> {
+    const serverType: PostgresServerType = nonNullProp(treeItem, 'serverType');
+    const client = createAbstractPostgresClient(serverType, treeItem.root);
+    client.firewallRules.listByServer
+    const result: FirewallRuleListResult = (await client.firewallRules.listByServer(nonNullProp(treeItem, 'resourceGroup'), nonNullProp(treeItem, 'azureName')))._response.parsedBody;
+    return (result.some(async value => value.name === 'azureDatabasesForVSCode-publicIp' && value.startIpAddress === await getPublicIp()));
```
ejizba, line 47: I don't think we need to check the name, right? This logic should work regardless of where the firewall rule is set (in our extension or outside of it)

src/postgres/commands/createPostgresServer/steps/PostgresServerSetCredentialsStep.ts
```diff
@@ -18,7 +18,14 @@ export class PostgresServerSetCredentialsStep extends AzureWizardExecuteStep<IPo
 
     public async execute(context: IPostgresServerWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
 
-        const user: string = nonNullProp(context, 'longUserName');
+        let user: string;
+        // Username doesn't contain servername prefix for Postgres Flexible Servers only
+        // https://docs.microsoft.com/en-us/azure/postgresql/howto-configure-sign-in-aad-authentication
```
ejizba, line 23: I don't see anything in this docs about flexible server username stuff. Where exactly does it say that? Just trying to figure out if this is a bug or intended behavior

src/postgres/getClientConfig.ts
```diff
@@ -48,51 +50,6 @@ export async function getClientConfig(treeItem: PostgresServerTreeItem, database
         await client.connect();
         return clientConfig;
     } finally {
-        await client.end();
```
ejizba, line 51: I think the change to this line should be reverted. We still want to call `end`

src/postgres/commands/checkAuthentication.ts
neelip: That's true, the name should be irrelevant, especially if a user sets the rule on the portal.

src/postgres/commands/createPostgresServer/steps/PostgresServerSetCredentialsStep.ts
neelip: Sorry, that is the wrong link. 
Here: https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/what-is-flexible-server-in-azure-database-for-postgresql/ba-p/1741346 
`"...The direct connection also eliminates the need to include @servername suffix in your username on Flexible Server..."`

I couldn't find a better documentation than this one. But the change is visible on the portal as well.

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -54,7 +55,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         try {
-            const clientConfig: ClientConfig = await getClientConfig(this.parent, this.databaseName);
+            const clientConfig: ClientConfig = await checkAuthentication(context, this);
```
ejizba, line 58: This should still be `getClientConfig` so that the errors are shown in the tree (the whole point of the `catch` block), as opposed to warning notifications like in `checkAuthentication`

src/postgres/commands/checkAuthentication.ts
```diff
@@ -24,10 +29,22 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
...
             }
         }
     }
     return clientConfig;
 }
+
+export async function isFirewallRuleSet(treeItem: PostgresServerTreeItem): Promise<boolean> {
+    const serverType: PostgresServerType = nonNullProp(treeItem, 'serverType');
+    const client = createAbstractPostgresClient(serverType, treeItem.root);
+    client.firewallRules.listByServer
```
ejizba, line 45: delete this line

src/postgres/commands/checkAuthentication.ts
```diff
@@ -24,10 +29,22 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
...
+    const result: FirewallRuleListResult = (await client.firewallRules.listByServer(nonNullProp(treeItem, 'resourceGroup'), nonNullProp(treeItem, 'azureName')))._response.parsedBody;
```
ejizba, line 46: how come this has to do `_response.parsedBody`?

src/postgres/commands/checkAuthentication.ts
```diff
@@ -24,10 +29,22 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
...
+    const publicIp: string = await getPublicIp();
+    return (result.some(async value => value.startIpAddress <= publicIp && publicIp <= value.endIpAddress));
```
ejizba, line 48: The ip addresses are not numbers, so I don't think we can compare like this. Just checking equality is probably good enough. The only other option is to parse and compare each subsection, but I'm not sure that extra logic is worth it.

src/postgres/commands/checkAuthentication.ts
neelip: `listByServer()` returns `Models.FirewallRulesListByServerResponse` which is in this format. 

image

src/postgres/commands/checkAuthentication.ts
neelip: Oops sorry

src/postgres/commands/checkAuthentication.ts
neelip: That's true!

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -77,7 +82,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
                 });
                 credentialsTreeItem.commandArgs = [this.parent];
                 return [credentialsTreeItem];
-            } else if (this.parent.azureName && parsedError.errorType === firewallNotConfiguredErrorType) {
+            } else if (this.parent.azureName && (parsedError.errorType === firewallNotConfiguredErrorType || 'ETIMEDOUT') && !(await this.isFirewallRuleSet(this.parent))) {
```
ejizba, line 85: I think you still need this logic in `checkAuthentication`. It likely handles the same error for slightly different scenarios

src/postgres/commands/checkAuthentication.ts
ejizba: Look at the type again - the screenshot is cutoff on the right, but I'm guessing it's `FirewallRuleListResult & {`, meaning the object itself is the same type as `_response.parsedBody`

src/postgres/commands/checkAuthentication.ts
```diff
@@ -22,7 +22,7 @@ export async function checkAuthentication(context: IActionContext, treeItem: Pos
                 await enterPostgresCredentials(context, treeItem.parent);
 
                 // Need to configure firewall only for Azure Subscritption accounts
-            } else if (treeItem.parent.resourceGroup && parsedError.errorType === firewallNotConfiguredErrorType) {
+            } else if (treeItem.parent.resourceGroup && (parsedError.errorType === firewallNotConfiguredErrorType || 'ETIMEDOUT') && !(await treeItem.isFirewallRuleSet(treeItem.parent))) {
```
ejizba, line 25:
```suggestion
            } else if (treeItem.parent.resourceGroup && (parsedError.errorType === firewallNotConfiguredErrorType || parsedError.errorType === 'ETIMEDOUT') && !(await treeItem.isFirewallRuleSet(treeItem.parent))) {
```

Same issue in `PostgresDatabaseTreeItem`. Would be better if the logic was shared so we don't duplicate problems like this

neelip: In this case, it might actually better to combine 'ETIMEDOUT' condition and 'isFirewallRuleSet()', as 'isFirewallRuleSet()' not needed if a 'firewallNotConfiguredErrorType' is thrown. 

I am working on cleaning this up and maybe get rid of 'ETIMEDOUT' check completely. I will get that out as separate PR.

src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -95,4 +100,12 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         const config = await getClientConfig(this.parent, postgresDefaultDatabase);
         await runPostgresQuery(config, `Drop Database ${wrapArgInQuotes(this.databaseName)};`);
     }
+
+    public async isFirewallRuleSet(treeItem: PostgresServerTreeItem): Promise<boolean> {
```
ejizba, line 104: This should probably have a comment in the code explaining why we're doing it. Something like:
> Flexible servers throw a generic 'ETIMEDOUT' error instead of the firewall-specific error, so we have to check the firewall rules

ejizba (approved):

PR #1898, opened 2021-07-21
Add telemetry for Postgres serverType

neelip:


src/tree/SubscriptionTreeItem.ts
```diff
@@ -149,7 +149,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             }
         }
     }
-    private async initPostgresChild(server: PostgresAbstractServer): Promise<AzureTreeItem> {
+    private async initPostgresChild(server: PostgresAbstractServer, context: IActionContext): Promise<AzureTreeItem> {
```
nturinski, line 152: I don't think you want to get telemetry for every time the children tree items load.  If you're trying to see how many people are creating these server types, I think this seems like where you'd want it:

https://github.com/microsoft/vscode-cosmosdb/blob/main/src/postgres/commands/createPostgresServer/steps/PostgresServerCreateStep.ts#L51

src/tree/SubscriptionTreeItem.ts
```diff
@@ -46,7 +46,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         const treeItemPostgres: AzExtTreeItem[] = await this.createTreeItemsWithErrorHandling(
             postgresServers,
             'invalidPostgreSQLAccount',
-            async (server: PostgresAbstractServer) => await this.initPostgresChild(server),
+            async (server: PostgresAbstractServer) => await this.initPostgresChild(server, context),
```
ejizba, line 49: The problem with doing it in SubscriptionTreeItem is that each child will overwrite the previous child's value. I would suggest you add this telemetry in PostgresServerTreeItem.loadMoreChildrenImpl instead. If you want to do it here, maybe add two separate properties `hasFlexibleServer` and `hasSingleServer` or something like that

src/tree/SubscriptionTreeItem.ts
ejizba: I asked Neelima to add this and I think it's helpful for various reasons. For example, the code you linked doesn't cover people who created outside of VS Code. Secondly, knowing the serverType here will help us investigate any errors - we've already seen unique errors for flexible vs. single (https://github.com/microsoft/vscode-cosmosdb/pull/1890)

ejizba (approved):

PR #1899, opened 2021-07-21
Shorten Postgres Flexible Servers connection timeout

neelip:
The recently merged PR #1890 catches the `ETIMEDOUT` error but it takes close to 1 min for "client.connect()" before timing out. 

This PR completely prevents `ETIMEDOUT` error for loadChildren() in PostgresDatabaseTreeItem by checking for firewall rules on the portal (for Flexible Servers only) and configuring a firewall rule if there is none. 

`ETIMEDOUT` error might still happen with "checkAuthentication()" especially if "executePostgresQuery()" or "copyConnectionString()" is called before loading PostgresDatabaseTreeItem children. This case is not as common in my opinion, so I think we can wait until next release to improve this.

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -98,7 +98,7 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         const serverType: PostgresServerType = nonNullProp(treeItem, 'serverType');
         const client = createAbstractPostgresClient(serverType, treeItem.root);
         const result: FirewallRuleListResult = (await client.firewallRules.listByServer(nonNullProp(treeItem, 'resourceGroup'), nonNullProp(treeItem, 'azureName')))._response.parsedBody;
-        const publicIp: string = await getPublicIp();
+        const publicIp: string = await getPublicIp('Getting Public IP to verify server connectivity...');
         return (result.some(async value => value.startIpAddress === publicIp));
     }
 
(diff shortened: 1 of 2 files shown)


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -72,24 +77,11 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
             return children;
         } catch (error) {
             const parsedError: IParsedError = parseError(error);
-
             if (this.parent.azureName && parsedError.errorType === invalidCredentialsErrorType) {
                 void context.ui.showWarningMessage(localize('couldNotConnect', 'Could not connect to "{0}": {1}', this.parent.label, parsedError.message), { stepName: 'loadPostgresDatabases' });
-                const credentialsTreeItem: AzExtTreeItem = new GenericTreeItem(this, {
```
ejizba, line 78: This feels like too many changes to squeeze in right before a release, after we've already done testing. Flexible server is still in preview so the experience doesn't need to be perfect, especially since this could actually be fixed on the platform side (if they threw the firewall error instead of ETIMEDOUT).

IMO our top priority should be to avoid a regression for single server. While your fix might be okay, front-loading the logic does risk causing perf issues.

neelip: I agree with you on this, checking for firewall rule with every load doesn't seem efficient. 
I was thinking if we highlight that Flexible Servers are added in Preview in the ChangeLog and stick with the previous fix (#1890), that should be enough for this release.

neelip:
Closing this as it was a work around just for last release.

PR #1907, opened 2021-07-21
Update "Connected to Database" label for Postgres Func/SProc

### LINKED ISSUE #1905: Add more info to "Database" name in a function file to better identify
Currently, this is how the description looks like...
image

As all Postgres Servers have a "postgres" database, I think this might get confusing to identify which server the database is from. 
My suggestion: "servername.databasename"

@fiveisprime is this a good way to display the name?

neelip:
Fixes #1905 

Went with design in pgAdmin -->
image

### DIFF
--- a/src/postgres/commands/connectPostgresDatabase.ts
+++ b/src/postgres/commands/connectPostgresDatabase.ts
@@ -25,7 +25,7 @@ export async function connectPostgresDatabase(context: IActionContext, treeItem?
     const serverId: string = treeItem.parent.id;
     let serverName: string | undefined = serverId;
     if (username) {
-        serverName = username.includes(serverId) ? username : username + `@{serverId}`;
+        serverName = username.includes(serverId) ? username : username + `@${serverId}`;
     }
     const database = treeItem && treeItem.label;
 
--- a/src/postgres/commands/connectPostgresDatabase.ts
+++ b/src/postgres/commands/connectPostgresDatabase.ts
@@ -21,9 +21,16 @@ export async function connectPostgresDatabase(context: IActionContext, treeItem?
     const oldTreeItemId: string | undefined = ext.connectedPostgresDB && ext.connectedPostgresDB.fullId;
     void ext.context.globalState.update(connectedPostgresKey, treeItem.fullId);
     ext.connectedPostgresDB = treeItem;
+    const username: string | undefined = (await treeItem.parent.getFullConnectionString()).username;
+    const serverId: string = treeItem.parent.id;
+    let serverName: string | undefined = serverId;
+    if (username) {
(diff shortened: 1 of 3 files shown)


src/postgres/commands/connectPostgresDatabase.ts
```diff
@@ -21,9 +21,16 @@ export async function connectPostgresDatabase(context: IActionContext, treeItem?
     const oldTreeItemId: string | undefined = ext.connectedPostgresDB && ext.connectedPostgresDB.fullId;
     void ext.context.globalState.update(connectedPostgresKey, treeItem.fullId);
     ext.connectedPostgresDB = treeItem;
+    const username: string | undefined = (await treeItem.parent.getFullConnectionString()).username;
```
nturinski, line 24: I think that this `partialConnectionString` should already exist so you don't have to make any async calls to retrieve the username.

```suggestion
    const username: string | undefined = treeItem.parent.partialConnectionString.username;
```

src/postgres/commands/connectPostgresDatabase.ts
```diff
@@ -21,9 +21,16 @@ export async function connectPostgresDatabase(context: IActionContext, treeItem?
...
+    const serverId: string = treeItem.parent.id;
+    let serverName: string | undefined = serverId;
+    if (username) {
+        serverName = username.includes(serverId) ? username : username + `@${serverId}`;
```
nturinski, line 28: So, I don't know about your Postgres server, but my `serverId` includes my entire subscription name and id so it ends up looking like this:

image

We definitely don't wanna be including that in their database name 😅

nturinski (changes requested):
I don't really have any strong preference on what the string looks like, though I think my vote would be just to include the `serverId`

src/postgres/commands/connectPostgresDatabase.ts
neelip: With the way that it is setup right now, when we load Postgres Servers the partialConnectionString is not automatically updated with the cached credentials. getFullConnectionString() is where credential values of partialConnectionString are updated.

src/postgres/commands/connectPostgresDatabase.ts
neelip: oh right, good point! Attached accounts behaves differently so I assumed serverId, but I think label() value of the Server might be more appropriate.

src/postgres/commands/connectPostgresDatabase.ts
nturinski: In that case, I recommend checking this property first before calling the async method.  We want to avoid any async calls unless they're absolutely necessary.

nturinski:
Closing due to inactivity

PR #1908, opened 2021-07-22
RC 0.18.0

neelip:


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## 0.18.0 - 2021-07-21
+### Added
+- Postgres Flexible Servers (Preview): CRUD support for Databases, Functions and Stored Procedures
```
ejizba, line 5: I wouldn't use an acronym like `CRUD` - not everyone remembers what those stand for

NOTICE.html
```diff
@@ -1373,31 +1373,6 @@ <h1>NOTICES AND INFORMATION</h1>
 
 3. Neither the name of Salesforce.com nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
```
ejizba, line 1374: We can't change the NOTICE.html in the same PR that we change package-lock.json because their content is directly linked. I would recommend you do `npm audit fix` as it's own PR, merge that to main, wait for the main build to pass, and then re-generate this file

ejizba (approved):

PR #1915, opened 2021-07-29
Correction to last release: ETIMEDOUT issue

### LINKED ISSUE #1913: There is an error when signing the PostgreSQL Flexible server on another virtual machine
**OS:** Win10
**Build Version:** 20210725.1

**Repro steps:**
1. Create a PostgreSQL Flexible server -> Successfully added firewall rule for IP to it.
2. Expand the same PostgreSQL Flexible server on another virtual machine -> Enter the username and password.
3. Check whether the "Configure firewall to connect..." action shows under the PostgreSQL Flexible server.

**Expect:**
The "Configure firewall to connect..." action shows under the PostgreSQL Flexible server.

**Actual:**
There is an error under the PostgreSQL Flexible server.
image

**More Info:**
This issue doesn't reproduce for the PostgreSQL Single server.

neelip:
Fixes #1913 

After testing, found out that the issue is not specific to Windows, it occurs with all OS types.

### DIFF
--- a/src/postgres/tree/PostgresDatabaseTreeItem.ts
+++ b/src/postgres/tree/PostgresDatabaseTreeItem.ts
@@ -107,11 +107,6 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         const client = createAbstractPostgresClient(serverType, treeItem.root);
         const result: FirewallRuleListResult = (await client.firewallRules.listByServer(nonNullProp(treeItem, 'resourceGroup'), nonNullProp(treeItem, 'azureName')))._response.parsedBody;
         const publicIp: string = await getPublicIp();
-        for (let i = 0; i < result.length; i++) {
-            if (result[0].startIpAddress === publicIp) {
-                return true;
-            }
-        }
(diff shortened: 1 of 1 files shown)


src/postgres/tree/PostgresDatabaseTreeItem.ts
```diff
@@ -107,6 +107,11 @@ export class PostgresDatabaseTreeItem extends AzureParentTreeItem<ISubscriptionC
         const client = createAbstractPostgresClient(serverType, treeItem.root);
         const result: FirewallRuleListResult = (await client.firewallRules.listByServer(nonNullProp(treeItem, 'resourceGroup'), nonNullProp(treeItem, 'azureName')))._response.parsedBody;
         const publicIp: string = await getPublicIp();
-        return (result.some(async value => value.startIpAddress === publicIp));
```
neelip, line 110: This check is not working correctly. This was working in some cases where the result is empty and it is resolving directly to false. `async value => value.startIpAddress === publicIp` is not resolving to checking each array object.

ejizba (approved):

PR #1916, opened 2021-07-30
RC-0.18.1

neelip:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 0.18.2 - 2021-07-30
```
wwlorey, line 3: This should be `0.18.1` - we just remove the `-alpha` part from the version for a patch release

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+### Fixed
+- [Fix for Timing Out issue when loading Postgres Flexible Servers](https://github.com/microsoft/vscode-cosmosdb/issues/1913)
```
wwlorey, line 5: Instead of making this whole sentence a link, I'd recommend doing something like [this](https://github.com/microsoft/vscode-azurestorage/blob/main/CHANGELOG.md#fixed-7) and link just the issue at the end:

```suggestion
- Fix time out issue when loading Postgres Flexible Servers [#1913](https://github.com/microsoft/vscode-cosmosdb/issues/1913)
```

nturinski: https://github.com/microsoft/vscode-azurefunctions/blob/main/CHANGELOG.md#0191---2019-10-15

For reference, this is how Eric usually does it

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 0.18.1 - 2021-08-11
+### [Fixed](https://github.com/microsoft/vscode-cosmosdb/milestone/32?closed=1)
+- Fixed the ETIMEDOUT issue "...when signing the PostgreSQL Flexible server on another virtual machine"
```
wwlorey, line 5: I don't think you have to quote the issue title exactly - something like this would be good:

```suggestion
- Fixed the ETIMEDOUT issue when signing into a PostgreSQL Flexible server on multiple machines
```

wwlorey (approved):

PR #1920, opened 2021-08-10
Make Postgres Firewall Rule label unique

### LINKED ISSUE #1918: Accessing same Single/Flexible Postgres servers from two different machines resets Firewall Rule
image

We currently use the same label every time a firewall rule is added to a server from a machine. The problem occurs when two machines configure a firewall rule, one would replace the other.

neelip:
Fixes #1918

### DIFF
--- a/src/postgres/commands/configurePostgresFirewall.ts
+++ b/src/postgres/commands/configurePostgresFirewall.ts
@@ -39,8 +39,7 @@ export async function setFirewallRule(context: IActionContext, treeItem: Postgre
     const resourceGroup: string = nonNullProp(treeItem, 'resourceGroup');
     const serverName: string = nonNullProp(treeItem, 'azureName');
 
-    // firewall rule name on the portal is restricted to 80 characters or less
-    const firewallRuleName: string = "azDbVSCode-Ip" + `-${randomUtils.getPseudononymousStringHash(ip, 'hex')}`;
+    const firewallRuleName: string = "azDbVSCode-Ip" + `-${randomUtils.getPseudononymousStringHash(ip, 'hex').substring(0, 10)}`;
 
     const newFirewallRule: AbstractFirewallRule = {
         startIpAddress: ip,
--- a/src/postgres/commands/configurePostgresFirewall.ts
+++ b/src/postgres/commands/configurePostgresFirewall.ts
@@ -39,7 +39,7 @@ export async function setFirewallRule(context: IActionContext, treeItem: Postgre
     const resourceGroup: string = nonNullProp(treeItem, 'resourceGroup');
     const serverName: string = nonNullProp(treeItem, 'azureName');
 
-    const firewallRuleName: string = "azDbVSCode-Ip" + `-${randomUtils.getPseudononymousStringHash(ip, 'hex').substring(0, 10)}`;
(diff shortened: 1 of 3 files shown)


src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -38,7 +38,8 @@ export async function setFirewallRule(context: IActionContext, treeItem: Postgre
     const resourceGroup: string = nonNullProp(treeItem, 'resourceGroup');
     const serverName: string = nonNullProp(treeItem, 'azureName');
 
-    const firewallRuleName: string = "azureDatabasesForVSCode-publicIp";
+    const hashCode = (s: string) => s.split('').reduce((a, b) => (((a << 5) - a) + b.charCodeAt(0)) | 0, 0);
+    const firewallRuleName: string = "azureDatabasesForVSCode-publicIp" + hashCode(ip).toString();
```
ejizba, line 42: What if you just use this:
https://github.com/microsoft/vscode-cosmosdb/blob/main/src/utils/randomUtils.ts#L9

This could be another alternative: https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/utils/randomUtils.ts#L9

Both options are used pretty regularly throughout our extensions, so I prefer them over a new hash code implementation.

src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -38,7 +39,8 @@ export async function setFirewallRule(context: IActionContext, treeItem: Postgre
     const resourceGroup: string = nonNullProp(treeItem, 'resourceGroup');
     const serverName: string = nonNullProp(treeItem, 'azureName');
 
-    const firewallRuleName: string = "azureDatabasesForVSCode-publicIp";
+    // firewall rule name on the portal is restricted to 80 characters or less
+    const firewallRuleName: string = "azDbVSCode-Ip" + `-${randomUtils.getPseudononymousStringHash(ip, 'hex')}`;
```
ejizba, line 43: How long is the result of `getPseudononymousStringHash`? We normally just add 6-10 random characters to make names unique, otherwise the names are pretty unreadable. That's what I've seen the portal doing, too.

neelip: It is way longer than 10, more like 60 or something. I agree it should be smaller, what I am worried about is duplicity. Right now, the publicIP is hashed to hex. Maybe once hashed I can just use part of the hash for the label. I am guessing this would still make it unique.

ejizba: I would probably just do `getRandomHexString(6)`. If you're worried about that creating duplicates, I think there's only a 1 in 16 million chance of that happening (16^6 = 16,777,216). Plus the names only have to be unique within a certain subscription, right? So virtually impossible for dupes.

I'm not sure how hashing the IP affects the probabilities, so that's the only reason I would lean towards avoiding `getPseudononymousStringHash`, but honestly I think it's also virtually impossible for duplicates.

neelip: That's true, I haven't thought about that. Using just a small string should work.

ejizba (approved):

PR #1921, opened 2021-08-12
Switch to shared ESLint config

wwlorey:


.eslintrc.js
```diff
@@ -1,38 +1,3 @@
...
-        ],
-        "@typescript-eslint/no-inferrable-types": "off",
-        "@typescript-eslint/unbound-method": "off",
-        "@typescript-eslint/no-unnecessary-type-assertion": "off",
-        "@typescript-eslint/restrict-template-expressions": "off",
-        "@typescript-eslint/no-namespace": "off",
-        "@typescript-eslint/ban-types": "off",
-        "@typescript-eslint/require-await": "off"
-    }
+    "extends": "@microsoft/eslint-config-azuretools"
```
alexweininger, line 2: Since it looks like all the `import/no-internal-modules` issues are dealing with the `antlr4ts` package, I think we should override the rule like so: 

```suggestion
    "extends": "@microsoft/eslint-config-azuretools"
    "rules": {
        "import/no-internal-modules": [ "error", {
            "allow": [ "antlr4ts/**" ]
        } ]
    }
```

alexweininger (approved):

PR #1930, opened 2021-09-16
Use // @ts-nocheck to ignore errors in mongo/grammar legacy code

nturinski:
This is a npm package to migrate TSC without dealing with all of the errors, which seems to suit our use-case of ignoring the mongo/grammar legacy code.

Reference: https://dev.to/15five/how-to-temporarily-ignore-errors-during-a-typescript-migration-doe

The error looks like this:
```
Using TypeScript compiler version 4.4.3 from C:\Users\Nathan\Desktop\Code\vscode-cosmosdb\node_modules\typescript\lib\typescript.js
  {
    "codes": {
      "2611": 19
    },
    "pathRegExp": "/\\/mongo\\/grammar/",
    "total": 19
  }
]
Visible errors: 0, suppressed errors: 19
```


alexweininger:
We need to install [yargs](https://www.npmjs.com/package/yargs) as a devDependency since it's a peer dependency for tsc-silent.

https://github.com/evolution-gaming/tsc-silent/blob/master/package.json#L32-L34

nturinski:
I ended up just using // @ts-nocheck for the Mongo parser files rather than trying to convert everything.  For some reason, this broke the cleanReadme gulp script, so I had to install some dependencies for that as well as declare exports.

neelip (approved):

PR #1938, opened 2021-10-18
Updated "Create Server/Database Account" functions with "databases" package

neelip:
Both repos share few files from the "utils" folder. I have kept them as is on both repos. 
The only updates are for all the "Create Server/Database Account" related files: 
drawing

Related #[vscode-azuretools#1010](https://github.com/microsoft/vscode-azuretools/pull/1010)


JasonYeMSFT:
The moved files are already deleted from the utils package. Closing this issue since it's no longer feasible.
https://github.com/microsoft/vscode-azuretools/pull/1193

PR #1943, opened 2021-11-15
Remove skipLibCheck

alexweininger:
Since this is now closed I think we're safe to turn this off https://github.com/Azure/ms-rest-js/issues/367


bwateratmsft (approved):
Unskip ALL the LibChecks!

PR #1953, opened 2021-12-27
Replace Promise.all with Promise.allSettled when deleting test RGs

alexweininger:
Attempt to mitigate leftover resource groups from test runs.

[`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) - method returns a promise that resolves after all of the given promises have either fulfilled or rejected. Whereas `Promise.all` resolves after any of the given promises fulfill or reject. So now even if one of the promises rejects, it doesn't stop all of the other promises.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,8 @@
         "outDir": "out",
         "lib": [
             "es6",
-            "dom"
+            "dom",
+            "ES2020.Promise"
         ],
         "sourceMap": true,
         "noImplicitThis": true,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -5,7 +5,8 @@
         "outDir": "out",
         "lib": [
             "es6",
-            "dom"
+            "dom",
+            "ES2020.Promise"
```
alexweininger, line 9: I had to add this to be able to use the `Promise.allSettled` method. Do we have to add polyfills for this? How do we know what we can or can't use here?

alexweininger:
This isn't working how I'd hoped. Let me try some more things. Closing this for now.

Issue #1952, opened 2021-12-25
BUG `!` at end of password for psql throws error


hrittikhere:
Hey,
Currently, I am using Azure Database for PostgreSQL server and found an error while trying to connect to DB using psql. The error shows that having
`!` at the end of the password for a connection string (copied using the extension) would make psql think
that `!` is included in the host. This is problematic and the following picture would give you more information. Changing the password to not
include the `!` at the end would resolve the error or adding quotes for password in the connection string.

![](https://i.imgur.com/vw9W5VK_d.webp?maxwidth=640&shape=thumb&fidelity=medium)

I can try to add a PR for this. Let me know!

hrittikhere:
Solution is simple: Password in the string should be in quotes

alexweininger:
Thanks for filing this issue! Nice looking screenshot btw :)

> I can try to add a PR for this. Let me know!

If you do want to submit a PR, that'd be awesome. We encourage open source contributions! The relevant file for this would probably be https://github.com/microsoft/vscode-cosmosdb/blob/649873572001d4717d293facff28b14698831f9d/src/postgres/postgresConnectionStrings.ts

hrittikhere:
Yes, I found the file and I have started to work on it too. Will update you once the PR is ready.

PR #1954, opened 2021-12-28
Add Single Quotes for PSQL Connection String Password

hrittikhere:
Hey @alexweininger,
This fixes: https://github.com/microsoft/vscode-cosmosdb/issues/1952
Also, I have committed the TS file. Do we need to commit the `package-lock.json` too? Please let me know if that's the case as I am very very new to JS/TS. I tried the extension and it works perfectly with single quotes as expected. 

image

### DIFF
--- a/.azure-pipelines/common/sbom.yml
+++ b/.azure-pipelines/common/sbom.yml
@@ -1,0 +1,13 @@
+steps:
+  - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
+    displayName: "SBoM Generation Task"
+    inputs:
+      BuildDropPath: "$(Build.ArtifactStagingDirectory)"
+    condition: ne(variables['System.PullRequest.IsFork'], 'True')
+
+  - task: PublishBuildArtifacts@1
+    displayName: "Publish artifacts: sbom"
+    inputs:
+      PathtoPublish: "$(Build.ArtifactStagingDirectory)/_manifest"
+      ArtifactName: _manifest
+    condition: ne(variables['System.PullRequest.IsFork'], 'True')
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -20,6 +20,7 @@ jobs:
   steps:
   - template: common/build.yml
   - template: common/package.yml
+  - template: common/sbom.yml # only generate on linux
   - template: common/lint.yml
   - template: common/test.yml
 
--- a/src/postgres/postgresConnectionStrings.ts
+++ b/src/postgres/postgresConnectionStrings.ts
@@ -25,7 +25,7 @@ export function createPostgresConnectionString(hostName: string, port: string = 
         const encodedUsername = encodeURIComponent(username);
         if (password) {
             const encodedPassword = encodeURIComponent(password);
-            const encodedPasswordwithQuotes = "'"+encodedPassword+"'";
+            const encodedPasswordwithQuotes = "'" + encodedPassword + "'";
             connectionString += `${encodedUsername}:${encodedPasswordwithQuotes}@`;
         } else {
             connectionString += `${encodedUsername}@`;
--- a/src/postgres/postgresConnectionStrings.ts
+++ b/src/postgres/postgresConnectionStrings.ts
@@ -25,8 +25,8 @@ export function createPostgresConnectionString(hostName: string, port: string = 
         const encodedUsername = encodeURIComponent(username);
         if (password) {
             const encodedPassword = encodeURIComponent(password);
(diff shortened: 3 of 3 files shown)


alexweininger:
Thanks for submitting a PR! You don't need to commit the package-lock.json changes. I'll get to reviewing this properly soon.

hrittikhere:
@alexweininger sure! Looking forward to it :)

alexweininger:
What about doing this:

```ts
await vscode.env.clipboard.writeText(`"${connectionString}"`);
```

on this line: https://github.com/microsoft/vscode-cosmosdb/blob/328a2cb675cedb012869411cdd739587363a55fa/src/postgres/commands/copyConnectionString.ts#L29

We have some other logic relying on the connection strings and I think this is a safer change since it only alters the connection string when it gets copied. Also, surrounding the entire connection string in quotes will prevent any issues like this for the other connection string components. What do you think?

hrittikhere:
Hey @alexweininger
I experimented with that solution and the following solution keeps the whole connection string in the quotes and not just the password. It would definitely work but for the issue, I have created the main concern is sanitizing the password and this approach doesn't satisfy that requirement. 

image


I believe my PR won't be an issue as I didn't disturb `encodedPassword` but created a new variable just for passing the password to `copyConnectionString.ts`. If there's some better approach let me know I would be happy to learn and update.

alexweininger:
@hrittikhere Surrounding the whole connection string in double quotes works on macOS, does it not work on Windows?

Just for clarity, here's what I have on my terminal:
Screen Shot 2022-01-07 at 12 09 27 PM
Notice I have a `!` at the end of my password, the connection string is surrounded in double quotes, and the command executes properly. 

Let me know if this works for you.

I'm also unsure if I communicated my proposed solution properly, check and see my branch: https://github.com/microsoft/vscode-cosmosdb/tree/alex/quoteCopiedConnectionString. This solution should surround the entire connection string in double quotes.

---

> I believe my PR won't be an issue as I didn't disturb encodedPassword but created a new variable just for passing the password to copyConnectionString.ts. If there's some better approach let me know I would be happy to learn and update.

While I _think_ your changes wouldn't alter any behavior, the `createPostgresConnectionString` function is used in a few places other than just the copy connection string command. And because of this, it's safer to add the quotes to the connection string in the `copyConnectionString` function. This way we know it won't impact any other logic that calls the `createPostgresConnectionString` command.

References to `createPostgresConnectionString`:
Screen Shot 2022-01-07 at 12 09 27 PM

hrittikhere:
I am using a Linux machine and it doesn't work. @alexweininger can you try your solution on cloud shell once? I think it doesn't work on there too. 

I missed the `createPostgresConnectionString` function but I tried using your solution with `""` but I guess it's os specific. Please let me know about cloud shell. I will test it on a windows machine till then

hrittikhere:
It's not working for me. Please let me know the status of your tests if it's working on azure cloud shell.

alexweininger:
> It's not working for me. Please let me know the status of your tests if it's working on the azure cloud shell. Will it be a good approach if we create a new `createPostgresConnectionString` with a different variable name and then pass the password with quotes to our `copyConnectionString.ts` for copying purposes?

Yes I think that is a good approach. I did some quick testing on if the connecting string parsing package we are using handles the quotes nicely, and it doesn't. See below image:

image

I can confirm the connection string with a quoted password is parsed correctly on Azure Cloud Shell.

hrittikhere:
Okay, so then should I go with the approach I proposed? If you want to discuss over teams about any other approach I would be happy to,

alexweininger:
> Okay, so then should I go with the approach I proposed? 

If you're talking about this:

> Will it be a good approach if we create a new `createPostgresConnectionString` with a different variable name and then pass the password with quotes to our `copyConnectionString.ts` for copying purposes?

Then yes, I would create a new function `createCopyableConnectionString` that is then called in `copyConnectionString.ts`. Let me know if you want to discuss over Teams.

hrittikhere:
Sure! I am texting you at the mail provided on your GitHub account!

hrittikhere:
Hey @alexweininger 
Does it look good? I replicated the function and renamed it to `copyPostgresConnectionString`. Then passed it to our `copyConnectionString.ts`

I have tested it and it works as expected. Please update me how should I move ahead with this PR

alexweininger:
Revert changes to function

hrittikhere:
I have already done it. But Git shows it in a different way. I'm not sure how to tackle that but here's a block diagram that would help you to understand the two functions and the one with the required logic: 
image

hrittikhere:
Thanks @alexweininger for the help!

alexweininger (approved):
Thanks for submitting the PR, sorry some of my feedback took a while. We really appreciate it! 🚀 🔥 ❤️

PR #1980, opened 2022-04-21
Nat/unified

nturinski:


alexweininger (approved):
You don't have to do it in this PR, but we added the api.d.ts typings to the utils package https://github.com/microsoft/vscode-azuretools/blob/main/utils/hostapi.d.ts

Some stuff was renamed, the old names are marked as deprecated.

nturinski:
Got it!

Also left to do: contribute a local provider for the attached database accounts/emulator.  I'm not sure whether or not "attached" (that is, a remote database but attached via connection string) should go into the Azure or Workspace view though.

src/resolver/ResolvedDatabaseAccountResource.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { AzExtTreeItem, IActionContext, ICreateChildImplContext, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
+import { AppResource, ResolvedAppResourceBase } from "../api";
+import { DocDBAccountTreeItemBase } from "../docdb/tree/DocDBAccountTreeItemBase";
+import { MongoAccountTreeItem } from "../mongo/tree/MongoAccountTreeItem";
+import { PostgresServerTreeItem } from "../postgres/tree/PostgresServerTreeItem";
+
+export class ResolvedDatabaseAccountResource implements ResolvedAppResourceBase {
+    public id: string;
+    public contextValuesToAdd?: string[] | undefined = [];
```
bwateratmsft, line 14: If we're initializing here then it's guaranteed to have a value; we don't need the `?` / `| undefined`, and the `?`s later on.

bwateratmsft (approved):

package.json
```diff
@@ -90,15 +90,27 @@
...
         "onLanguage:mongo",
-        "onView:azureDatabasesExplorer"
+        "onView:azureResourceGroups"
     ],
     "contributes": {
-        "views": {
-            "azure": [
+        "x-azResources": {
+            "activation": {
+                "onFetch": [
```
bwateratmsft, line 98: Just `onResolve` is sufficient unless the activation needs a head start.

nturinski:
Will do Alex's feedback in another PR

PR #1993, opened 2022-04-29
Fixes opening file from Workspace view

nturinski:
Originally, all of the tree items were in one view, so we only had one FileSystemProvider.  In order to get the data, it would traverse the tree and find the appropriate node, and then display that in the vs code editor.  If you use today's build, you can't open any of the attached accounts documents.

Now that there are two separate views, each with their own tree, we need a FileSystemProvider for both.  However, it's tricky to know what node they clicked.  Basically, I'm looking at what the treeProvider on the node that they clicked is to figure out which provider to serve them.

`azureDatabases.update` was a weird one where it only passes in the uri.  My understanding of this command is that it should only happen on Azure resources, so I'm just assuming `appResourceFileSystem`.  It also only has 190 events so I'm not super concerned about it because I don't even know how to test it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.18.2-alpha.3",
+    "version": "0.18.2-alpha.4",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-cosmosdb",
-            "version": "0.18.2-alpha.3",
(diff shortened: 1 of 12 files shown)


nturinski:
Closing in favor of https://github.com/microsoft/vscode-cosmosdb/pull/1995

nturinski:
Closing in favor of https://github.com/microsoft/vscode-cosmosdb/pull/1995

PR #1998, opened 2022-05-05
Replace showTreeItemPicker with pickAppResource

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1146 and https://github.com/microsoft/vscode-azureresourcegroups/pull/221

### DIFF
--- a/src/postgres/commands/configurePostgresFirewall.ts
+++ b/src/postgres/commands/configurePostgresFirewall.ts
@@ -6,6 +6,7 @@
 import { DialogResponses, IActionContext } from "@microsoft/vscode-azext-utils";
 import * as publicIp from 'public-ip';
 import * as vscode from 'vscode';
+import { postgresFlexibleFilter, postgresSingleFilter } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { localize } from "../../utils/localize";
 import { nonNullProp } from '../../utils/nonNull';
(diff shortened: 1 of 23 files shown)


src/postgres/commands/configurePostgresFirewall.ts
```diff
@@ -16,7 +17,9 @@ import { PostgresServerTreeItem } from "../tree/PostgresServerTreeItem";
 
 export async function configurePostgresFirewall(context: IActionContext, treeItem?: PostgresServerTreeItem): Promise<void> {
     if (!treeItem) {
-        treeItem = <PostgresServerTreeItem>await ext.rgApi.appResourceTree.showTreeItemPicker(PostgresServerTreeItem.contextValue, context);
+        treeItem = await ext.rgApi.pickAppResource<PostgresServerTreeItem>(context, {
+            filter: [postgresSingleFilter, postgresFlexibleFilter]
```
bwateratmsft, line 21: Just double checking again, `PostgresServerTreeItem.contextValue` is intentionally dropped?

bwateratmsft: Also this same pattern appears in a couple of the other commands.

bwateratmsft (approved):

src/postgres/commands/configurePostgresFirewall.ts
nturinski: Yeah, the way this works is the filter will give you the `AppResource` that lives on `SubscriptionTreeItem`.  The `expectedContextValue` is only used for searching for children on that `AppResource` so if it's omitted, then it's assumed that the picker ends at the `AppResource`

alexweininger: Yes, the context value is only used for selecting children of top level app resources tree items (AppResource).

PR #2004, opened 2022-05-16
Fix pickTreeItem API command

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2229
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2226

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.18.2-alpha.8",
+    "version": "0.18.2-alpha.9",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1034,7 +1034,7 @@
         "@azure/arm-postgresql-flexible": "^5.0.0",
         "@azure/cosmos": "^3.6.3",
         "@microsoft/vscode-azext-azureutils": "^0.2.0",
```
bwateratmsft, line 1036: Should update this to 0.3.0 at the same time.

bwateratmsft (approved):
Keep in mind #2003; we need to make sure these changes eventually make it back to `main`.

PR #2008, opened 2022-05-16
Hardcode experience tags

### LINKED ISSUE #2005: Fail to create a new database account
**OS:** Win10
**Build Version:** 20220514.1

**Repro Steps:**

1. F1 -> Azure Databases: Create Server... -> Select "Core (SQL)" -> Enter a name -> Select a capacity model -> Create a new resource group -> Select a location.
2. Check whether succeeds to create a new database account.

**Expect:**
Succeed to create a new database account.

**Actual:**
Fail to create a new database account.
image

**More Info:**
1. This issue also reproduces for MongoDB, Table and Graph accounts.
2. This issue doesn't reproduce for PostgreSQL Single and Flexible servers.

nturinski:
Fixes #2005 

The way the compiler is loading the constants is throwing things off here.  Having constants import and use constants can cause weirdness depending on when it's loaded.

In this case, we were seeing the tags were `undefined` because the experiences were being created before all of the tags were.  We could probably move things around and fix this, but it seems a bit too high-risk of breaking other things that import these constants.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.18.2-alpha.9",
+    "version": "0.18.2-alpha.10",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-cosmosdb",
-            "version": "0.18.2-alpha.9",
+            "version": "0.18.2-alpha.10",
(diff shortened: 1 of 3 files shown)


alexweininger (dismissed):
Don't forget to bump the version

alexweininger (approved):

Issue #2013, opened 2022-05-17
The created database server is not focused after clicking "Click to view resource" in "ACTIVITYLOG"


v-xinda:
**OS:** Win10
**Azure Database Build Version:** 20220516.20
**Resource Goup Build Version:** 20220516.8

**Repro Steps:**
1. Create a database server.
2. Click "Click to view resource" in "ACTIVITYLOG".
3. Check whether the created database server is focused.

**Expect:**
The created database server is focused.

**Actual:**
The created database server is not focused.

**More Info:**
The created database server will be focused after changing the way resource groups are grouped.

alexweininger:
I believe this is fixed in the latest RGs build: 20220517.3

alexweininger:
This isn't fixed in 20220517.6

alexweininger:
Verified that this bug doesn't reproduce for creating storage accounts.

v-xinda:
Verified on the latest Database build 20220517.18 and Resource Group build 20220518.13, this issue doesn't reproduce.

PR #2018, opened 2022-05-17
Refresh tree after creating a server

alexweininger:
Fixes https://github.com/microsoft/vscode-cosmosdb/issues/2013


nturinski (approved):
Did you try if `refreshUIOnly` works as well?  I imagine it might be better for performance.

alexweininger:
> Did you try if `refreshUIOnly` works as well? I imagine it might be better for performance.

Tested it out, and it just doesn't work.

PR #2024, opened 2022-05-20
Hack into the unknown

nturinski:
Fixes a bug that was similar to https://github.com/microsoft/vscode-azurestorage/issues/1044

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.18.2-alpha.17",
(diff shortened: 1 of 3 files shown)


alexweininger:
Does this fix https://github.com/microsoft/vscode-azureresourcegroups/issues/270?

bwateratmsft:
> Does this fix [microsoft/vscode-azureresourcegroups#270](https://github.com/microsoft/vscode-azureresourcegroups/issues/270)?

Yep, that's the goal. We considered a handful of things but this seemed to be the least-worst way to fix it before //Build.

PR #2058, opened 2022-07-28
Remove Preview label for Postgres Flexible

tonybaloney:
Postgres Flexible Server is no longer in Preview.

https://docs.microsoft.com/en-us/azure/postgresql/flexible-server/quickstart-create-server-portal


nturinski (approved):
Thanks for this!

PR #2081, opened 2022-11-16
Skip test suite to enable CodeQL build

nturinski:
I commented out most of the testing code for Cosmos because it's very specific to Cosmos where as a lot of the other extensions had shared/duplicated code.

Had to upgrade the test/dev packages to get the Mac builds working as well.

Based off of this PR: https://github.com/microsoft/vscode-azurefunctions/pull/3275/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-cosmosdb",
     "version": "0.19.3-alpha.0",
-    "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
+    "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
     "publisher": "ms-azuretools",
     "displayName": "Azure Databases",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "vscode-cosmosdb",
     "version": "0.19.3-alpha.0",
-    "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
```
bwateratmsft, line 4: This needs to remain at the value currently in the code.

bwateratmsft (approved):

bwateratmsft (changes requested):
AI key shouldn't be changed

package.json
nturinski: Got it. 

Just so I know, why did we change it for Functions? I thought it had to do with upgrading the telemetry package.

bwateratmsft (approved):

PR #2138, opened 2023-06-20
Remove usage of keytar

### LINKED ISSUE #2137: Keytar Migration Plan
Hello :wave: I'm Tyler from the VS Code team. I'm opening this issue because you use the [keytar](https://github.com/atom/node-keytar) shim that VS Code provides.

Keytar is [officially archived](https://github.com/atom/node-keytar) and not being maintained... In an effort to promote good security practices by _not_ depending on an archived piece of software for storing secrets, we are working on a plan to remove this shim from VS Code. We want to communicate this to you so we minimize the disruption due to this change, hence this issue. I know this isn't the best news... we had a fair amount of work to remove our dependency on keytar as well, but we believe this is the right thing to do to ensure extensions are using secure APIs.

There are a couple of options for you to consider:
* _(recommended)_ Use the [SecretStorage API](https://code.visualstudio.com/api/references/vscode-api#SecretStorage) that VS Code provides on the [ExtensionContext](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext). This API is cross-platform and will work on all platforms that VS Code supports. It is also maintained by the VS Code team and will continue to be maintained and has been a part of the VS Code API for years at this point.
* _(not recommended)_ You can bundle the [keytar](https://github.com/atom/node-keytar) module with your extension. Keep in mind that keytar is a native node module which means that you will need to [publish a platform specific extension](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) for each platform you want to support.

### Timeline

The current plan is to remove this shim from VS Code Insiders in **early July** which means that **August 2023's** stable release will be the first release without the shim.

### Questions?

If you have any questions, please feel free to ask them here. I will do my best to answer them as quickly as possible. Your cooperation is greatly appreciated :heart:

bwateratmsft:
Fixes #2137

Unresolved question--how do we want to deal with the data "disappearing"?


src/postgres/tree/PostgresServerTreeItem.ts
```diff
@@ -181,31 +182,30 @@ export class PostgresServerTreeItem extends AzExtParentTreeItem {
...
             const storedValue: string | undefined = ext.context.globalState.get(PostgresServerTreeItem.serviceName);
-            if (storedValue && ext.keytar) {
+            if (storedValue) {
                 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
                 const servers: IPersistedServer[] = JSON.parse(storedValue);
                 for (const server of servers) {
                     if (server.id === this.id) {
                         this.partialConnectionString.username = server.username;
-                        this.partialConnectionString.password = await ext.keytar.getPassword(PostgresServerTreeItem.serviceName, this.id) || undefined;
+                        // TODO: keytar: migration plan?
```
bwateratmsft, line 207: TODO: do

src/tree/AttachedAccountsTreeItem.ts
```diff
@@ -310,7 +306,8 @@ export class AttachedAccountsTreeItem extends AzExtParentTreeItem {
                     isEmulator = (<IPersistedAccount>account).isEmulator;
                     label = isEmulator ? `${getExperienceFromApi(api).shortName} Emulator` : `${id} (${getExperienceFromApi(api).shortName})`;
                 }
-                const connectionString: string = nonNullValue(await keytar.getPassword(this._serviceName, id), 'connectionString');
+                // TODO: keytar: migration plan?
```
bwateratmsft, line 309: TODO: do

bwateratmsft: This is what the user is presented with:
image

There are no attach actions on this tree item, though. Should add them.

alexweininger: > There are no attach actions on this tree item, though. The user is forced to go through the command palette. Should add them.

If you merge with main, then you can use the new actions submenu. #2136

alexweininger:
image

image

Had to enter in credentials before doing any operations on a database within a postgres account. Worked as expected

alexweininger (approved):

PR #2158, opened 2023-08-23
Re-enable tests

JasonYeMSFT:


alexweininger (approved):

nturinski (commented):
I think that we may still want to skip the long running tests. I'm pretty sure that we haven't converted the Azure account credentials to be compatible.

test/global.test.ts
```diff
@@ -11,7 +11,6 @@ export const longRunningTestsEnabled: boolean = !/^(false|0)?$/i.test(process.en
 
 // Runs before all tests
 suiteSetup(async function (this: Mocha.Context): Promise<void> {
-    this.skip();
```
nturinski, line 14: I think that we may still want to skip the long running tests. I'm pretty sure that we haven't converted the Azure account credentials to be compatible.

JasonYeMSFT: I explicitly set the long running tests flag to false.

alexweininger (approved):

PR #2178, opened 2023-09-19
Show inline action when using password

JasonYeMSFT:
resolves https://github.com/microsoft/vscode-cosmosdb/issues/2177


package.json
```diff
@@ -434,6 +434,12 @@
                 "category": "PostgreSQL",
                 "command": "postgreSQL.openStoredProcedure",
                 "title": "Open Stored Procedure"
+            },
+            {
```
JasonYeMSFT, line 438: This is what it looks like
image

alexweininger (approved):

PR #2202, opened 2023-10-12
Extend the policheck exclusion list to include dist

JasonYeMSFT:


alexweininger (commented):
Just add `DIST` I don't think those other entries are needed.

JasonYeMSFT:
> Just add `DIST` I don't think those other entries are needed.

Done.

alexweininger (approved):

PR #2213, opened 2023-11-20
Enable OAuth for CosmosDB resources

JasonYeMSFT:
Enable OAuth for CosmosDB resources using Azure CLI for logging in. To use OAuth to authenticate
- Download Azure CLI
- Login via Azure CLI
- Add the path to Azure CLI install location to the setting.
- Restart the extension

The master key will still be used if the user is able to list it. Those who want to only give OAuth access to the resources will need to configure the permission to prevent listing master keys.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
                 "@azure/arm-cosmosdb": "^15.0.0",
                 "@azure/arm-postgresql": "^6.0.0",
                 "@azure/arm-postgresql-flexible": "^5.0.0",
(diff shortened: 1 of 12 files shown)


mudanzasIPROM (approved):

mudanzasIPROM (approved):
ok

JasonYeMSFT:
This PR was created when I don't have access to the main repository. Now I does and I have stopped working on my fork. I'll abandon this one and create a separate PR in the main repo to solve this problem.

PR #2214, opened 2023-11-21
Upgrade vscode-azext-dev to resolve critical alerts

JasonYeMSFT:


JasonYeMSFT:
If you know there is a potential breaking change that I need to take care of, please comment.

nturinski (approved):
I think the main change was the webpack stuff (which is going to be reverted anyway)

PR #2215, opened 2023-11-21
Improve warning message on firewall error

JasonYeMSFT:
The original warning message is not 100% correct since Azure Portal doesn't guarantee that it can add the actual IP address to the firewall rules. The string is a slightly modified version of the Portal string indicating what to do for the next step.

### DIFF
--- a/src/postgres/tree/ClientConfigFactory.ts
+++ b/src/postgres/tree/ClientConfigFactory.ts
@@ -68,8 +68,17 @@ export class PostgresClientConfigFactory {
                 } else if (parsedError.errorType === firewallNotConfiguredErrorType || parsedError.errorType === timeoutErrorType) {
                     // The time out error are common when the firewall rules doesn't grant access from the current IP address.
                     // If the client is blocked by the firewall, let the user go to Azure Portal to grant access.
+                    const publicIp = PostgresServerTreeItem.ipAddr;
+                    let ipMessage: string;
+                    if (publicIp !== undefined) {
+                        ipMessage = localize("ipAlreadyInFirewall", "The IP address '{0}' already exists in the firewall rules.", publicIp);
+                    } else {
+                        // The code should never reach here but handle it just in case.
+                        ipMessage = "Your IP address is already in the firewall rules.";
+                    }
+                    const configureFirewallMessage = localize("mustConfigureFirewall", "Some network environments may not report the actual public-facing IP address needed to access your server. Contact your network administrator to add the actual IP address to the firewall rules.");
                     throw {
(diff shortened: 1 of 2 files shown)


src/postgres/tree/ClientConfigFactory.ts
```diff
@@ -69,7 +69,7 @@ export class PostgresClientConfigFactory {
                     // The time out error are common when the firewall rules doesn't grant access from the current IP address.
                     // If the client is blocked by the firewall, let the user go to Azure Portal to grant access.
                     throw {
-                        message: localize("mustConfigureFirewall", 'Must configure firewall from Azure Portal to grant access.'),
+                        message: localize("mustConfigureFirewall", 'Your IP address we tested is already in the firewall rules. Some network environments may not report the actual public-facing IP address needed to access your server. Contact your network administrator to add the actual IP address to the firewall rules.'),
```
nturinski, line 72: Do we have the IP address that was tested? That may be helpful information to display to the user.

Maybe something like:

"The IP address ${0} already exists in the firewall rules. Some network environments may not report the actual public-facing IP address required to access your server. Contact your network administrator to add the actual IP address to the firewall rules."

src/postgres/tree/ClientConfigFactory.ts
```diff
@@ -68,8 +68,17 @@ export class PostgresClientConfigFactory {
                 } else if (parsedError.errorType === firewallNotConfiguredErrorType || parsedError.errorType === timeoutErrorType) {
                     // The time out error are common when the firewall rules doesn't grant access from the current IP address.
                     // If the client is blocked by the firewall, let the user go to Azure Portal to grant access.
+                    const publicIp = PostgresServerTreeItem.ipAddr;
+                    let ipMessage: string;
+                    if (publicIp !== undefined) {
+                        ipMessage = localize("ipAlreadyInFirewall", "The IP address '{0}' already exists in the firewall rules.", publicIp);
```
nturinski, line 74: nit: You could just do something like this:

`const ipMessage = PostgresServerTreeItem.ipAddr ? localize("ipAlreadyInFirewall", "The IP address '{0}' already exists in the firewall rules.", publicIp) : localize('ipAlreadyInFirewall2', "Your IP address is already in the firewall rules.");

src/postgres/tree/ClientConfigFactory.ts
```diff
@@ -68,8 +68,17 @@ export class PostgresClientConfigFactory {
...
+                    } else {
+                        // The code should never reach here but handle it just in case.
+                        ipMessage = "Your IP address is already in the firewall rules.";
```
nturinski, line 77: nit: localize

nturinski (approved):

src/postgres/tree/ClientConfigFactory.ts
JasonYeMSFT: Technically speaking this string should never be displayed as indicated by the comment. I am not going to worry about having someone translate it.

Issue #2209, opened 2023-11-17
An error occurs after adding firewall rule to the PostgreSQL server


v-xinda:
**OS:** Win10
**Build Version:** 20231116.4
**Regression:** Previous release (0.20.0)

**Repro Steps:**
1. Create a PostgreSQL server -> Configure firewall to connect to account.
2. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
![image](https://github.com/microsoft/vscode-cosmosdb/assets/57662131/149249c7-7f85-406e-a03c-2d90781d0990)

v-xinda:
Verified this issue on the build 20240112.17, it doesn't reproduce for now. So I close it.

PR #2229, opened 2024-01-12
Fix auto getting azure user id

JasonYeMSFT:
fixes https://github.com/microsoft/vscode-cosmosdb/issues/2209


src/postgres/tree/ClientConfigFactory.ts
```diff
@@ -22,14 +23,20 @@ export class PostgresClientConfigFactory {
         clientConfig: ClientConfig
     }> {
         const parsedConnectionString = await treeItem.getFullConnectionString();
-        const azureUserSession = await getAzureAdUserSession();
```
JasonYeMSFT, line 25: This no longer works for most of the users because the authentication has been moved out of Account extension.

alexweininger (approved):

alexweininger (approved):

PR #2230, opened 2024-01-12
Bump version to 0.20.1

JasonYeMSFT:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## 0.20.1
+
+### Changed
+
```
nturinski, line 6: nit Usually we don't have a linebreak between header titles and the first bullet point.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.20.0",
```
nturinski, line 3: In the future, we should add the alpha suffix after we release.

nturinski (approved):

package.json
JasonYeMSFT: We did, this PR shows 0.20.0 because it branches off the commit where 0.20.0 is released.

CHANGELOG.md
JasonYeMSFT: When I was working on Azure docs, the checking tool recommends adding a blank line before list of items. As far as I know, it's a more generally accepted markdown rule.

Issue #2066, opened 2022-10-03
unable to execute cosmos stored procedures inside of VS code


Orddu:
Does this occur consistently? Yes
Repro steps:

1. connect to azure account 
2. open a stored procedure inside of cosmos
3. attempt to run stored procedure, as available in the web app 

Version: 0.19.2
OS: win32
OS Release: 10.0.19044
Product: Visual Studio Code
Product Version: 1.71.2
Language: en

nturinski:
Hi @Orddu
What is the error that you're seeing when you try to run the stored procedure?

Issue #896, opened 2018-10-23
Add support to execute a stored procedure


PrashanthCorp:
in the node-documentdb package, the document client class has an endpoint `client.executeStoredProcedure`. We could investigate using that to enable the functionality.

tec-goblin:
Hi, I had this in my wish list :). We'd need a way to request the arguments, a way to show the results, and a way to show statistics (as is the case for graphClient - the main reason why I'd use it outside an application would be to see if it works and the AUs).
If we have all that, then we could extend it (as a different issue) to show similar stats and results from doing any arbitrary query to DocumentDb. That's very useful in order to detect the impact of various alternatives of querying or modelling the data.
Once you start having a plan, feel free to assign me some of the tasks if you want :) - as long as it's not terribly urgent and you can wait for a month.

PR #2235, opened 2024-01-17
feature: CosmosDB execute stored procedure

JasonYeMSFT:
There were work items for this:
https://github.com/microsoft/vscode-cosmosdb/issues/2066
https://github.com/microsoft/vscode-cosmosdb/issues/896

### DIFF
--- a/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
+++ b/src/docdb/tree/DocDBStoredProcedureTreeItem.ts
@@ -9,8 +9,11 @@ import * as vscode from "vscode";
 import { IEditableTreeItem } from '../../DatabasesFileSystem';
 import { ext } from '../../extensionVariables';
 import { nonNullProp } from '../../utils/nonNull';
+import * as vscodeUtil from "../../utils/vscodeUtils";
 import { DocDBStoredProceduresTreeItem } from './DocDBStoredProceduresTreeItem';
 import { IDocDBTreeRoot } from './IDocDBTreeRoot';
+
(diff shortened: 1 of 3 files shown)


JasonYeMSFT:
The UX will be:

1. Right click a stored procedure and click "Execute Stored Procedure"
2. Enter the partition key
3. Enter the parameters

src/docdb/tree/DocDBStoredProcedureTreeItem.ts
```diff
@@ -73,4 +74,17 @@ export class DocDBStoredProcedureTreeItem extends AzExtTreeItem implements IEdit
...
     }
+
+    public async execute(context: IActionContext, partitionKey: string, parameters?: any[]): Promise<void> {
+        const client = this.root.getCosmosClient();
+        const result = await this.parent.getContainerClient(client).scripts.storedProcedure(this.id).execute(partitionKey, parameters);
+
+        try {
+            const resultFileName = `${this.label}-result`;
+            const text = JSON.stringify(result, undefined, 2);
+            await vscodeUtil.showNewFile(text, resultFileName, ".json");
```
alexweininger, line 85: I think it'd be better to use our `openReadOnlyJson` function here instead. Since it's readonly it doesn't show as unsaved like `showNewFile` does.

JasonYeMSFT: Done. And I used a randomly generated hex string for the unique document id since the result is supposed to be used as a temporarily.

alexweininger (approved):

PR #2242, opened 2024-02-15
chore: reorganize MongoDB command registration

JasonYeMSFT:
Reorganize code so that MongoDB command registration is separate from command implementations.

### DIFF
--- a/src/mongo/registerMongoCommands.ts
+++ b/src/mongo/registerMongoCommands.ts
@@ -3,25 +3,29 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzExtTreeItem, callWithTelemetryAndErrorHandling, IActionContext, IErrorHandlerContext, ITreeItemPickerContext, registerCommandWithTreeNodeUnwrapping, registerErrorHandler, registerEvent } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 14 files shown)


src/mongo/registerMongoCommands.ts
```diff
@@ -33,95 +37,68 @@ export function registerMongoCommands(): void {
...
-        context.telemetry.properties.isActivationEvent = 'true';
-
-        try {
-            const persistedNodeId: string | undefined = ext.context.globalState.get(connectedMongoKey);
-            if (persistedNodeId) {
-                const persistedNode = await ext.rgApi.appResourceTree.findTreeItem(persistedNodeId, context);
-                if (persistedNode) {
-                    await ext.mongoLanguageClient.client.onReady();
-                    await vscode.commands.executeCommand('cosmosDB.connectMongoDB', persistedNode);
+export async function loadPersistedMongoDB(): Promise<void> {
```
nturinski, line 76: Why did you not move this one to the `loadPersistedMongoDB.ts` file?

JasonYeMSFT: Sorry, I forgot this. I will move it to a separate file as well.

nturinski: Yeah, looks like you already have one but it was just empty.

JasonYeMSFT: I decided to put it in the same file of connectMongoDBDatabase because loadPersistedMongoDB doesn't have its own command and is closely related to the connectMongoDBDatabase command.

I am going to keep setUpErrorReporting() and updateErrorsInScrapbook() in the registerMongoCommands.ts for now since they aren't used in any command.

nturinski (approved):

PR #2279, opened 2024-04-24
chore: Prepare 0.22.0 release

JasonYeMSFT:
related to: https://github.com/microsoft/azcode-internal/issues/125

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-cosmosdb",
-    "version": "0.21.1-alpha",
+    "version": "0.22.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1070,10 +1070,10 @@
                     "default": 8081,
                     "description": "Port to use when connecting to a CosmosDB Emulator instance"
                 },
-                "azureDatabases.testCosmosAuth": {
+                "azureDatabases.useCosmosOAuth": {
```
JasonYeMSFT, line 1073: My testing indicates this setting is still needed. The built-in Cosmos DB roles both allow people to list keys even if they cannot use the key due to disabling local auth. This setting allows people to prefer OAuth credential over the listed key.

alexweininger (approved):

PR #2311, opened 2024-09-10
Dev/sda/adds prettier updates eslint

bk201-:
This pull request includes significant updates to the linting configuration, code formatting, and various code improvements. The most important changes include updating the ESLint configuration, adding a Prettier configuration, modifying import statements for consistency, and updating dependencies in `package.json`.

### Linting and Formatting Configuration:
* [`.eslintrc.js`](diffhunk://#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L2-R42): Replaced the existing configuration with a more comprehensive setup using `@typescript-eslint` and additional rules for better code quality.
* [`.prettierrc`](diffhunk://#diff-663ade211b3a1552162de21c4031fcd16be99407aae5ceecbb491a2efc43d5d2R1-R12): Added a new Prettier configuration file to enforce consistent code formatting.
* [`.eslintignore`](diffhunk://#diff-a0fdacd2ade87c5238dde44378f574f7e123e669acdf8b740878b14b33b20275R5-R6): Updated to include `webpack.config.js` and `webpack.config.*.js` files.

### Dependency Updates:
* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L1103): Updated various dependencies, including `eslint`, `@typescript-eslint`, and added `prettier` and `prettier-plugin-organize-imports`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L1103) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L1113-R1128)

### Code Improvements:
These changes collectively enhance the code quality, maintainability, and consistency across the project.


sevoku (dismissed):

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        '@typescript-eslint/ban-types': 'off',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/no-inferrable-types': 'off',
+        '@typescript-eslint/no-namespace': 'off',
+        '@typescript-eslint/no-unnecessary-type-assertion': 'off',
+        '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+        '@typescript-eslint/prefer-regexp-exec': 'off',
+        '@typescript-eslint/require-await': 'off',
+        '@typescript-eslint/restrict-template-expressions': 'off',
+        '@typescript-eslint/unbound-method': 'off',
```
tnaum-ms, line 27: I don't speak from experience here so I'm open to your feedback. My suggestion to turn this back on it's based on the research of this topic I did:

This rule ensures that class methods are always bound to the correct this context, avoiding runtime errors.

Example:
```jsx
class MyClass {
    name = 'Test';
    printName() {
        console.log(this.name);
    }
}

const myClass = new MyClass();
const printName = myClass.printName; // Unbound method
printName(); // Error: 'this' is undefined or refers to the wrong context
```

Solution with correct binding:
```jsx
const printName = myClass.printName.bind(myClass); // Properly bound method
printName(); // Outputs: "Test"
```

By turning this rule **on**, we'd ensure that other code contributors are prompted to bind methods properly, preventing hard-to-debug issues with `this` context.

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        eqeqeq: ['error', 'always'],
+        'import/consistent-type-specifier-style': ['error', 'prefer-inline'],
+        'import/no-internal-modules': [
+            'error',
+            {
+                allow: ['antlr4ts/**', 'yaml/types'],
+            },
+        ],
+        'no-case-declarations': 'off',
```
tnaum-ms, line 36: Let's keep it **on**, so that we enforce proper scoping when variables are defined.

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        'no-constant-condition': ['error', { checkLoops: false }],
```
tnaum-ms, line 37: I'm not strong about this one. I use `while(true)` often ;-)
It's more about helping external contributors to making conscious decisions.

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        'no-inner-declarations': 'off',
```
tnaum-ms, line 38: This one is risky. Are we disabling it due to legacy-code?

```jsx
if (condition) {
    function sayHello() { // Declaration inside a block
        console.log('Hello');
    }
}

sayHello(); // Error: 'sayHello' is not defined (if `condition` is false)
```

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        'no-restricted-imports': ['error', { patterns: ['**/*/extension.bundle'] }],
+        'no-unused-vars': 'off',
+        'no-useless-escape': 'off',
```
tnaum-ms, line 41: I'd personally prefer review future code without useless escapes ;-)
If that's for legacy code, let's just add an ignore around that legacy code but keep the guards up?

tnaum-ms (changes requested):
I'd like to discuss some of the changes in `.eslintrc.js` and better understand the reasoning behind them. I've commented on these individually.

I assume that some changes are due to existing legacy code. Perhaps we can add exceptions at the file level for those cases but maintain stricter rules moving forward.

Please note that I did not look closely at the `src` and `test` directories, as I expected only formatting changes due to the Prettier config update.

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        sourceType: 'module',
+    },
+    plugins: ['@typescript-eslint', 'import'],
+    extends: [
+        'eslint:recommended',
+        'plugin:@typescript-eslint/recommended',
+        'plugin:@typescript-eslint/recommended-requiring-type-checking',
+    ],
+    rules: {
+        '@typescript-eslint/ban-types': 'off',
```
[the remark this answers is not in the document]
tnaum-ms: `        '@typescript-eslint/ban-types': 'off',`

.eslintrc.js
tnaum-ms: `        '@typescript-eslint/unbound-method': 'off',`

.eslintrc.js
tnaum-ms: `        'no-case-declarations': 'off',`

.eslintrc.js
tnaum-ms: `        'no-constant-condition': ['error', { checkLoops: false }],`

.eslintrc.js
tnaum-ms: `        'no-inner-declarations': 'off',`

.eslintrc.js
tnaum-ms: `        'no-useless-escape': 'off',`

.eslintrc.js
bk201-: It should be `eslint no-constant-condition: "error"` without any exceptions. In 99.99% everyone can avoid endless loops

.eslintrc.js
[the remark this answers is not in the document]
bk201-: This rule anyway doesn't work, it was removed from 8 version. I've changed it to correct one

.eslintrc.js
bk201-: Switched to 'error'

.eslintrc.js
bk201-: Fixed

.eslintrc.js
bk201-: Fixed

.eslintrc.js
```diff
@@ -1,8 +1,43 @@
...
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/no-inferrable-types': 'off',
+        '@typescript-eslint/no-namespace': 'off',
+        '@typescript-eslint/no-unnecessary-type-assertion': 'off',
+        '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+        '@typescript-eslint/prefer-regexp-exec': 'off',
+        '@typescript-eslint/require-await': 'off',
```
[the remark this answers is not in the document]
bk201-: Disagree. It is legal way to create function witch returns Promise. I know that in this case it is possible to use `Promise.resolve()`. Therefore I switched this rule to `warning` in this case you will have ability at least detect this and fix if you want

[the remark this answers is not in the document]
sevoku: I agree, it's always easy to miss and debug a missing await and it's legit to disable the warning in single lines if it's intended.

.eslintrc.js
sevoku: Maybe we should define https://eslint.org/docs/latest/rules/no-fallthrough as well then? Although... I'm really not a fan of  brackets in switch/case in general...

sevoku (approved):

.eslintrc.js
tnaum-ms: Thanks for sharing, I wasn't aware of the fact that bind returning an `any`.

tnaum-ms (approved):
We've discussed the changes and comments and have resolved the open questions.

PR #2319, opened 2024-09-17
mongoClusters: Document Schema 

tnaum-ms:


tnaum-ms:
I started using `denque` for a more performant schema scanner, which required me to modify our `tsconfig.js` and enable:

```js
"esModuleInterop": true
```

This is my only concern; the rest of the code has no side effects outside of the `mongoClusters` implementation.

PR #2329, opened 2024-10-02
Add ErrorBoundary connected to telemetry

languy:
Add an `<ErrorBoundary>` component to catch any errors during rendering. For now, we do this for `QueryTab`.

It is inserted deep enough in the component tree to access the dispatcher that talks to the webview host. The limitation is that if rendering crashes right away, the dispatcher may not have had time to initialize and it won't report to telemetry. But for now, it's probably better than replicating the dispatcher functionality.

`reportError` can be used to report any error to telemetry. If it's a rendering error in the webview, the stack is swapped with the `componentStack` which is more helpful.

image

### DIFF
--- a/src/Telemetry.ts
+++ b/src/Telemetry.ts
@@ -27,7 +27,7 @@ export class TelemetryContext {
      * @returns
      */
     public reportWebviewError = (message: string, stack: string, componentStack: string | undefined): Promise<void> =>
-        callWithTelemetryAndErrorHandling<void>('cosmosdb.common.query-editor.webview-error', (actionContext) => {
+        callWithTelemetryAndErrorHandling<void>('cosmosdb.common.query-tab.webview-error', (actionContext) => {
             actionContext.errorHandling.suppressDisplay = true;
             actionContext.valuesToMask = Array.from(this.valuesToMask);
 
--- a/src/Telemetry.ts
+++ b/src/Telemetry.ts
@@ -20,18 +20,14 @@ export class TelemetryContext {
     }
 
     public reportError = (message: string, stack: string, componentStack: string | undefined): Promise<void> =>
-        callWithTelemetryAndErrorHandling<void>('nosql.querytaberror', (actionContext) => {
+        callWithTelemetryAndErrorHandling<void>('queryTab.webviewError', (actionContext) => {
             actionContext.errorHandling.suppressDisplay = true;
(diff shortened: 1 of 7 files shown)


src/webviews/utils/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+    {
+        textAlign: 'left',
+    }
+});
+
+const ErrorDisplay: React.FC<{ message: string | undefined, details: string | null | undefined }> = ({ message, details }) => {
+    const styles = useStyles();
+    return <div className={styles.container}>
+        <h1>An unexpected error occurred</h1>
+        <p>Please try again. If the error persists, please <Link href="https://github.com/microsoft/vscode-cosmosdb/issues/new/choose">report the issue</Link></p>
```
sevoku, line 34: Since at this point we have probably logged the error already, we should be able to reuse the report a problem command which would create the GH issue and add the error details with other metadata to it.

src/panels/QueryEditorTab.ts
```diff
@@ -211,6 +221,12 @@ export class QueryEditorTab {
                 return this.prevPage(payload.params[0] as string);
             case 'firstPage':
                 return this.firstPage(payload.params[0] as string);
+            case 'reportError':
+                return QueryEditorTab.telemetryContext.reportError(
+                    payload.params[0] as string,
```
sevoku, line 226: minor: let's add a comment to each with a reminder what is what?

sevoku (approved):

src/webviews/utils/ErrorBoundary.tsx
languy: Wired the "azureDatabases.reportIssue" command.

src/Telemetry.ts
```diff
@@ -0,0 +1,36 @@
...
+            if (connection.masterKey) {
+                this.addMaskedValue(connection.masterKey);
+            }
+            this.addMaskedValue(connection.databaseId);
+            this.addMaskedValue(connection.containerId);
+        }
+    }
+
+    public reportError = (message: string, stack: string, componentStack: string | undefined): Promise<void> =>
+        callWithTelemetryAndErrorHandling<void>('queryTab.webviewError', (actionContext) => {
```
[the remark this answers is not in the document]
languy: Per latest discussions, replaced with: "cosmosdb.common.query-editor.webview-error"

PR #2336, opened 2024-10-16
feat: Implementing BaseTab as parent class for QueryEditorTab and DocumentTab

bk201-:


src/panels/DocumentTabV2.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    protected initController() {
+        super.initController();
+
+        this.channel.on<void>('ready', async () => {
+            await this.channel.postMessage({
+                type: 'event',
+                name: 'initState',
+                params: [this.mode, this.connection.databaseId, this.connection.containerId, this.documentId],
```
languy, line 111: This goes to the webview, right? As a future optimization, we could send this directly in the template.

languy (approved):

PR #2342, opened 2024-10-22
vCore: adding table view data navigation into objects

tnaum-ms:


tnaum-ms (commented):
reviewed, verified file contents, restored correct package.json version

PR #2344, opened 2024-10-22
vCore: add import + export support for vCore collections (EJSON)

tnaum-ms:


tnaum-ms:
@tnaum-ms Missing correct EJSON import, add correct handling for vCore execution path. Add a ticket for MongoRU imports.

PR #2351, opened 2024-10-24
Fix an issue where an internal error is not meaningful

sevoku:
There are several scenarios where loading database resources might fail. This change adds an optional message to a commonly used nonNullValue check function and improves database resource resolution error handling. Additioally the resource ID and Name gets now masked when reporting errors to telemetry and GitHub issues.

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -163,8 +163,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     ): Promise<AzExtTreeItem> {
         const experience = tryGetExperience(databaseAccount);
         const id: string = nonNullProp(databaseAccount, 'id');
-        const name: string = nonNullProp(databaseAccount, 'name');
-        const documentEndpoint: string = nonNullProp(databaseAccount, 'documentEndpoint');
+        const name: string = nonNullProp(databaseAccount, 'name', `of the database account ${databaseAccount.id}`);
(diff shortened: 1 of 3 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -163,8 +163,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     ): Promise<AzExtTreeItem> {
         const experience = tryGetExperience(databaseAccount);
         const id: string = nonNullProp(databaseAccount, 'id');
-        const name: string = nonNullProp(databaseAccount, 'name');
-        const documentEndpoint: string = nonNullProp(databaseAccount, 'documentEndpoint');
+        const name: string = nonNullProp(databaseAccount, 'name', `of the database account ${databaseAccount.id}`);
+        const documentEndpoint: string = nonNullProp(databaseAccount, 'documentEndpoint', `of the database account ${databaseAccount.id}`);
```
bk201-, line 167: I met the error in this place today. every time when I open DB it shows in the console and doesn't catch properly. Is it an critical issue? Should we fix the cause or not? I think it might be an issue for users since I see the VSCode notification that something went wrong internally

sevoku: I get this error for the two sevoku-test accounts which failed to provision and lack the endpoint. Without the endpoint we can not connect. I'm wondering which other accounts are affected.

PR #2355, opened 2024-10-28
vCore: add an extended confirmation dialog for "risky" operations

tnaum-ms:
**Currently it's used in vCore only, but it also adds a new "Setting":**

image

**A confirmation prompt when option is DISABLED:**

image

**A confirmation prompt when option is ENABLED:**

image

------------

BTW: I just noticed that Azure Resources has a similar setting in place, however, they'd ask the user to input the name of the resource being deleted.

image


sevoku:
Couple of thoughts:

1. what is the default setting for `azureResourceGroups.deleteConfirmation`? Seems to be `Enter Name`? I'm wondering if it'd make more sense to reuse that to have a centralised setting, why would users want to have different preference for databases?
2. Data Explorer is always asking for the name: 
image
3. If we stick with the existing setting, we could still do the number puzzle for `Click Button` preference depending on the severity of the operation.
4. `Enter Name` is not applicable for bulk operations such as bulk deleting documents, so for these operations we could always show the number puzzle.
5. In your example the "default" button seems to be the correct number:
image
So that hitting "Enter" would be enough to confirm with the keyboard. We should make sure the "required" number is not the first/default keyboard selection.

tnaum-ms:
Thank you @sevoku ! I wasn't aware of `azureResourceGroups.deleteConfirmation` existence. This is the best way to go.
Also, Dmitrii made good remarks over chat about testing and e2e scripts. I'll update this code to use the `deleteConfirmation` that's already implemented and tested.

sevoku:
as I wrote, I'm not advocating for the old experience, I like the puzzle especially for operations without an available name (we can't expect folks to enter document ID + partition key, and not what for bulk operations). I think it makes a lot of sense to have this "simplified" experience for the database context. However there are some more caveats:
1. The Resources setting is speaking of "resource groups":
image
That's not obvious that it also affects other operations, so we should PR a wording change to Azure Resources
2. The puzzle dialog should also have a link to the settings page, since it's mentioning that the behaviour can be changed in settings.

tnaum-ms:
Updated settings, the "quiz" is not in there, just the two basic ones as in "Azure Resources":

image

This is the behavior. I'm using "Drop Container" from vCore:

**Word Confirmation**
image

**Button Click**
image

tnaum-ms:
We have a simple helper function that takes care of reading settings and showing an appropriate warning:

```js
/**
 * Prompts the user for a confirmation based on the configured confirmation style.
 *
 * @param title - The title of the confirmation dialog.
 * @param message - The message to display in the confirmation dialog. This message will be suffixed with instructions for a specific prompt.
 * @param expectedConfirmationWord - The word that the user must type to confirm the action when the confirmation style is set to 'Word Confirmation'.
 * @returns A promise that resolves to a boolean indicating whether the user confirmed the action.
 */
export async function getConfirmationAsInSettings(
    title: string,
    message: string,
    expectedConfirmationWord: string,
): Promise<boolean> {
    const deleteConfirmation: string | undefined = vscode.workspace
        .getConfiguration()
        .get<string>(ext.settingsKeys.confirmationStyle);

    if (deleteConfirmation === 'Word Confirmation') {
        return await getConfirmationWithWordQuestion(title, message, expectedConfirmationWord);
    }
    return await getConfirmationWithClick(title, message);
}
```

package.json
```diff
@@ -1266,6 +1266,19 @@
                     "description": "Whether to use OAuth credential for Cosmos DB NoSQL resources",
                     "default": false
                 },
+                "azureDatabases.confirmationStyle": {
+                    "type": "string",
+                    "description": "The behavior to use when confirming operations that cannot be undone, such as deleting resources.",
+                    "enum": [
```
sevoku, line 1272: The enum values should be unique and internal for localisation purposes, such as `WordConfigmration` and `ButtonClick`, otherwise localisation will be difficult. There is an additional `enumItemLabels` property to define labels for the values, which can be localised afterwards. See the docs: https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema

package.json
```diff
@@ -1266,6 +1266,19 @@
...
+                        "Word Confirmation",
+                        "Buton Click"
```
sevoku, line 1274: Typo: Button

src/utils/confirmations.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function getConfirmationAsInSettings(
+    title: string,
+    message: string,
+    expectedConfirmationWord: string,
+): Promise<boolean> {
+    const deleteConfirmation: string | undefined = vscode.workspace
+        .getConfiguration()
+        .get<string>(ext.settingsKeys.confirmationStyle);
+
+    if (deleteConfirmation === 'Word Confirmation') {
```
sevoku, line 27: Let's define a const or an enum for this

src/utils/confirmations.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export async function getConfirmationWithClick(title: string, message: string): Promise<boolean> {
+    const confirmation = await vscode.window.showWarningMessage(
+        title,
+        {
+            modal: true,
+            detail: message,
+        },
+        'Yes',
```
sevoku, line 85: `Yes` will be the default button in this case, I'm still not convinced that a destructive operation should be the default dialog action. It makes it too easy to confirm. We should define our own `No` or `Cancel` button and make it the default choice, so that at least one additional "TAB" would be needed to confirm with the keyboard.

sevoku (commented):
Sorry for being late to the party :)

package.json
tnaum-ms: Thanks, I started using `enumItemLabels`

src/utils/confirmations.ts
tnaum-ms: 👍

src/utils/confirmations.ts
tnaum-ms: @sevoku that's a tricky one, when going for a modal dialog, we automatically get a "Cancel" button provided. The "Cancel" button doesn't get the default flag assigned.. So in the end, the only way to get there.. it's another QuickPick and not a dialog.

I'll push the other changes and will put this one on hold after the blocking ones are fixed.

tnaum-ms:
to be continued here: https://github.com/microsoft/vscode-cosmosdb/pull/2377

src/utils/confirmations.ts
sevoku: Oh I see now what's going on, since the cancel button has always `isCloseAffordance: true` it makes the other "Yes" button the default automatically. This can only be avoided by either adding more buttons or using a custom Cancel button with `isCloseAffordance: false`, however that would break ESC key behaviour. Probably not much we can do here then. However for the quiz button, consider making the "right" answer not the first button in the list.

PR #2366, opened 2024-10-31
feat: Index Metrics in Stats

bk201-:
image

image


languy (approved):

sevoku (changes requested):
PEr discussion: let's move it to the "Stats" tab as a last row.

sevoku (approved):

PR #2378, opened 2024-11-05
Fix an issue where Mongo nodes did not have a local parent assigned

sevoku:
This led to the following error:
```
Error: Internal error: 'branchDataProvider.getTreeItem' threw exception Internal error: Expected value to be neither null nor undefined: parent
```
image

Which is now fixed:
image


sevoku:
The same issue affected other databases as well, such as PG:
image

tnaum-ms (approved):
I reviewed the code, understand the changes, and have tried it out with my test instance. I confirm that I'm now able to see data in my Mongo RU system:

image

sevoku:
I rewrote the commit message a bit, forced push didn't have any code changes.

PR #2384, opened 2024-11-06
Fix an issue where a MongoDB connection error might be hidden

sevoku:
If the connection to a Mongo RU account fails, the actual error is not shown, instead there is another error:
image since `this._root.isEmulator` might be undefined.


tnaum-ms (approved):
Looks good. Plus I tried this with my test env that used to fail before this PR and can confirm that it resolved the issue.

PR #2385, opened 2024-11-06
Improve server names when creating a new resource

sevoku:
When creating a new database resource it currently looks like this:
image

This change unifies Cosmos DB naming to align better with branding and documentation:
image


tnaum-ms:
I'll merge this with my current work on MongoDB vCore and connection strings to make sure that everything is in sync.
I'll then come back to this PR.

tnaum-ms (approved):
Reviewed.

PR #2394, opened 2024-11-11
fix: Decrease toolbar paddings

bk201-:


sevoku (approved):
Yep, this looks better:
image
image

PR #2395, opened 2024-11-11
vCore: adds support for connecting to MongoDB Clusters with a connection string

tnaum-ms:
This PR introduces support for "MongoDB Cluster Accounts" in the Workspace area. It provides an option to add a MongoDB cluster connection string to the workspace, store it, and use it to interact with a MongoDB cluster account. This feature is not exclusive to Azure domains and allows interaction with any MongoDB cluster.

image

###  Workspace Tree View

It is built on the v2 tree view API provided by `azure-resources`. This PR introduces `SharedWorkspaceResourceProvider` and `SharedWorkspaceStorage` for future use, enabling MongoDB Clusters support through `MongoClustersWorkspaceBranchDataProvider` and `MongoDBAccountsWorkspaceItem`

###  Behavior When Connecting to MongoDB Clusters

A new node has been added to the workspace tree view, allowing users to add new connections. There is no limit to the number of accounts that can be added:

image

When a user attempts to add a connection:

image

The connection string is verified, and any errors are shown. Users are then prompted to confirm the username and re-enter the password. If the user is confident that the password in the connection string is correct, they can confirm by pressing Enter.

image

image

###  Behavior When Connecting to Azure Cosmos DB for MongoDB (RU)

The connection flow begins similarly to MongoDB Clusters. However, if an RU cluster is detected (by checking the domain name suffix), the following actions occur:

If an RU connection string is detected:

1. The user is informed:

image

2. The connection is added to the "Attached Database Account" section, with focus set to the newly added node:

image

###  Other Changes

Most changes in the  `mongoClusters`  folder support the display of MongoDB cluster tree items within the workspace area.


sevoku:
I currently get a connection timeout no matter whether I go though the new workspace or the Azure experience... after entering my password it fails.

sevoku:
Also using the "Azure Databses" 
image
Button works fine for NoSQL and MongoDB RU accounts, howerver it fails when I use Clusters Accounts > New Connection with an RU account
image
it asks for username and password which is not appiclable and fails at the end.

src/AzureDBExperiences.ts
```diff
@@ -101,7 +101,7 @@ export function getExperienceQuickPick(api: API): IAzureQuickPickItem<Experience
 
 export function getExperienceQuickPickForAttached(api: API): IAzureQuickPickItem<Experience> {
     const exp = getExperienceFromApi(api);
-    return { label: exp.shortName, description: exp.description, data: exp };
+    return { label: exp.longName, description: exp.description, data: exp };
```
sevoku, line 104: Why this change and where is it effective?

src/mongoClusters/wizards/addWorkspaceConnection/ConnectionStringStep.ts
```diff
@@ -0,0 +1,76 @@
...
+                placeHolder: localize(
+                    'mongoClusters.addWorkspaceConnection.connectionString.placeholder',
+                    'Starts with mongodb:// or mongodb+srv://',
+                ),
+                validateInput: ConnectionStringStep.validateInput,
+                asyncValidationTask: (connectionString: string) => this.validateConnectionString(connectionString),
+            })
+        ).trim();
+
+        context.valuesToMask.push(context.connectionString);
```
sevoku, line 32: maybe add connectionString.username and connectionString.password to the mask list as single strings as well, just to be on the safe side.

src/tree/workspace/sharedWorkspaceStorage.ts
```diff
@@ -0,0 +1,160 @@
...
+
+        // Check for existing item
+        const existingItem = ext.context.globalState.get<SharedWorkspaceStorageItem>(storageKey);
+        if (existingItem && !overwrite) {
+            throw new Error(`An item with id '${item.id}' already exists for workspaceType '${workspaceType}'.`);
+        }
+
+        // Save all secrets
+        if (item.secrets && item.secrets.length > 0) {
+            const secretKey = `${storageKey}/secrets`;
```
sevoku, line 109: Maybe let's use the existing [`getSecretStorageKey`](https://github.com/microsoft/vscode-cosmosdb/blob/main/src/utils/getSecretStorageKey.ts#L7) func? And probably reuse the [existing key for attached accounts](https://github.com/microsoft/vscode-cosmosdb/blob/main/src/tree/AttachedAccountsTreeItem.ts#L56) so all creds are stored in the same place.

src/AzureDBExperiences.ts
tnaum-ms: It's accessed when a user attempts to attach an account using the "Attached Database Accounts" path. It's used to display the options available. Previously, the `shortName` was shown.

image

src/tree/workspace/sharedWorkspaceStorage.ts
tnaum-ms: I looked into this while introducing the new `SharedWorkspaceStorage`.
Here, I reused the prefix `ms-azuretools.vscode-cosmosdb` and added `workspace` to the namespace:
```ts
    private static readonly storageName: string = 'ms-azuretools.vscode-cosmosdb.workspace';
```
This approach introduces a new namespace structure to support the storage of API-specific data and secrets beyond connection string, i.e. we might store additional keys/passwords, etc.. For example, I introduced the `secrets` branch instead of a `connectionString` branch.

Each workspace branch data provider is assigned its own namespace, nested under the common prefix.

The goal is to start with a unified but extensible structure. I’d like to migrate the existing "Attached Database Accounts" to this new structure as part of the V1 to V2 migration.

src/mongoClusters/wizards/addWorkspaceConnection/ConnectionStringStep.ts
tnaum-ms: good catch, added masking of these in their "steps"

tnaum-ms:
> But this is minor, we should probably add the same behaviour with a message and add the Cluster to the appropriate section. So basically the opposite behavior to adding RU to Clusters section.

Indeed. I'll look into it in a separate PR once that build is out.

tnaum-ms:
Thank you for your review and the hints. Merging.

PR #2406, opened 2024-11-13
feat: Add "Open Query Editor" item to the Documents node submenu

bk201-:


sevoku (changes requested):
@bk201- this adds the context menu item but the Query Editor does not pick up the context and doesn't connect when called this way:
image

sevoku (approved):

PR #2424, opened 2024-11-15
fix: Security improvements

bk201-:
- Dev server is disabled for production build
- CSP rules for production build use only VSCode uri, for development build dev-server host is added

### DIFF
--- a/src/webviews/index.tsx
+++ b/src/webviews/index.tsx
@@ -13,12 +13,7 @@ import { type WebviewState, WithWebviewContext } from './WebviewContext';
 
 export type ViewKey = keyof typeof WebviewRegistry;
 
-export function render<V extends ViewKey>(
-    key: V,
-    vscodeApi: WebviewApi<WebviewState>,
-    _publicPath: string,
-    rootId = 'root',
-): void {
+export function render<V extends ViewKey>(key: V, vscodeApi: WebviewApi<WebviewState>, rootId = 'root'): void {
     const container = document.getElementById(rootId);
     if (!container) {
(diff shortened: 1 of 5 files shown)


src/webviews/index.tsx
```diff
@@ -16,19 +16,14 @@ export type ViewKey = keyof typeof WebviewRegistry;
 export function render<V extends ViewKey>(
     key: V,
     vscodeApi: WebviewApi<WebviewState>,
-    publicPath: string,
+    _publicPath: string,
```
tnaum-ms, line 19: can we just drop it then? it's just us using it.

webpack.config.views.js
```diff
@@ -42,17 +42,7 @@ module.exports = (env, { mode }) => {
...
                             jsc: {
-                                minify: {
-                                    compress: !isDev,
-                                    mangle: isDev
-                                        ? false
-                                        : {
-                                              keep_classnames: true,
-                                              keep_fnames: true,
-                                          },
-                                },
```
tnaum-ms, line 55: are we indeed disabling minification, or is it happening by default anyway?

tnaum-ms (commented):
Thank you for changes for both webview paths.
I'm not sure what's the motivation behind removing `minify` config. Is it obsolete?

webpack.config.views.js
bk201-: It happens by default on another step. These flags don't make any sense here.

tnaum-ms: Thank you for clarifying! This helps a lot.

tnaum-ms (approved):

PR #2426, opened 2024-11-16
fix: remove unnecessary code since patch fix was applied in upstream repo

ghiscoding:
- this was fixed a while ago in upstream repos, which I maintain, Slickgrid-React v5.7 (Slickgrid-Universal) via these 2 PRs 
  - https://github.com/ghiscoding/slickgrid-universal/issues/1658
  - https://github.com/ghiscoding/slickgrid-universal/pull/1675
- this was discussed in this upstream [Discussion](https://github.com/ghiscoding/slickgrid-universal/discussions/1655) 

cc @tnaum-ms

Side note: I don't have any Azure DBs, so I can't really test this myself.

### DIFF
--- a/src/webviews/QueryEditor/ResultPanel/ResultTabViewTree.tsx
+++ b/src/webviews/QueryEditor/ResultPanel/ResultTabViewTree.tsx
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { useEffect } from 'react';
 import { FieldType, Formatters, SlickgridReact, type GridOption } from 'slickgrid-react';
 
 type ResultTabViewTreeProps = {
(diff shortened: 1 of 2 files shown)


ghiscoding:
@microsoft-github-policy-service agree

tnaum-ms:
@ghiscoding Thank you very much for your contribution here. I truly appreciate your help and for following up!

By the way, we are expanding the extension to support databases beyond Azure. You can now connect to any MongoDB cluster out there. This feature is available in the "Workspaces" section. Give it a try and see the grid in action :) Once you do, you'll get a sense of how often we'll be collaborating in the future, as there are many exciting features we haven't activated yet.

image

ghiscoding:
@tnaum-ms  That's great to hear, I will give it a try soon. The funny thing is that I started using a bit more [Azure Data Studio](https://learn.microsoft.com/en-us/azure-data-studio/) which is also using SlickGrid (the original one though, not mine) to show the data, however it doesn't have any Tree features like your project. 

Are you intending to provide more database connector in the future? Because at the moment, my DBs are more around MySQL, Postgres and SQL Server. So I'm not using much of DocumentDB or Mongo yet. 

Cheers and thanks for the SlickGrid use 😉

PR #2434, opened 2024-11-20
Add tsaoptions.json

alexweininger:
See https://github.com/microsoft/vscode-azuretools/pull/1823

### DIFF
--- a/.azure-pipelines/compliance/tsaoptions.json
+++ b/.azure-pipelines/compliance/tsaoptions.json
@@ -4,15 +4,15 @@
     "codeBaseName": "vscode-cosmosdb",
     "tsaStamp": "DevDiv",
     "notificationAliases": [
-        "AzCode@microsoft.com"
+        "cosmosdbdevtooling@microsoft.com"
     ],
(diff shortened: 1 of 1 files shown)


.azure-pipelines/compliance/tsaoptions.json
alexweininger, line 1: Likely need to change these values to the new owners. Made the PR to start the process else I'll forgot.

bk201- (dismissed):

.azure-pipelines/compliance/tsaoptions.json
sevoku: here: #2444

alexweininger: > @alexweininger I'd advocate for going ahead and updating this right away, but do we have a way to test it? We'd route to a different AzDO service. I'll create a PR with the changes.

I've tested it in one of our repos. Also, the failure case isn't destructive, the TSA task gives helpful error messages.

sevoku (approved):

PR #2447, opened 2024-11-20
Add missing multi-tenant support.

sevoku:
Select the right Tenant when authenticating SDK calls and pass it along to VSCode when retrieving the token from the session.

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -252,8 +252,9 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
                         }
                     }
 
+                    const tenantId = parent.subscription.tenantId ?? databaseAccount.identity?.tenantId;
                     // OAuth is always enabled for Cosmos DB and will be used as a fall back if key auth is unavailable
-                    const authCred = { type: 'auth' };
+                    const authCred = { type: 'auth', tenantId: tenantId };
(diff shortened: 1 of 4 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -252,8 +252,9 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
                         }
                     }
 
+                    const tenantId = parent.subscription.tenantId ?? databaseAccount.identity?.tenantId;
                     // OAuth is always enabled for Cosmos DB and will be used as a fall back if key auth is unavailable
-                    const authCred = { type: 'auth' };
+                    const authCred = { type: 'auth', tenantId: tenantId };
```
languy, line 257: nit: `const authCred = { type: 'auth', tenantId };` should work, too.

languy (approved):

src/tree/SubscriptionTreeItem.ts
```diff
@@ -252,8 +252,9 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
                         }
                     }
 
+                    const tenantId = parent.subscription.tenantId ?? databaseAccount.identity?.tenantId;
```
bk201-, line 255: Workspace TreeVIew V2 doesn't have subscription. How to get tenantId having only connection string?

PR #2448, opened 2024-11-20
fix: Default color for Run and Connect icons

bk201-:
- Default color always has high contrast ration with primary color

image
image

The contrast ratio of the "Run" icon is greater than 3:1 in query editor.

### DIFF
--- a/src/webviews/QueryEditor/QueryPanel/QueryToolbar.tsx
+++ b/src/webviews/QueryEditor/QueryPanel/QueryToolbar.tsx
@@ -31,19 +31,13 @@ import {
 import { useQueryEditorDispatcher, useQueryEditorState } from '../state/QueryEditorContext';
 
 const useClasses = makeStyles({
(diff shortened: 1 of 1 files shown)


src/webviews/QueryEditor/QueryPanel/QueryToolbar.tsx
```diff
@@ -31,19 +31,13 @@ import {
...
         color: tokens.colorStatusDangerBorderActive,
     },
-    iconConnect: {
-        color: tokens.colorStatusSuccessBorderActive,
-    },
     iconDisconnect: {
         color: tokens.colorStatusDangerBorderActive,
     },
-});
+}); // #005aa1  #1765BF
```
sevoku, line 40: huh?

bk201-: Forgot to remove, done

Issue #2449, opened 2024-11-20
NoSQL Emulator support on Linux


sevoku:
With the recent announcement for [the Emulator on Linux](https://devblogs.microsoft.com/cosmosdb/introducing-the-new-linux-based-azure-cosmos-db-emulator-preview/) we should look into supporting this. Here is the highly voted community feedback item for the emulator itself: https://feedback.azure.com/d365community/idea/277b3add-0d25-ec11-b6e6-000d3a4f0858, so I think this would be a great feature to add sooner than later.

PR #2450, opened 2024-11-20
Enable Attach Emulator on Linux and MacOS

sevoku:
This enables the Emulator support on Linux. Unfortunately I don't have a Linux setup right now, @mkrueger could you maybe help with this?

Note for investigation: this also enables it for Mongo, we should look into that as well since there is MongoDB for Linux, not sure how functional Emulator support is for that.

Addresses #2449

### DIFF
--- a/src/commands/attachEmulator/attachEmulator.ts
+++ b/src/commands/attachEmulator/attachEmulator.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizard, type IActionContext } from '@microsoft/vscode-azext-utils';
-import { isLinux, isWindows } from '../../constants';
+import { isEmulatorSupported } from '../../constants';
 import { type CosmosDBAttachEmulatorResourceItem } from '../../tree/attached/CosmosDBAttachEmulatorResourceItem';
 import { localize } from '../../utils/localize';
 import { type AttachEmulatorWizardContext } from './AttachEmulatorWizardContext';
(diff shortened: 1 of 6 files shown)


xgerman:
@sevoku our goal is http support so that you want need certificates - per default we run in http and are waiting for ther SDKs to roll out support. Once this is there it *should* all work out of the box and only if you really need HTTPS there would be work. /cc @christopheranderson

tnaum-ms:
> Note for investigation: this also enables it for Mongo, we should look into that as well since there is MongoDB for Linux, not sure how functional Emulator support is for that.

I'm currently working on updating MongoDB's emulator's integration on Linux and can confirm that its current release works on Linux.

mkrueger:
Can't create a container (I get an error about partition key kind).
Changing that fixes it for me:
```
--- a/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
+++ b/src/docdb/tree/DocDBDatabaseTreeItemBase.ts
@@ -4,6 +4,7 @@
 
 import {
+    PartitionKeyKind,
     type ContainerDefinition,
     type ContainerResponse,
     type CosmosClient,
@@ -101,6 +102,7 @@ export abstract class DocDBDatabaseTreeItemBase extends DocDBTreeItemBase<Contai
             containerDefinition.partitionKey = {
                 paths: [partitionKey],
             };
+            containerDefinition.partitionKey.kind = PartitionKeyKind.Hash;
         }
         const options: RequestOptions = {};
```

sevoku:
Thanks for checking @tnaum-ms and @mkrueger! I'll rebase this PR on top of the current V2 migration work, it seems @bk201- has changed it to always be set if I'm not mistaken: https://github.com/microsoft/vscode-cosmosdb/blob/dev/sda/tree-api-migration/src/commands/createContainer/DocumentDBExecuteStep.ts#L31

sevoku:
@mkrueger done, could you give it another try?

sevoku:
Yay it works for me on MacOS now as well! Fixed a small bug where auth didn't work right and enabled the Emulator on MacOS (Intel)

PR #2455, opened 2024-11-21
vCore: Updated username input to align with current API

tnaum-ms:
Up until now, during the preview phase, we used to read secondary admin users from the Control Plane. This feature is no longer available. Now, during authentication, we ask for the username, suggesting the current admin username as the default input.


src/mongoClusters/tree/MongoClusterResourceItem.ts
```diff
@@ -57,18 +55,11 @@ export class MongoClusterResourceItem extends MongoClusterItemBase {
...
                 const clusterConnectionString = nonNullValue(clusterInformation.connectionString);
 
-                context.valuesToMask.push(clusterConnectionString);
-
-                // Fetch non-admin users using the extracted method
-                const clusterNonAdminUsers = await this.fetchNonAdminUsersFromAzure(
-                    managementClient,
-                    clusterInformation,
-                );
+                context.valuesToMask.push(clusterConnectionString, clusterInformation.administratorLogin ?? '');
```
sevoku, line 58: perf nit: we should not push empty values, it might same some ticks here, but telemetry call will need to reiterate on that empty string. I know in this case it's a noop, but let's have a separate 
```
if (clusterInformation.administratorLogin) {
    context.valuesToMask.push(clusterInformation.administratorLogin);
}
```
just to keep awareness of performance since in other cases it might have a bigger impact.

sevoku (approved):

src/mongoClusters/tree/MongoClusterResourceItem.ts
tnaum-ms: you're right, I remember typing that quick `?? ''` and knowing that I do not like it at all.. I'll pay more attention to it. Thanks!

sevoku (approved):

PR #2464, opened 2024-11-22
Created new issue templates for bug reports and feature requests

tnaum-ms:
Not as wild as this one: https://github.com/microsoft/PowerToys/issues/new/choose
but, it's the first step :)

**Bug Report**

template1

**Feature Request**

image


sevoku:
This looks great, but looking at the Powertoys examle... maybe we could leverage some of the yaml functionality? Also we should check whether this still works with https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/reportAnIssue.ts and maybe even look into making it more flexible in aztools and allow prepopulating yaml forms or at least leverage custom templates.

tnaum-ms:
Thank you for pointing out https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/reportAnIssue.ts I'll look into it.

tnaum-ms:
I've take care of it before the end of July 2025.

tnaum-ms:
I'll close it for now. It's not a pressing issue.

PR #2467, opened 2024-11-25
feat: Deleted documents are highlighted

### LINKED ISSUE #2437: The selected document is not removed if deleting it from document query editor in "Cosmos DB for NoSQL" account
**OS:** Windows 10
**Build Version:** 20241120.1

**Repro Steps:**
1. Create a "Cosmos DB for NoSQL" account.
2. Create a database and a container for the above account.
3. Expand the above account -> Right click the "Documents" node -> Create several documents -> Open query editor for the document.
4. Run the query to show the documents.
5. Select one document from the "Result" list -> Click "Delete selected document" button.
Image
6. Check whether the selected document is removed or not.

**Expect:**
The selected document is removed.

**Actual:**
The selected document is not removed.

**More Info:**
The selected document will be removed after clicking "Reload query results" button.

bk201-:
Fixes #2437

- Deleted documents highlighted by line-through decoration
- No actions with deleted rows

image
image
image
image


bk201-:
Well, then we will say that result is just Query Representation what we can't change as well as undo any actions.

Why can't we remove these records from the table? Because the have statistics and metrics which linked to certain results in the table. My opinion, it is obvious that any actions with records will be represented only after re-running query.

tnaum-ms:
> Well, then we will say that result is just Query Representation what we can't change as well as undo any actions.
> 
> Why can't we remove these records from the table? Because the have statistics and metrics which linked to certain results in the table. My opinion, it is obvious that any actions with records will be represented only after re-running query.

I'd leave it for you, @bk201-, and @sevoku to decide, as it's in your domain. I wouldn't prefer to go down that UX path. If the correlation between the result set and the stats is a concern, I suggest replacing the metrics with a message stating that it's not available as the result set has changed.

sevoku (approved):
I don't have a strong opinion on this, I'd say let's go with this approach and see what feedback we get? We could also use a grey-ish text or background color.

@tnaum-ms don't we have support for deletion in Mongo? How do you handle it? We should shoot for the same experience, especially in such cases.

bk201-:
>  I suggest replacing the metrics with a message stating that it's not available as the result set has changed.

@sevoku @tnaum-ms In this case any deletion will fail all metrics at all, not only for this page for every page.  I'm not competent at this question, I would ask someone who uses both tabs.  Questions like
- How often the metrics are required?
- How is important these data?

tnaum-ms:
> I don't have a strong opinion on this, I'd say let's go with this approach and see what feedback we get? We could also use a grey-ish text or background color.
> 
> @tnaum-ms don't we have support for deletion in Mongo? How do you handle it? We should shoot for the same experience, especially in such cases.

I simply remove it from the database and from the view. It's gone once a user confirms they want to delete. Deleted data being visible suggests an undo option that we don't and likely won't have.

sevoku:
seems we kinda forgot about this :)

@bk201- wrt metrics I'd suggest we show a warning inside the stats tab like "Dataset has changed, please rerun the query to update the metrics". Also we should check and see what happens with the continuation token? If we simply rerun the request would it update properly or do we really need to rerun the whole query?
@tnaum-ms I'm not too much concerned about deleted items being visible. As long as they're visible (no query rerun) we can show a tooltip saying "Item deleted permanently" and if customers really ask for a recover feature, it won't be too hard, we could reconstruct the document from the row and create a new unsaved recovery document users could verify and re-upload.

sevoku:
Let's resurrect :lol: this topic, I have a couple of thoughts how it can be made more useful also with recovery:
1. as @bk201- proposed here, items should stay there in the results (with deleted marker) to still reflect the query run state, even if query metrics might be off
   * we already don't really care about deleted items and iterators, what if something else removes a document externally which is part of the current query? I don't think that we should or even could care about all possible scenarios here. Running a query, especially a large one (many results with pagination) can always be a bit off if the account is operational, that's the nature of any DB queries.
3. Viewing/Editing a deleted Item should open the document editor in "New Item" mode with the content recovered from the result table.
3. Deleted Items should have a tooltip explaining:
   > "This Item has been deleted. You can still edit it to create a new Item with recovered content from this query result, but some data might be missing"
4. what about bulk edits? Maybe we could use the Import feature? i.e. if you select one item, it'd open the editor (see above) and if multiple deleted items are selected, we could generate a valid JSON with multiple items and pass it to the import feature? After showing some confirmation.

PR #2470, opened 2024-11-26
fix: Rename collection to container for NoSQL commands

### LINKED ISSUE #2453: [Suggestion] It would be better to change the container action from "Import Document into a Collection..." to "Import Document into a Container..."
**OS:** Windows 10
**Build Version:** 20241121.1

**Repro Steps:**
1. Create a "Cosmos DB for NoSQL" account.
2. Create a database and a container for the above account.
3. Right click the container node to open its context menu.

**Actual:**
There is an action "Import Document into a Collection..." in the container context menu.
Image

**Suggestion:**
It would be better to change the container action from "Import Document into a Collection..." to "Import Document into a Container...".

bk201-:
Fixes #2453

### DIFF
--- a/package.json
+++ b/package.json
@@ -454,7 +454,7 @@
             {
                 "category": "Cosmos DB",
                 "command": "cosmosDB.importDocument",
-                "title": "Import Document into a Collection..."
+                "title": "Import Document into a Container..."
             },
             {
                 "category": "MongoDB",


bk201-:
@microsoft-github-policy-service agree

tnaum-ms (approved):

PR #2471, opened 2024-11-26
fix: Removes UI blinking while loading the document

### LINKED ISSUE #2457: A brief flash of a screen before the document file is opened in "Cosmos DB for NoSQL" account
**OS:** Windows 10
**Build Version:** 20241120.1

**Repro Steps:**
1. Expand one "Cosmos DB for NoSQL" account -> Right click the "Documents" node and select "Create Document..." action.
2. Check whether the document file is opened directly or not.

**Expect:**
The document file is opened directly.

**Actual:**
A brief flash of a screen before the document file is opened.
Image

bk201-:
Fixes #2457


tnaum-ms (dismissed):
however, I'd probably go with something like `isReady` state and make it more general than just that `isInit` one that's very specific. With `isReady` you'd be able to cover more cases.

bk201-:
@tnaum-ms Agree, renamed

sevoku (approved):

tnaum-ms (approved):

Issue #2440, opened 2024-11-20
Remove hidden input under Monaco Editor


bk201-:
The new hidden input field was added under Monaco Editor component to fix the issue #2431. This input is visible for Narrator and impacts usability for a11y users. After fixing related issues this input must be deleted.

Related issues: microsoft/tabster#316 and microsoft/fluentui#31526

PR #2492, opened 2024-12-03
fix: Fixes tabster behavior around Monaco Editor

bk201-:
Fixes: #2440


tnaum-ms (changes requested):
The current version doesn't allow the use of `tab` when editing documents in the 'Document Editor'. I ran a test for both the NoSQL and the vCore editors. Please confirm whether this is intended.

bk201-:
@tnaum-ms Fixed, please check it again

tnaum-ms:
I have some bad news, on the first "tab"-press in the document editor, I see this:

image

I can't investigate more at the moment and wanted to let you know ASAP, that's why it's just a screenshot.

sevoku:
I tried the current PR build and can't reproduce the error @tnaum-ms, does it consistently reproduce for you?

The only problem I see is missing support for the "Tab Key Moves Focus" toggle (command: `editor.action.toggleTabFocusMode`). When enabled, Monaco should not accept the tab key and let the user leave the editor area. The state is also visible in the status bar:
image

tnaum-ms:
@sevoku @bk201- I wasn’t able to reproduce this bug anymore. Back then, it occurred after just a few steps, but now I’ve tried **everything**, and it no longer happens. Maybe one of the dependencies has been updated... hard to say for sure, but well... 🚀

tnaum-ms (approved):

PR #2533, opened 2025-01-15
Dev/sda/tree api migration

bk201-:


src/commands/deleteDatabaseAccount/deleteCosmosDBAccount.ts
```diff
@@ -27,7 +28,13 @@ export async function deleteCosmosDBAccount(
...
         accountName = getDatabaseAccountNameFromId(node.fullId);
     } else if (node instanceof CosmosAccountResourceItemBase) {
-        const subscriptionContext = createSubscriptionContext(node.account.subscription);
+        // Not all CosmosAccountResourceItemBase instances have a subscription property (attached account does not),
+        // so we need to create a subscription context
+        if (!('subscription' in node.account)) {
+            throw new Error('Subscription is required to delete an account.');
+        }
+
+        const subscriptionContext = createSubscriptionContext(node.account.subscription as AzureSubscription);
```
tnaum-ms, line 37: I see/I get it, it's an extra safety level, however:
- we won't attach this command to the workspace node.
- `CosmosAccountResourceItemBase` is/will/would be different that the future `CosmosAccountWorkspaceItemBase`?, at least that's what I've done with monog-clusters. one base with common functionality, and then one for workspace with connection string handling, and the other one for the resources tree view with subscription.

Let's see if my commend here changes once I finish looking at the rest.

src/commands/deleteDatabaseAccount/deleteDatabaseAccount.ts
```diff
@@ -26,10 +27,14 @@ export async function deleteDatabaseAccount(
...
-        subscription = createSubscriptionContext(node.account.subscription);
+    } else if (node instanceof CosmosAccountResourceItemBase && 'subscription' in node.account) {
+        subscription = createSubscriptionContext(node.account.subscription as AzureSubscription);
+    } else if (node instanceof MongoClusterResourceItem) {
+        subscription = createSubscriptionContext(node.subscription);
     } else {
-        subscription = createSubscriptionContext((node as MongoClusterResourceItem).subscription);
+        // Not all CosmosAccountResourceItemBase instances have a subscription property (attached account does not),
+        // so we need to create a subscription context
+        throw new Error('Subscription is required to delete an account.');
```
tnaum-ms, line 37: same here, I didn't anticipate this code path. but, the safer, the better 👍

src/mongoClusters/tree/CollectionItem.ts
```diff
@@ -35,12 +39,14 @@ export class CollectionItem implements TreeElementWithId, TreeElementWithExperie
     ) {
         this.id = `${mongoCluster.id}/${databaseInfo.name}/${collectionInfo.name}`;
         this.experience = mongoCluster.dbExperience;
+        this.experienceContextValue = `experience.${this.experience?.api ?? API.Common}`;
```
tnaum-ms, line 42: superb, good idea with `experience.` prefix here, solves the issue I stumbled upon when defining `where` for commands in the `package.json`

src/tree/docdb/DocumentDBAccountAttachedResourceItem.ts
```diff
@@ -14,48 +14,30 @@ import { getSignedInPrincipalIdForAccountEndpoint } from '../../docdb/utils/azur
...
-        const result = await callWithTelemetryAndErrorHandling('getChildren', async (context: IActionContext) => {
-            context.telemetry.properties.experience = this.experience.api;
-            context.telemetry.properties.parentContext = this.contextValue;
-            context.errorHandling.rethrow = true;
-
-            const accountInfo = await this.getAccountInfo(this.account);
-            const cosmosClient = getCosmosClient(accountInfo.endpoint, accountInfo.credentials, false);
-            const databases = await this.getDatabases(accountInfo, cosmosClient);
-            return await this.getChildrenImpl(accountInfo, databases);
-        });
```
tnaum-ms, line 44: Just to make sure : telemetry is removed here as the branch data provider is taking care of it.

tnaum-ms (approved):
@bk201- a really clean solution and improvement to the overall structure. thank you for the extra work!

I added some questions/comments mainly to better understand the code in our review later on.

src/tree/docdb/DocumentDBAccountAttachedResourceItem.ts
bk201-: Yes, everything moves to src/tree/CosmosDBWorkspaceBranchDataProvider.ts

src/commands/deleteDatabaseAccount/deleteCosmosDBAccount.ts
bk201-: I didn't make any difference between usual Account and AttachedAccount. Both of them have the same property account, but with different type. As well as MongoAccount. Because it is CosmosAccountResourceItem**Base** class for all type of accounts. 

I suggest to implement the following classes or something like that.
```
CosmosAccountResourceItem**Base** 
|__CosmosAccountResourceItem
|__CosmosAttachedAccountResourceItem
|__MongoAccountResourceItem**Base**
   |__ MongoDBAccountResourceItem
   |__MongoClusterAccountResourceItem
```   
   
Suffix **Base** might be substituted but any appropriate: common, interface, abstract, etc

src/commands/deleteDatabaseAccount/deleteDatabaseAccount.ts
bk201-: It was need only for quick fix merge conflict. See comment above. It might be fixed by perfect class hierarchy

src/commands/deleteDatabaseAccount/deleteCosmosDBAccount.ts
tnaum-ms: Got it. The confusion here came from the fact that I had this type of structure in mongoClusters, ahead of the v1->v2 migration. Once we're done with migration, I'll restructure mongo-related space.

That's what mongo clusters used:

```
MongoClusterItemBase
|__MongoClusterResourceItem
|__MongoClusterWorkspaceItem
```

(`Workspace Item` behaves the same as your `Attached Account ... Item`)

tnaum-ms (approved):

PR #2572, opened 2025-02-20
Refactor emulator configuration to use encapsulated objects

tnaum-ms:
chore: refactores code to use `EmulatorConfiguration`


tnaum-ms:
PR was created by accident

PR #2574, opened 2025-02-20
Refactor emulator configuration to use encapsulated objects

tnaum-ms:
chore: refactored code to use `EmulatorConfiguration`.

Currently MongoDB API emulator support works with two properties: `isEmulator` and `disableEmulatorSecurity`. Instead of working with two properties, a new object `EmulatorConfiguration` was created to encapsulate these properties.


src/docdb/getCosmosClient.ts
```diff
@@ -51,7 +51,9 @@ export function getCosmosClientByConnection(
     const commonProperties: CosmosClientOptions = {
         endpoint,
         userAgentSuffix: appendExtensionUserAgent(),
-        agent: new https.Agent({ rejectUnauthorized: isEmulator ? !isEmulator : vscodeStrictSSL }),
+        agent: new https.Agent({
+            rejectUnauthorized: emulatorConfiguration.isEmulator ? !emulatorConfiguration.isEmulator : vscodeStrictSSL,
```
tnaum-ms, line 55: @sevoku @bk201- FYI: his looks like a special treatment for the Core emulator, I've not investigated it, just noticed during my refactoring session.

tnaum-ms:
✅ I've tested each feature around Mongo* and around adding connections / adding emulators for both, Mongo* and Core.

sevoku (approved):

PR #2591, opened 2025-03-07
feat: Localization

bk201-:
- Moving to the modern VSCode l10n API
- Fixing imports
- Using Azure Translate to generate l10n files for 13 languages


tnaum-ms:
@bk201- any help from my side need at this stage?

PR #2598, opened 2025-03-14
Add tests for the HaTS survey qualification

sevoku:


sevoku:
The tests pass now for da8147d: https://github.com/microsoft/vscode-cosmosdb/actions/runs/13860341031/job/38787010956#step:10:13

src/utils/survey.ts
```diff
@@ -10,65 +10,101 @@ import * as nls from 'vscode-nls';
...
+class SurveyState {
+    private _isCandidate: boolean | undefined = undefined;
+    private _wasPromptedInSession = false;
+    private _usageScoreByExperience: Record<ExperienceKind, number> = {
+        [ExperienceKind.Mongo]: 0,
+        [ExperienceKind.NoSQL]: 0,
+    };
+
+    // In-memory state getters/setters
+    public get isCandidate(): boolean | undefined {
```
bk201-, line 46: NIT: What is a point to use getter and setter if you don't do any checks? It is the same that public isCandidate or public wasPromotedInSession

sevoku: Ah yeah, good catch, that was Copilot and I forgot to change it

tnaum-ms (approved):

PR #2599, opened 2025-03-16
Add a Feedback button to NoSQL query editor

sevoku:
Adds a "Provide Feedback" button to the NoSQL query editor if the user has been identified as a survey candidate. Usage scores will be ignored, hence the button is always visible if the user has not opted out from surveys.

The "smiley" button opens a dropdown with the option to provide feedback:
image
if the toolbar is collapsed, the button turns to a simple menu item without a second step.
image

### DIFF
--- a/src/webviews/mongoClusters/collectionView/components/toolbar/ToolbarMainView.tsx
+++ b/src/webviews/mongoClusters/collectionView/components/toolbar/ToolbarMainView.tsx
@@ -3,18 +3,58 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Menu, MenuItem, MenuList, MenuPopover, MenuTrigger, Toolbar, ToolbarButton } from '@fluentui/react-components';
-import { ArrowClockwiseRegular, ArrowExportRegular, ArrowImportRegular, PlayRegular } from '@fluentui/react-icons';
+import {
+    Menu,
+    MenuItem,
+    MenuList,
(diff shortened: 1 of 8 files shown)


src/webviews/mongoClusters/collectionView/components/toolbar/ToolbarMainView.tsx
```diff
@@ -3,18 +3,58 @@
...
+                                        triggerAction: 'cosmosDB.mongo.collectionView.provideFeedback',
+                                    })
+                                    .catch(() => {});
+                            }}
+                        >
+                            Provide Feedback
+                        </MenuItem>
+                    </MenuList>
+                </MenuPopover>
+            </Menu>
```
tnaum-ms, line 57: I added it as a 'copy&paste' for a reason - I don't have overflow support in yet... and the show/hide button logic is now to be simplified anyway.
Once we start adding more feedback features, we'll be able to extract it to a common element.

bk201- (approved):

src/webviews/mongoClusters/collectionView/components/toolbar/ToolbarMainView.tsx
sevoku: You should probably add a check for `getIsSurveyDisabledGlobally()` which will be changed to support the admin conset setting once it's available. Otherwise we'll need to modify the Mongo part when we get there.

tnaum-ms (approved):

tnaum-ms (approved):

PR #2627, opened 2025-03-26
Add setting for the CSV separator

mkrueger:
This pull request introduces several changes to the CSV export functionality for Cosmos DB queries and metrics. 
It adds a new configuration option for the CSV separator which defaults to the ';' char. Excel has less problems to use the ';' char  as separator.

For accessing the settings it was needed to move the CSV generation from frontend to backend.

### Enhancements to CSV export functionality:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R1280-R1284): Added a new configuration option `cosmosDB.csvSeparator` to specify the separator for CSV exports.
* [`package.nls.json`](diffhunk://#diff-9b29d8b47a74097a04da353b72a8632d485d658282457b97b4caf26575a42dd0R84): Added a description for the new `cosmosDB.csvSeparator` configuration option.

### Updates to `QueryEditorTab` class:

* [`src/panels/QueryEditorTab.ts`](diffhunk://#diff-481b7c7d679172d98c1a1aabd3cda871ebd10b3d747f2afa2ea1dfc3245842b7L16-R23): Added methods to handle saving and copying CSV and metrics CSV, and updated the `getCommand` method to include these new operations. [[1]](diffhunk://#diff-481b7c7d679172d98c1a1aabd3cda871ebd10b3d747f2afa2ea1dfc3245842b7L16-R23) [[2]](diffhunk://#diff-481b7c7d679172d98c1a1aabd3cda871ebd10b3d747f2afa2ea1dfc3245842b7R152-R170) [[3]](diffhunk://#diff-481b7c7d679172d98c1a1aabd3cda871ebd10b3d747f2afa2ea1dfc3245842b7R440-R476)

### Modifications to CSV conversion utilities:

* [`src/utils/convertors.ts`](diffhunk://#diff-58cc5c765c72f86b21c2cf2e270040ae521edbb242645ed484107e77f47a7286R503-R512): Introduced functions to set and get the CSV separator, and updated the CSV conversion functions to use the configurable separator. [[1]](diffhunk://#diff-58cc5c765c72f86b21c2cf2e270040ae521edbb242645ed484107e77f47a7286R503-R512) [[2]](diffhunk://#diff-58cc5c765c72f86b21c2cf2e270040ae521edbb242645ed484107e77f47a7286L512-R524) [[3]](diffhunk://#diff-58cc5c765c72f86b21c2cf2e270040ae521edbb242645ed484107e77f47a7286L527-R537) [[4]](diffhunk://#diff-58cc5c765c72f86b21c2cf2e270040ae521edbb242645ed484107e77f47a7286L546-R559)

### Refactoring in `ResultPanelToolbarOverflow`:

* [`src/webviews/cosmosdb/QueryEditor/ResultPanel/ResultPanelToolbarOverflow.tsx`](diffhunk://#diff-c128b1a5dade1fa53b33a1df356e88d39834d06b5b8fd9057153d56a4005c078L50-R50): Updated the methods for saving and copying CSV to use the new dispatcher methods. [[1]](diffhunk://#diff-c128b1a5dade1fa53b33a1df356e88d39834d06b5b8fd9057153d56a4005c078L50-R50) [[2]](diffhunk://#diff-c128b1a5dade1fa53b33a1df356e88d39834d06b5b8fd9057153d56a4005c078L441-R444) [[3]](diffhunk://#diff-c128b1a5dade1fa53b33a1df356e88d39834d06b5b8fd9057153d56a4005c078L513-R510)

### Dispatcher updates:

* [`src/webviews/cosmosdb/QueryEditor/state/QueryEditorContextProvider.tsx`](diffhunk://#diff-1df36211913c2cc5ead27d10a2777b18da82c9fb9bd8628954373aa96ae7c4eeR110-R133): Added new methods to the `QueryEditorContextProvider` class to handle CSV and metrics CSV operations.


sevoku (changes requested):
Can we fix SettingsService then? ~or just use `vscode.workspace.getConfiguration()` instead. ~We still use `vscode.workspace.getConfiguration` across the board and can update it everywhere once Settings Service is fixed. Let's don't block on that here.~ EDIT: I'm stupid, this is on the WebView side.

src/utils/convertors.ts
```diff
@@ -543,8 +548,8 @@ export const queryResultToCsv = (
                 rowValues.push(escapeCsvValue(value));
             });
 
-            return rowValues.join(',');
+            return rowValues.join(getCsvSeperator());
         })
         .join('\n');
-    return `sep=,\n${headers}\n${rows}`;
+    return `\n${headers}\n${rows}`;
```
bk201-, line 554: I think we have still set `sep=` in order to Excel/other tools understand which separator to use

mkrueger: sep= seems to be a excel only thing.  so doesn't give good results on other tools

sevoku: Not really, Mike and I tested it, if the Sep is `;` then Excel detects it properly without the `sep=;` header and at the same time the header is not supported by others (i.e. Apple Numbers or LibreOffice), so we decided to remove it again. Does it work for you without the header @bk201-? We could also introduce an additional setting for it...

src/utils/convertors.ts
```diff
@@ -500,6 +501,10 @@ export const escapeCsvValue = (value: string): string => {
     return `"${value.replace(/"/g, '""')}"`;
 };
 
+function getCsvSeperator():string {
+    return SettingsService.getSetting<string>('cosmosDB.csvSeparator') ?? ';';
```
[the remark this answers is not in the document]
sevoku: NVM, can't be used in a web view.

sevoku:
> Can we fix SettingsService then? ~or just use `vscode.workspace.getConfiguration()` instead. We still use `vscode.workspace.getConfiguration` across the board and can update it everywhere once Settings Service is fixed. Let's don't block on that here.~ EDIT: I'm stupid, this is on the WebView side.

@bk201- is the SettingService supposed to be available on the WebView side at all? We probably need to either update the setting from the VS Code side with the state, or move CSV conversion our of the web view?

src/utils/convertors.ts
bk201-: OK, I've just asked to be sure. Then @mkrueger could you please remove first `\n`

src/utils/convertors.ts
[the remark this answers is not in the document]
mkrueger: was another issue.

src/utils/convertors.ts
mkrueger: did it in the last commit.

mkrueger:
Moved the csv generation to the vs code side - webview just creates the events now.

src/panels/QueryEditorTab.ts
```diff
@@ -412,4 +437,41 @@ export class QueryEditorTab extends BaseTab {
...
+    }
+
+    private async saveCSV(
+        name: string,
+        currentQueryResult: SerializedQueryResult | null,
+        partitionKey?: PartitionKeyDefinition,
+    ): Promise<void> {
+        this.updateCsvSeparator();
+        const text = queryResultToCsv(currentQueryResult, partitionKey);
+        await vscodeUtil.showNewFile(text, name, 'csv');
```
sevoku, line 453:
```suggestion
        await vscodeUtil.showNewFile(text, name, '.csv');
```

src/panels/QueryEditorTab.ts
```diff
@@ -412,4 +437,41 @@ export class QueryEditorTab extends BaseTab {
...
+    }
+
+    private async saveMetricsCSV(name: string, currentQueryResult: SerializedQueryResult | null): Promise<void> {
+        this.updateCsvSeparator();
+        const text = queryMetricsToCsv(currentQueryResult);
+        await vscodeUtil.showNewFile(text, name, 'csv');
```
sevoku, line 459:
```suggestion
        await vscodeUtil.showNewFile(text, name, '.csv');
```

src/webviews/cosmosdb/QueryEditor/ResultPanel/ResultPanelToolbarOverflow.tsx
```diff
@@ -510,15 +503,11 @@ const ExportButton = forwardRef(
         async function onSaveAsCSV() {
             const filename = `${state.dbName}_${state.collectionName}_${state.currentQueryResult?.activityId ?? 'query'}`;
             if (selectedTab === 'result__tab') {
-                await dispatcher.saveToFile(
-                    queryResultToCsv(state.currentQueryResult, state.partitionKey),
-                    `${filename}_result`,
-                    'csv',
-                );
+                await dispatcher.saveCSV(`${filename}_result`, state.currentQueryResult, state.partitionKey);
```
sevoku, line 506: Should this also support selection?

src/panels/QueryEditorTab.ts
```diff
@@ -412,4 +437,41 @@ export class QueryEditorTab extends BaseTab {
         openSurvey(ExperienceKind.NoSQL, 'cosmosDB.nosql.queryEditor.provideFeedback');
         return Promise.resolve();
     }
+
+    private updateCsvSeparator() {
```
sevoku, line 441: Now that this lives in the host, there is no need for this, just pass the result from `SettingsService.getSetting` to the converter function as parameter, or even better, just let the converter function get the setting directly.

src/utils/convertors.ts
```diff
@@ -500,6 +500,16 @@ export const escapeCsvValue = (value: string): string => {
     return `"${value.replace(/"/g, '""')}"`;
 };
 
+let csvSeparator: string = ';';
+
+export function setCsvSeperator(seperator: string): void {
```
sevoku, line 505: See above, loading the setting directly here would be much cleaner without any getter/setter.

mkrueger: Problem is that I always get doing that:
image

src/panels/QueryEditorTab.ts
mkrueger: Problem is that I always get doing that - so this is a work around:
image

Not sure why TBH - doesn't really make sense to me.

src/utils/convertors.ts
mkrueger: see above.

src/panels/QueryEditorTab.ts
sevoku: @mkrueger Well the problem is that convertors.ts is now imported from both the WebView side and the host side, with the WebView side not being able to access vscode APIs. You can refactor the CSV converter in a separate csvConverter.ts so it won't get imported by the WebView code.

package.nls.json
```diff
@@ -81,6 +81,7 @@
   "cosmosdb.configuration.oauth.enable": "Always authenticate using Entra ID RBAC when connecting to Cosmos DB NoSQL Accounts. [Enable RBAC and assign Roles](https://aka.ms/cosmos-native-rbac).",
   "cosmosdb.configuration.operation.summaries": "Show detailed operation summaries, displaying messages for actions such as database drops, document additions, deletions, or similar events.",
   "cosmosdb.configuration.save.prompt": "Show warning dialog when uploading a document to the cloud.",
+  "cosmosdb.configuration.csv.separator": "The separator to use when exporting data to a CSV file (i.e. ;, | or ,).",
```
sevoku, line 84: Let's use real markup here:
```suggestion
  "cosmosdb.configuration.csv.separator": "The separator to use when exporting data to a CSV file (i.e. `;`, | or `,`).",
```

sevoku (commented):
Looks good! Let's address the remaining comments and it should be good to go.

src/webviews/cosmosdb/QueryEditor/ResultPanel/ResultPanelToolbarOverflow.tsx
sevoku: Filed separately: https://github.com/microsoft/vscode-cosmosdb/issues/2639

sevoku (approved):

Issue #2620, opened 2025-03-21
Allow attaching a Cosmos DB Account without providing an AccountKey


sevoku:
Attaching a Cosmos DB Account is currently only supported if the connections string contains an Account Key, however we do support Entra ID auth for attached accounts as well, hence the check should be removed.

This would also allow attaching accounts with `localAuth` disabled.

Limitation: for Entra ID we need an optional tenantId for Multi-Tenant support, otherwise VS Code won't be able to get the right token if the current proncipal is not part of the same tenant.

PR #2635, opened 2025-03-29
Add an UriHandler to handle external requests to open/attach a specific account and open the Query Editor

### LINKED ISSUE #2532: Support opening the Query Editor from outside (Portal, Fabric) using registerUriHandler
We want to add "Open in VS Code" buttons to the Azure Portal and Fabric, for this we need to implement a handler using [vscode.window.registerUriHandler](https://code.visualstudio.com/api/references/vscode-api#window.registerUriHandler) (on activation) which will open the selected database/container in the Query Editor. 

Ideally we would detect whether the resource exists in the Azure Resources tree and navigate to there, however that's currently not supported by the Tree View API.

The simplest approach (per offline discussion) is to support following Uri parameters:
* connection string (can support any database type)
   * we should be able to detect which database SDK it is (NoSQL/Mongo) and act accordingly
* Database Name
* Container Name

Following actions would be taken:
* Look up if an Attached Account with the same connection string exists
   * Add an "Attached Account" entry if not
   * Expand the node
* Open the Query editor for the given database and collection

sevoku:
This PR adds a UriHandler with following features:
* Reveal an existing account in the Azure Resources tree
* Attach the account to the Workspace if it is not an Azure Resource
* Open the Query editor for the specified Collection/Container if available

Following parameters are supported (with dependencies):
* `resourceId`: The full Azure Resource Id of the account
   * Pattern:`vscode://ms-azuretools.vscode-cosmosdb?resourceId=[accountResourceId]`
* `cs`: The NoSQL/Mongo Connection String (For NoSQL optionally with or without the `AccountKey` parameter, `AccountEndpoint` is enough to authenticate with EntraID RBAC)
   * Optional Parameters (required to reveal the account in Azure Resources, will create an Attached Workspace Item otherwise)
      * `subscriptionId`: Azure Subscription ID 
      * `resourceGroup`: Azure Resource Group Name (required for `cs`)
      * `tenantId`: Id of the tenant the resource belongs to (TODO: not yet implemented!)
   * Pattern: 
   `vscode://ms-azuretools.vscode-cosmosdb?subscriptionId=[subscriptionId]&resourceGroup=[resourceGroupName]&cs=AccountEndpoint=[connectionString]`
* (optional) required to open the Query Editor:
   * `database`: Database Name
   * `container`: Container/Collection Name
   * Pattern: `vscode://ms-azuretools.vscode-cosmosdb?resourceId=[accountResourceId]&database=[databaseName]&container=[containerName]`

**IMPORTANT:** parameters – especially `cs` – must be encoded using `encodeURIComponent()`!

Examples:
```
vscode://ms-azuretools.vscode-cosmosdb?resourceId=/subscriptions/b15d5cd6-bff3-48c1-8d58-25eeb753034b/resourceGroups/my-resource-group/providers/Microsoft.DocumentDB/databaseAccounts/my-cosmosdb-account&database=Shop&container=Products
```
```
vscode://ms-azuretools.vscode-cosmosdb?subscriptionId=b15d5cd6-bff3-48c1-8d58-25eeb753034b&resourceGroup=my-resource-group&database=Shop&container=Products&cs=AccountEndpoint=https://cosmos.documents.azure.com:443/
```

### Dependency Updates:
* Updated activation events in `package.json` to include `onView:azureResourceGroups` and `onUri`.
* Updated `@microsoft/vscode-azext-utils` to version `2.6.6`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L151-R153)

### Function Enhancements:
* Enhanced `openNoSqlQueryEditor` function to handle different input types and added a helper function to extract connections from container nodes in `src/commands/openNoSqlQueryEditor/openNoSqlQueryEditor.ts`.

### Cosmos DB Connection Handling:
* Added `accountName` getter to `ParsedConnectionString` class in `src/ParsedConnectionString.ts`.
* Improved error handling and parsing logic in `parseCosmosDBConnectionString` and `ParsedCosmosDBConnectionString` classes in `src/cosmosdb/cosmosDBConnectionStrings.ts`. [[1]](diffhunk://#diff-1c064d64fc2d186a21a72d65d30d5a0c6e3620230addb9b90d778ecd0f01283fL7-R14) [[2]](diffhunk://#diff-1c064d64fc2d186a21a72d65d30d5a0c6e3620230addb9b90d778ecd0f01283fL34-R50)
* Added support for handling connection strings in `getAccountInfo` function in `src/tree/cosmosdb/AccountInfo.ts`.

### Miscellaneous:
* Registered a global URI handler in `src/extension.ts` to handle URIs.

TODO:
* [x] ~~add `tenantId` parameter to support multi-tenant configs~~ Should be done separately as part of https://github.com/microsoft/vscode-cosmosdb/pull/2635
* [x] Implement `ClustersWorkspaceBranchDataProvider.getParent()` similar to `CosmosDBWorkspaceBranchDataProvider`
* [x] Implement `MongoVCoreBranchDataProvider.getPArent()` similar to `CosmosDBBranchDataProvider`
* [x] Implement Mongo support in [`openAppropriateEditorForConnection()`](https://github.com/microsoft/vscode-cosmosdb/pull/2635/files#diff-e658a4603f6f310947489252c2deea4645903dc5b8fcf18c7f1ccc94ab9a21f8R362-R377)
* [x] Update [`revealAttachedInWorkspaceExplorer`](https://github.com/microsoft/vscode-cosmosdb/pull/2635/files#diff-e658a4603f6f310947489252c2deea4645903dc5b8fcf18c7f1ccc94ab9a21f8R316-R321) to use `revealWorkspaceResource` once https://github.com/microsoft/vscode-azureresourcegroups/pull/1108 becomes available in [@microsoft/vscode-azext-azureutils](https://www.npmjs.com/package/@microsoft/vscode-azureresources-api/).
* [ ] `BranchDataProvider.refresh()` is never called for root node, i.e. select to "Refresh" a resource group, it will call `getResourceItem(resource)` right away, so there is no way for us to cache root (account/cluster) nodes. We need to either patch this in Azure Resources or find another way to detect whether to invalidate cache or not (maybe a timestamp on the resource model?)
* [x] Review telemetry and masking

Fixes #2532
Fixes partially #2620


src/cosmosdb/cosmosDBConnectionStrings.ts
```diff
@@ -31,15 +29,26 @@ export class ParsedCosmosDBConnectionString extends ParsedConnectionString {
...
+        databaseName: string | undefined,
+    ) {
         super(connectionString, databaseName);
         this.documentEndpoint = endpoint;
         this.masterKey = masterKey;
 
-        const parsedEndpoint = url.parse(endpoint);
-        this.hostName = nonNullProp(parsedEndpoint, 'hostname', 'hostname');
-        this.port = nonNullProp(parsedEndpoint, 'port', 'port');
+        const parsedEndpoint = new URL(endpoint);
```
bk201-, line 44: if I'm not mistaken the `new URL()` throws an error in case of wrong url besides `url.parse`

image

image

sevoku: @bk201- I think that's fine, the JS SDK does the same: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/cosmosdb/cosmos/src/CosmosClient.ts#L98 -> https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/cosmosdb/cosmos/src/utils/checkURL.ts#L4-L6

sevoku: @bk201- See 46101db

src/utils/emulatorUtils.ts
```diff
@@ -0,0 +1,136 @@
...
+ * Generates a unique ID for an emulator item based on the connection string.
+ * The ID is prefixed with 'emulator-' and is derived from a hash of the connection string.
+ *
+ * @param connectionString - The connection string to hash
+ * @returns A unique ID for the emulator item
+ */
+export function getEmulatorItemUniqueId(connectionString: string): string {
+    const migratedMarker = 'emulator-';
+    return `${migratedMarker}${randomUtils.getPseudononymousStringHash(connectionString, 'hex').substring(0, 24)}`;
+}
```
tnaum-ms, line 105: BTW: I think we should modify `TreeItemWithId` or add an id generator somewhere that ensures that no illegal characters are in the id (I think `/` might break the parent search.. I'm looking for it during this review.)

src/tree/BaseCachedBranchDataProvider.ts
```diff
@@ -0,0 +1,443 @@
...
+     * This helps identify which view/provider an item belongs to when constructing
+     * context value strings for commands, menu visibility, and error elements.
+     *
+     * @returns A string prefix for context values (e.g., 'cosmosDB', 'cosmosDB.workspace')
+     * @example
+     * protected get contexValuePrefix(): string {
+     *     return 'cosmosDB.workspace';
+     * }
+     */
+    protected abstract get contextValuePrefix(): string;
```
tnaum-ms, line 81: I'd suggest a change to `contextValue` and then using the `createContextValue` function. It generates a `;`-separated list of context values.
This will work well with the usecase here for of creating error elements: `contextValue: createContextValue(contextValue, 'item.error')`

It's cleaner and easier to work with when writing regex for the `package.json`

src/tree/BaseCachedBranchDataProvider.ts
```diff
@@ -0,0 +1,443 @@
...
+
+        // Element's own parent method
+        if (element.getParent && typeof element.getParent === 'function') {
+            // some tree elements keep track of their parents (documentdb clusters).
+            // we rely on this to get the parent element.
+            return element.getParent();
+        }
+
+        // ID-based parent lookup
+        const parentId = element.id?.substring(0, element.id.lastIndexOf('/'));
```
tnaum-ms, line 238: For now we need to remember that an `id` can't contain illegal characters (here it's a `/`).
Maybe we can embed it into `TreeItemWithId`, however it's an interface/type today, and not an abstract class, so it'd be some work to update this.

src/tree/workspace-view/documentdb/AccountsItem.ts
```diff
@@ -33,7 +33,7 @@ export class AccountsItem implements TreeElement, TreeElementWithExperience {
                 .filter((item) => !item.properties?.isEmulator) // filter out emulators
                 .map((item) => {
                     const model: ClusterModel = {
-                        id: item.id,
+                        id: `${this.id}/${item.id}`, // To enable TreeView.reveal, we need to have a unique nested id
```
tnaum-ms, line 36: we're changing the `id` so now the tree item and the storage item will have different `id` values. This can be a critical issue.

**I might be wrong, but I'm flagging it now so that we test creating / deleting of entries for all connections.**

src/tree/workspace-view/documentdb/LocalEmulators/LocalEmulatorsItem.ts
```diff
@@ -26,29 +27,33 @@ export class LocalEmulatorsItem implements TreeElement, TreeElementWithContextVa
...
+                        const { id, name, properties, secrets } = await migrateRawEmulatorItemToHashed(item);
+                        // we need to create the emulator configuration object from
+                        // the flat properties object
+                        const emulatorConfiguration: EmulatorConfiguration = {
+                            isEmulator: true,
+                            disableEmulatorSecurity: !!properties?.disableEmulatorSecurity,
+                        };
+
+                        const model: ClusterModel = {
+                            id: `${this.id}/${id}`, // To enable TreeView.reveal, we need to have a unique nested id
```
tnaum-ms, line 44: we're changing the `id` so now the tree item and the storage item will have different `id` values. This can be a critical issue.

**I might be wrong, but I'm flagging it now so that we test creating / deleting of entries for all connections.**

src/tree/workspace-view/cosmosdb/LocalEmulators/LocalCoreEmulatorsItem.ts
```diff
@@ -45,32 +46,32 @@ export class LocalCoreEmulatorsItem implements TreeElement, TreeElementWithConte
...
+                    .filter((item) => item.properties?.isEmulator) // only show emulators
+                    .map(async (item) => {
+                        const { id, name, properties, secrets } = await migrateRawEmulatorItemToHashed(item);
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                        const api: API = nonNullValue(properties?.api, 'api') as API;
+                        const isEmulator: boolean = !!nonNullValue(properties?.isEmulator, 'isEmulator');
+                        const connectionString: string = nonNullValue(secrets?.[0], 'connectionString');
+                        const experience = getExperienceFromApi(api);
+                        const accountModel: CosmosDBAttachedAccountModel = {
+                            id: `${this.id}/${id}`, // To enable TreeView.reveal, we need to have a unique nested id
```
tnaum-ms, line 61: we're changing the `id` so now the tree item and the storage item will have different `id` values. This can be a critical issue.

**I might be wrong, but I'm flagging it now so that we test creating / deleting of entries for all connections.**

src/tree/workspace-view/cosmosdb/CosmosDBWorkspaceItem.ts
```diff
@@ -64,7 +64,7 @@ export class CosmosDBWorkspaceItem implements TreeElement, TreeElementWithContex
                     const connectionString: string = nonNullValue(secrets?.[0], 'connectionString');
                     const experience = getExperienceFromApi(api);
                     const accountModel: CosmosDBAttachedAccountModel = {
-                        id,
+                        id: `${this.id}/${id}`, // To enable TreeView.reveal, we need to have a unique nested id
```
tnaum-ms, line 67: we're changing the `id` so now the tree item and the storage item will have different `id` values. This can be a critical issue.

**I might be wrong, but I'm flagging it now so that we test creating / deleting of entries for all connections.**

src/tree/workspace-api/SharedWorkspaceResourceProvider.ts
```diff
@@ -66,3 +67,69 @@ export class SharedWorkspaceResourceProvider implements WorkspaceResourceProvide
...
+    }
+    // Extract the last segment of the ID
+    return trimmedId.substring(lastIndex + 1);
+}
+
+/**
+ * Extracts the workspace resource ID from a MongoDB tree item.
+ *
+ * Unlike standard tree items, MongoDB items contain '/' characters in their IDs,
+ * requiring special handling to correctly extract the resource identifier.
```
tnaum-ms, line 103: :) we've seen the same issue, but from two different angles (you'll see it when reading my other comments)

src/tree/workspace-api/SharedWorkspaceResourceProvider.ts
```diff
@@ -66,3 +67,69 @@ export class SharedWorkspaceResourceProvider implements WorkspaceResourceProvide
...
+ *
+ * This function removes the expected prefix structure to isolate the actual resource ID.
+ *
+ * @param node - The MongoDB tree item node containing an ID property
+ * @returns The extracted MongoDB resource ID
+ * @throws Error if the ID is not a valid workspace resource ID
+ *
+ * @remarks Long-term solution would be to avoid '/' characters within individual ID segments.
```
tnaum-ms, line 111: 👍

src/tree/documentdb/DatabaseItem.ts
```diff
@@ -21,7 +21,12 @@ export class DatabaseItem implements TreeElement, TreeElementWithExperience, Tre
 
     private readonly experienceContextValue: string = '';
 
+    getParent(): TreeElement | undefined | null {
+        return this.parent;
+    }
+
     constructor(
+        readonly parent: TreeElement,
```
tnaum-ms, line 29: marked for deletion, waiting for your feedback in my other comment in the `TreeElement` file

src/tree/documentdb/CollectionItem.ts
```diff
@@ -21,7 +21,12 @@ export class CollectionItem implements TreeElement, TreeElementWithExperience, T
 
     private readonly experienceContextValue: string = '';
 
+    getParent(): TreeElement | undefined | null {
+        return this.parent;
+    }
+
     constructor(
+        readonly parent: TreeElement,
```
tnaum-ms, line 29: marked for deletion, waiting for your feedback to my other comment in the `TreeElement` file

tnaum-ms (commented):
I've read all files, left comments. I will test saving/deleting connections in the workspace as I'm worried about it.
I suggested a change to the handling of the `contextValue` in the BaseCachedBranchDataProvider.

src/tree/TreeElement.ts
```diff
@@ -9,6 +9,17 @@ import type * as vscode from 'vscode';
...
+    /**
+     * Optional method to return the parent of Tree Element / Item.
+     * Return `null` or `undefined` in case of a child of root.
+     *
+     * **NOTE:** This method should be implemented in order to use the {@link TreeView.reveal reveal} API
+     *           implemented in branch data providers.
+     *
+     * @returns Parent of the tree item / tree element.
+     */
+    getParent?(): ExtTreeElementBase | undefined | null;
```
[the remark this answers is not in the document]
sevoku: Reverted [ad31d31](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/ad31d3143c36b5f736933253f58945d0c9aba541)

src/tree/documentdb/DatabaseItem.ts
sevoku: Reverted [ad31d31](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/ad31d3143c36b5f736933253f58945d0c9aba541)

src/tree/documentdb/CollectionItem.ts
sevoku: Reverted [ad31d31](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/ad31d3143c36b5f736933253f58945d0c9aba541)

src/tree/documentdb/ClusterItemBase.ts
```diff
@@ -25,6 +25,10 @@ export abstract class ClusterItemBase implements TreeElement, TreeElementWithExp
 
     private readonly experienceContextValue: string = '';
 
+    getParent(): ExtTreeElementBase | undefined | null {
+        return null;
+    }
```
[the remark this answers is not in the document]
sevoku: Reverted [ad31d31](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/ad31d3143c36b5f736933253f58945d0c9aba541)

src/tree/BaseCachedBranchDataProvider.ts
sevoku: done: 85a83bf81ccf6fc1a0bf6d519c13a6f1f8128a82

src/tree/workspace-view/cosmosdb/LocalEmulators/LocalCoreEmulatorsItem.ts
sevoku: Fixed with https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/a7057df2da95479e9e576e5606b83118e63b27dd

src/tree/workspace-view/documentdb/AccountsItem.ts
sevoku: Fixed with https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/a7057df2da95479e9e576e5606b83118e63b27dd

src/tree/workspace-view/documentdb/LocalEmulators/LocalEmulatorsItem.ts
sevoku: Fixed with https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/a7057df2da95479e9e576e5606b83118e63b27dd

src/tree/BaseCachedBranchDataProvider.ts
sevoku: It'd be good to have this checked, but probably out of scope here right now. In first place we need to make sure not to create invalid items.

src/utils/emulatorUtils.ts
sevoku: see https://github.com/microsoft/vscode-cosmosdb/pull/2635#discussion_r2068562797

src/tree/workspace-api/SharedWorkspaceResourceProvider.ts
sevoku: This is not used anymore: [6c1311f](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/6c1311f8e5a315df0b293304788f0059afb55694)

src/tree/workspace-api/SharedWorkspaceResourceProvider.ts
sevoku: This is not used anymore [6c1311f](https://github.com/microsoft/vscode-cosmosdb/pull/2635/commits/6c1311f8e5a315df0b293304788f0059afb55694), however in theory Attached Mongo accounts could still have an "/" in their IDs because we use `parsedCS.username + '@' + parsedCS.redact().toString()` fir `id` here: https://github.com/microsoft/vscode-cosmosdb/blob/8465ff28ab3249b8377586ec178f6bbd5090ac56/src/commands/newConnection/MongoExecuteStep.ts#L35
@tnaum-ms: can't we change that to use just the `username@hostname:port` instead of the full redacted string? I don't see any point in having all the other optional parameters being part of the ID.

tnaum-ms: @sevoku I'm on it today. This is correct, it doesn't have to be the connection string. The rationale behind it was that the connection string can contain a database name so the current solution was the quick one. However, I'll improve it.

src/utils/emulatorUtils.ts
tnaum-ms: agreed

src/tree/workspace-api/SharedWorkspaceResourceProvider.ts
tnaum-ms: Fixed here:

https://github.com/microsoft/vscode-cosmosdb/blob/728401dcae50acaf030361d7449025b78fff79f4/src/commands/newConnection/MongoExecuteStep.ts#L35-L43

matching migration code is in place as well, so that now all entries are consistent.

tnaum-ms:
> I tried out creating/listing/deleting connections. I discovered only one failing scenario:
> 
> * create a MongoDB account in the workspace from a connections string `mongodb://usd:asdf@test.com`
> * attempt to delete it

✅ The issue no longer occurs.

tnaum-ms:
@sevoku I looked at the `BaseCachedBranchDataProvider` implementation and its role/function. Caching seems to be more the 'how it works', but the actual role of the class is different. 

To get there faster, I asked Copilot to come up with some suggestions, for you to choose from :)

---
Here are three better name suggestions that more accurately reflect the class's purpose:

**BaseNavigableBranchDataProvider**
Emphasizes the navigation capabilities (getParent, findNodeById, findChildById) that are the core functionality needed by VS Code's tree view system.

**BaseHierarchicalBranchDataProvider**
Focuses on the parent-child relationship management that enables tree traversal while also acknowledging the resource-based nature of the tree elements.

**BaseStructuredBranchDataProvider**
This name focuses on the class providing the underlying structure and relationships for the tree data, including caching and navigation capabilities.

tnaum-ms (dismissed):

src/commands/newConnection/MongoExecuteStep.ts
```diff
@@ -31,8 +32,11 @@ export class MongoExecuteStep extends AzureWizardExecuteStep<NewConnectionWizard
                 async () => {
                     await new Promise((resolve) => setTimeout(resolve, 250));
 
+                    const hashedCS = randomUtils.getPseudononymousStringHash(connectionString, 'hex').substring(0, 24);
+                    const storageId = `storageId-${parsedCS.hosts.join('_')}-${hashedCS}`;
```
sevoku, line 36: The same change needs to be made in [vscodeUriHandler.createAttachedForConnection](https://github.com/microsoft/vscode-cosmosdb/pull/2635/files#diff-e658a4603f6f310947489252c2deea4645903dc5b8fcf18c7f1ccc94ab9a21f8R359-R438)

But I'm also wondering... do we really need to hash here? A host is a domain and not an URI hence it should not include any '/', or can it be something else? If it can't, we should just use the host as ID.

tnaum-ms: Thank you for the pointer to `createAttachedForConnection`, I forgot about it.

I'm hashing because the MongoDB / DocumentDB connection string can contain a database name, and maybe there are other critical parameters that I might not be aware of today... I just realized that I need to handle TLS config for emulators correctly...

So I'd rather keep it that way to ensure that I don't overwrite any stored item by accident.

tnaum-ms (approved):

PR #2684, opened 2025-05-12
insert with buffer

xingfan-msft:


src/commands/importDocuments/importDocuments.ts
```diff
@@ -20,6 +20,92 @@ import { CollectionItem } from '../../tree/documentdb/CollectionItem';
...
+        }
+
+        let flushResponse: { count: number; error: Array<{ index: number; errmsg: string }> | undefined } | undefined;
+        if (this.documents.length >= this.maxBatchSize || this.totalSize + docSize > this.maxTotalSize) {
+            flushResponse = await this.flush();
+        }
+
+        this.documents.push(document);
+        this.totalSize += docSize;
+        return flushResponse;
```
tnaum-ms, line 51: I understand the logic here.

We insert documents, but get no feedback, i.e. the return value is `undefined` and I'm supposed to interpret it as `success`: the document has been added to the buffer.

Then, once the buffer is flushed on it's own due to it being full, I'd get a response stating how many documents have been inserted.

On top, the `flush` function is public so I can call it in case of me being done with a buffer and no documents left. I see that you're doing it in another section.

**However**, it's hard to read and the behavior of the entire class is unexpected with use over time. Some calls will be fast, some will be slow. I'd suggest splitting it so that `addDocuments` focuses on filling the buffer and returning a success/failure information. Boolean would do, but a status response might be helpful as well (although today the only rejection reason would be the fact that the buffer is full.).

Then, The user would call flush when needed.

Having a `size` function on the buffer would be helpful.

src/commands/importDocuments/importDocuments.ts
```diff
@@ -9,7 +9,7 @@ import { AzExtResourceType } from '@microsoft/vscode-azureresources-api';
 import { parse as parseJson } from '@prantlf/jsonlint';
 import * as l10n from '@vscode/l10n';
 import { EJSON, type Document } from 'bson';
-import * as fs from 'node:fs/promises';
+import * as fse from 'fs-extra';
 import * as vscode from 'vscode';
```
tnaum-ms, line 13: what is the benefit of `fse` here? (out of interest, I'm not saying it's a bad choice.)

tnaum-ms (changes requested):
@xingfan-msft **Thank you for your first PR. Congratulations!** 🎉

My comments in this review are there to help us align with the way we work in this repository.

- I suggested changes to the buffer implementation to make it easier to use.
- I've not started the extension with your changes, but it looks like it would work
- I've spotted changes to `package.json` that were unexpected, this reminds me that I didn't share our dev-guide with you (note-to-self: we should really add it to this repo), but in general, in this case: we prefer more commits instead of a single one where many change are bundled together - we're aware that a clean cut is not always possible, but even if it's fuzzy, it still helps to follow the 'chain-of-thought' ;)

I'll follow up on this in our call later today.

PR #2685, opened 2025-05-13
FIX: adds support for launching `mongosh` when the user work with PS on Windows

tnaum-ms:
This pull request updates the `launchShell` command to ensure compatibility with different terminal environments on Windows. The main change introduces logic to correctly handle environment variable syntax based on whether the terminal is PowerShell or another shell.

### Compatibility improvements for Windows terminals:

* [`src/commands/launchShell/launchShell.ts`](diffhunk://#diff-eab2a39f022b72c2579c2a4745ffc8ca8a1dd5194daf877365d63520bc8f0465L58-R74): Added a check to determine if the terminal is PowerShell (`PowerShell` or `pwsh`) and updated the `username` and `password` variable syntax accordingly. PowerShell now uses `$env:USERNAME` and `$env:PASSWORD`, while other Windows shells continue to use `%USERNAME%` and `%PASSWORD%`.

### DIFF
--- a/src/commands/launchShell/launchShell.ts
+++ b/src/commands/launchShell/launchShell.ts
@@ -55,8 +55,23 @@ export async function launchShell(
     const username = connectionString.username;
     const password = connectionString.password;
 
-    connectionString.username = isWindows ? '%USERNAME%' : '$USERNAME';
-    connectionString.password = isWindows ? '%PASSWORD%' : '$PASSWORD';
+    // Check if PowerShell is being used on Windows
+    const isWindowsPowerShell =
+        isWindows &&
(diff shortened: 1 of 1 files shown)


bk201- (approved):

tnaum-ms:
The shell integration has been improved in our partner extension `DocumentDB for VS Code`, the goal now is to bring this update over here. Since it's a `DocumentDB` only related challenge, it will be moved over here manually:

The task is to understand the terminal detection and env variable handling with the sentinel in the following file:
https://github.com/microsoft/vscode-documentdb/blob/main/src/commands/launchShell/launchShell.ts

The same file with the name `launchShell.ts` in this repository follows a very similar structure. Update this file to match the terminal detection, env variable handling, and the sentinel as seen in the file provided by the link above.

once done, remember to run `npm run prettier-fix` and `npm run l10n`

PR #2692, opened 2025-05-16
Import Document with Buffer

### LINKED ISSUE #2582: Importing multiple documents is very slow, need bulk-import support
When importing multiple documents from file, we upload only one document at a time. This should be improved by uploading multiple documents with one single API call. For example Data Explorer uploads 100 documents with a single API call which is much faster.

Version: 0.25.0
OS: darwin
OS Release: 24.3.0
Product: Visual Studio Code
Product Version: 1.97.2
Language: en

xingfan-git:
June 10th Update: https://github.com/microsoft/vscode-cosmosdb/pull/2692#issuecomment-2959781683

---

This pull request update the insert document feature for Document DB. To improve importing performance, it uses buffer to reduce the number of calls to server, and update error message handling accordingly

- [x] Implement the core architecture
- [x] Simplify the `ClusterDocumentBufferManager`: we don’t really need a manager here—a buffer that handles buffering and size measurement should be sufficient.  
- [x] Create the buffer when needed during the import—this would behave the same as the current `ClusterDocumentBufferManager` management.  
- [x] Once we have that buffer, can you make it independent of the `Document` class? Or make it generic?  
- [x] The buffer should accept some configuration options when being created—provide default configurations for `Mongo` and for `CosmosDB`.  
- [x] Use this improved import implementation for `CosmosDB Core` as well, so that `Azure Databases` benefits from the improvement too.  
- [ ] Ship it 🚀

Fixes #2582


tnaum-ms (commented):
Thank you @xingfan-git !

It looks good. It doesn't compile due to a minor error, please make sure to review the feedback in the action log
(note to self: modify actions to write errors back to the PR as a comment somehow).

I'd like to finalize with a simplification of the code base and a release. I'll share a dedicated comment in the PR in a couple of minutes.

tnaum-ms:
## Summary and Closing Steps

Good approach, and great catch 🕵️ with the correct configuration of the `insertMany` command! I'm very happy to see the improvement and can’t wait to ship it 🚀

Now, in order to get there, I’d like to finalize this ticket with:

- [ ] Simplify the `ClusterDocumentBufferManager`: I think that we don’t really need a manager here, a buffer that handles buffering and size measurement should be sufficient.  
- [ ] Create the buffer when needed during the import—this would behave the same as the current `ClusterDocumentBufferManager` management.  
- [ ] Once we have that buffer, can you make it independent of the `Document` class? Or make it generic?  
- [ ] The buffer should accept some configuration options when being created—provide default configurations for `Mongo` and for `CosmosDB`.  
- [ ] Use this improved import implementation for `CosmosDB Core` as well, so that `Azure Databases` benefits from the improvement too.  
- [ ] Ship it 🚀

I'll add the tasks to the PR as well for better tracking.

xingfan-git:
The error message for Cosmos Core was slightly changed: 
we won't track the position of failed document in bulk insertion, so the error message would only say 'insertion failed with error code xxx' but not specify which file failed to be inserted

src/commands/importDocuments/importDocuments.ts
```diff
@@ -282,11 +303,15 @@ async function parseAndValidateFileForCosmosDB(
...
     document: unknown,
-): Promise<{ document: unknown; error: string }> {
+    buffer: ClusterBufferManager | undefined,
+): Promise<{ document: unknown; error: string } | { count: number; errorOccured: boolean }> {
     try {
         if (node instanceof CollectionItem) {
             // await needs to catch the error here, otherwise it will be thrown to the caller
-            return await insertDocumentIntoCluster(node, document as Document);
+            if (!buffer) {
+                return { count: 0, errorOccured: true };
```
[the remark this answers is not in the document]
xingfan-git: updated

src/documentdb/ClustersClient.ts
```diff
@@ -451,13 +452,43 @@ export class ClustersClient {
...
+        } catch (error) {
+            // print error messages to the console
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if (error.writeErrors) {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                for (const writeError of error.writeErrors) {
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                    ext.outputChannel.appendLog(`Write error: ${writeError.errmsg}`);
+                    ext.outputChannel.show();
+                }
```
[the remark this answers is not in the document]
xingfan-git: updated, thanks

src/commands/importDocuments/importDocuments.ts
```diff
@@ -114,41 +116,47 @@ export async function importDocumentsWithProgress(
             }
 
             const countDocuments = documents.length;
-            const incrementDocuments = 50 / (countDocuments || 1);
+            const incrementDocuments = 100 / (countDocuments || 1);
```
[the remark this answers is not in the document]
xingfan-git: where do I add a final 50% increment?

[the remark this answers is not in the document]
tnaum-ms: I see why you did it, the increment around URI loading was broken, I updated it so that now everything is reported correctly. I reduced the loading's share to 25% and import has now 75% so that it's easier to see the progress.

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,196 @@
...
+     * Create a new document buffer
+     *
+     * @param options Configuration options for the buffer
+     */
+    constructor(private readonly options: DocumentBufferOptions) {}
+
+    /**
+     * Calculate the size of a document using the provided size calculation function
+     */
+    public getSize(document?: T): number {
```
tnaum-ms, line 75: Please rename it to getDocumentSize

tnaum-ms (changes requested):
First of all, congrats ⭐

We're at this stage, imports are so much faster!

https://github.com/user-attachments/assets/6b72cd65-76e6-4c77-96c9-7508c07556bd

I added comments in the PR, essentially an few minor things and essentially an ask to simplify the Buffer API even further. I get the idea behind the `auto flush` you've build and I know it helps with high-throughput systems. We're in a much simpler environment, with no multithreading in the end, so I'd put the emphasis on keeping the code simpler for future maintainers and external contributors.

Let's just check whether a buffer is "full" (i.e. flush is required), and then do it. If you want to keep the auto-flush behavior, please make sure function names are more verbose, so that the intent is clear + ensure that we don't only rely on `success === false` to decide about the next step but do add some sort of status flag.

**A:** Simplify the API (no auto-flush),
**B:** Improve the API (if you want to keep auto-flush).

I'd suggest moving forward with **A**, but it's up to you :)

---

These changes are API tweaks. I added a few changes to ensure progress reporting is correct (URI/file loading for only one file wasn't being reported correctly - this was an error in the original code).

Once these changes are in, feel free to move forward with a similar PR for our DocumentDB extension without waiting for my feedback here.

---

I just noticed that my comments where I just comment `Copilot`'s comments are not linked, please scroll up and read other 'unresolved' discussions and close them.

tnaum-ms (dismissed):
@xingfan-git 🥳 Congratulations on the first PR to be shipped!

- Everything looks great! I added more details to write error logging for better UX.

tnaum-ms:
@xingfan-git I bumped into a blocking issue :-/

During extensive testing, I encountered a **blocking issue** when attempting to import data into **Azure Cosmos DB for MongoDB (RU-based)**. The import fails due to **request unit (RU) throttling**, and the driver in use does not appear to handle RU-based flow control automatically.

In contrast, testing against **Azure Cosmos DB NoSQL** did _not_ show this issue — likely because the driver handles throttling internally or in a more robust way.

### Details

The following error is observed during import:

```
Write error: Failed with code "16500". - Error=16500, Details='Insert error.'
```


This indicates that the server is rejecting operations due to exceeding available RUs.

### Root Cause

The MongoDB API driver we use does **not account for RU throttling**, as it's unaware of the Cosmos DB-specific RU model. Unlike Cosmos DB’s native NoSQL SDKs, it doesn’t manage retry or backoff logic by default.

---

### Required Fix Before Shipping

To avoid failed imports and improve resilience, we should implement RU throttling handling explicitly:

- [ ] Add **delay + retry** logic when encountering `16500` errors  
  (RUs are replenished every second, so short delays may suffice)
  
- [ ] Optionally, investigate whether **Cosmos DB for MongoDB (RU)** exposes a **server status command or metric** we can query to assess RU availability in advance

xingfan-git:
The issue occurred because the bulk insert operation exceeded the collection RU limit.
We cannot directly retry the insert operation:

- First, RU has a retry mechanism, but it is not robust enough to resolve the issue we encountered;
- Secondly, the bulk insert operation partially succeeded, so if we simply retry, we risk 1) inserting duplicate items or 2) facing insertion failure if the _id field was specified in the inserted items.

Due to the complexity of this issue, we decided to fix it in two steps:

- For the current iteration, we noticed that inserting documents one by one could resolve most situations where throttling occurs, so we decided to use a small buffer (single document buffer) for RU resources.
- In the future, there is a feature called [server-side retry for Mongo RU](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/prevent-rate-limiting-errors) that could resolve the throttling issue. We can investigate if we can perform a server-side retry via driver parameters. If not, @tnaum-ms mentioned we can retry only on the failed documents in a later iteration.

tnaum-ms:
@sevoku This looks good from my point of view. It improves performance for vCore and Cosmos DB, but there is no improvement for RU.

With RU, we encountered issues with throughput and will need to address this in a separate ticket. As a workaround, we reduced the buffer size to "1," essentially reverting to the insertion of individual documents. This is slow enough for most cases, so throughput limits won't be hit.

This is a temporary solution for RU (or rather, no solution), but we'll address it properly during the copy-and-paste work. There is already a dedicated issue for it. We also have a ticket for better reporting of the `id` of failing documents, but I would prefer to integrate this with the copy-and-paste ticket.

---

I tested this with vCore and RU for various configurations and error scenarios. It's being merged into DocumentDB. For Cosmos DB, I conducted tests, including partition key conflicts, but I might have missed something.

🎯 Please try it out in your Cosmos DB setups.

src/commands/importDocuments/importDocuments.ts
```diff
@@ -4,19 +4,31 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { type ItemDefinition, type JSONObject, type JSONValue, type PartitionKeyDefinition } from '@azure/cosmos';
-import { parseError, type IActionContext } from '@microsoft/vscode-azext-utils';
+import { nonNullProp, parseError, type IActionContext } from '@microsoft/vscode-azext-utils';
```
bk201-, line 7: Please use nonNullProp from our utils, our function provides information about properties in the message

src/commands/importDocuments/importDocuments.ts
```diff
@@ -87,18 +99,21 @@ export async function importDocumentsWithProgress(
             progress.report({ increment: 0, message: l10n.t('Loading documents…') });
 
             const countUri = uris.length;
-            const incrementUri = 50 / (countUri || 1);
+            const incrementUri = 25 / (countUri || 1);
```
bk201-, line 102: Please either add comment or create a constant variables for 25 and 75 (25% and 25% * 3 of 100% progress)

src/commands/importDocuments/importDocuments.ts
```diff
@@ -87,18 +99,21 @@ export async function importDocumentsWithProgress(
...
             const documents: unknown[] = [];
             let hasErrors = false;
 
-            for (let i = 0, percent = 0; i < countUri; i++, percent += incrementUri) {
+            for (let i = 0; i < countUri; i++) {
+                const increment = (i + 1) * incrementUri;
```
bk201-, line 107: Taking into consideration the code below where also there is increment, I can't figure out what approach is right. Here you increase increment every time and it looks like you **set** the percentage. But in the code below the increment never increases and it looks like you **add** it value to progress. 

**updated**
I figured out why it does not increases. Because you really don't know how many documents you already inserted and how many in the buffer.

In this case the progress bar message will be untransparent for user. 
1. The progress bar doesn't move but the message shows how many documents were inserted
2. Even if the message shows that 20 document were inserted it is absolutely does not mean that they were actually inserted. Again for user it is not unclear.
3. You move buffer logic to the insert function, but the logic has to be reversed. You have to insert into a buffer and when it return the error that it is full, you have to flush and insert one batch. 
4. Progress bar might be easily counted. One step: 75 / the number of document. And when you insert a batch, you take a number of inserted document multiply them to value of one step and add to progress bar. 
5. In this case you also remove weird check if buffer has document since you will know this. If the cycle for is end you just do flush and insert, always.

src/commands/importDocuments/importDocuments.ts
```diff
@@ -282,55 +314,137 @@ async function parseAndValidateFileForCosmosDB(
...
     document: unknown,
-): Promise<{ document: unknown; error: string }> {
+    buffer: DocumentBuffer<unknown> | undefined,
+): Promise<{ count: number; errorOccurred: boolean }> {
     try {
-        if (node instanceof CollectionItem) {
-            // await needs to catch the error here, otherwise it will be thrown to the caller
-            return await insertDocumentIntoCluster(node, document as Document);
+        // Check for valid buffer
+        if (!buffer) {
```
bk201-, line 321: If without buffer you return the fixed object, why do not you set type as DocumentBuffer<unknown> only

```typescript
async function insertDocument(
    node: CosmosDBContainerResourceItem | CollectionItem,
    document: unknown,
    buffer: DocumentBuffer<unknown>, // <-- it is more strict type
): Promise<{ count: number; errorOccurred: boolean }> {
```

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+    maxSingleDocumentSizeBytes: 2 * 1024 * 1024, // 2MB
+    calculateDocumentSize: (document: unknown) => {
+        return document ? Buffer.byteLength(EJSON.stringify(document)) : 0;
+    },
+};
+
+/**
+ * Create a document buffer configured for MongoDB
+ */
+export function createMongoDbBuffer<T>(customConfig?: Partial<DocumentBufferOptions>): DocumentBuffer<T> {
```
bk201-, line 258: Function name already has name Mongo so you can narrow a type

```typescript
// This buffer can keep only Document and all inherited classes
export function createMongoDbBuffer<T extends Document>(customConfig?: Partial<DocumentBufferOptions>): DocumentBuffer<T>
```

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+    return new DocumentBuffer<T>({
+        ...defaultMongoBufferConfig,
+        ...customConfig,
+    });
+}
+
+/**
+ * Create a document buffer configured for Cosmos DB
+ */
+export function createCosmosDbBuffer<T>(customConfig?: Partial<DocumentBufferOptions>): DocumentBuffer<T> {
```
bk201-, line 268: As well as above. 
```typescript
export function createCosmosDbBuffer<T extends ItemDefinition>(customConfig?: Partial<DocumentBufferOptions>): DocumentBuffer<T>
```

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+
+// Default configuration for Cosmos DB buffers
+const defaultCosmosBufferConfig: DocumentBufferOptions = {
+    maxBufferSizeBytes: 4 * 1024 * 1024, // 4MB
+    maxDocumentCount: 100,
+    maxSingleDocumentSizeBytes: 2 * 1024 * 1024, // 2MB
+    calculateDocumentSize: (document: unknown) => {
+        return document ? Buffer.byteLength(EJSON.stringify(document)) : 0;
+    },
+};
```
bk201-, line 253: Please add `as const`. 
The object without `as const` might be modified, since only reference is constant. 

`as const` freezes object and does not allow change it.

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+const defaultMongoBufferConfig: DocumentBufferOptions = {
+    maxBufferSizeBytes: 32 * 1024 * 1024, // 32MB
+    maxDocumentCount: 50,
+    maxSingleDocumentSizeBytes: 16 * 1024 * 1024, // 16MB
+    calculateDocumentSize: (document: unknown) => {
+        // Use EJSON to calculate the size of MongoDB documents
+        // Adding 20% for BSON overhead compared to JSON
+        return document ? Buffer.byteLength(EJSON.stringify(document)) * 1.2 : 0;
+    },
+};
```
bk201-, line 243: Please add as const.
The object without as const might be modified, since only reference is constant.

as const freezes object and does not allow change it.

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+                return {
+                    ...insertResult,
+                    documentsToProcess: [],
+                };
+        }
+    }
+
+    public insert(document: T): BufferInsertResult {
+        // Check if the document is valid
+        if (!document) {
```
bk201-, line 180: But if I create a new buffer like `new DocumentBuffer<undefined>` this condition will be wrong. Please see comment above about T

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+     * the contents of the buffer if it's full and needs to be flushed
+     */
+    documentsToProcess?: T[];
+}
+
+/**
+ * Document buffer for a specific database/collection pair.
+ * Used for batching document inserts to improve performance.
+ */
+export class DocumentBuffer<T> {
```
bk201-, line 93: It is wrong generic definition. 
1. class name has name `Document` so the type has to be narrowed to `T extends Document | ItemDefinition | <any document type>`. 
2. Just T might lead to wrong behavior and redundant checks. See the comment below.

src/utils/documentBuffer.ts
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Function to calculate the size of a document
+     */
+    calculateDocumentSize: (document: unknown) => number;
+}
+
+/**
+ * Error codes for document buffer operations
+ */
+export enum BufferErrorCode {
```
bk201-, line 36: NIT: Please try to avoid string enum. You can find more information in the internet. Just for example
https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

src/commands/importDocuments/importDocuments.ts
```diff
@@ -108,47 +123,64 @@ export async function importDocumentsWithProgress(
...
+            const incrementDocuments = 75 / (countDocuments || 1);
             let count = 0;
+            let buffer: DocumentBuffer<unknown> | undefined;
+            if (selectedItem instanceof CollectionItem) {
+                const hosts = getHostsFromConnectionString(nonNullProp(selectedItem.cluster, 'connectionString'));
+                const isRuResource = hasDomainSuffix(AzureDomains.RU, ...hosts);
+
+                if (isRuResource) {
+                    // For Azure MongoDB RU, we use a buffer with maxDocumentCount = 1
+                    buffer = createMongoDbBuffer<unknown>({
```
bk201-, line 141: But now you know a type moreover the function name already says what type it must be. See my comment below

```typescript
buffer = createMongoDbBuffer<Document>
```

src/commands/importDocuments/importDocuments.ts
```diff
@@ -282,55 +314,137 @@ async function parseAndValidateFileForCosmosDB(
...
+    } catch {
+        return { count: 0, errorOccurred: true };
+    }
+}
+
+async function insertDocumentWithBufferIntoCluster(
+    node: CollectionItem,
+    buffer: DocumentBuffer<unknown>,
+    document?: Document,
+    // If document is undefined, it means that we are flushing the buffer
```
bk201-, line 343: Looks like hack and unexpected behavior. In order to use function everyone HAS TO read this comment. This code smells.

src/utils/documentBuffer.ts
tnaum-ms: @bk201- never disappoints, thank you for sharing! I can always learn something new (to me) thanks to you.

src/commands/importDocuments/importDocuments.ts
tnaum-ms: Good point. It's something I discussed "offline" with @xingfan-git
Let's leave it as is in this it iteration - there is another thing we're working on here and I wanted to finally get it started:
https://github.com/microsoft/vscode-documentdb/issues/63

Once it finalizes, we'll be moving import and export to the new task service and, while working on it, we'll improve the overall code and comment quality around import and export.

This work will be packaged as a module for sharing.

PR #2757, opened 2025-09-17
feat: Implement sorting and filtering tree items (#2726)

bk201-:


src/tree/mixins/Filterable.ts
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        this.filterProperty = property;
+        this.filterPattern = pattern;
+
+        // Show feedback
+        if (this.filterPattern) {
+            vscode.window.showInformationMessage(`Filter applied: ${this.filterPattern}`);
+        } else {
+            vscode.window.showInformationMessage('Filter cleared');
```
languy: Does this need to be translated?

bk201-: Fixed

PR #2758, opened 2025-09-22
feat: Deploy LLM instructions files

languy:
This change allow the extension to deploy LLM instructions files to the user's profile folder.
On windows, it is the `C:\Users\<username>\AppData\Roaming\Code\User\prompts` folder.
This is where the vscode `Chat: new instructions file...` command creates the instructions.
There is also a vscode `Chat: configure instructions` to edit/delete the files.

Add two new commands:
- Deploy instructions files
- Remove instructions files

Upon activation, the `.md` files in the `resources/llm-assets` folder are copied to the user's profile.
The deployed filenames are stored in the global state (`llm.assets.manifest` key), so that the extension always knows what to remove or update in subsequent releases.

### DIFF
--- a/src/cosmosdb/commands/deployLLMInstructionsFiles.test.ts
+++ b/src/cosmosdb/commands/deployLLMInstructionsFiles.test.ts
@@ -55,13 +55,6 @@ jest.mock('@vscode/l10n', () => ({
     t: jest.fn(),
 }));
 
-jest.mock('vscode', () => ({
-    window: {
-        showInformationMessage: jest.fn(),
-        showErrorMessage: jest.fn(),
-    },
-}));
-
 import * as l10n from '@vscode/l10n';
 import * as crypto from 'crypto';
 import * as fs from 'fs';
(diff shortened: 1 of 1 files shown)


src/cosmosdb/registerCosmosDBCommands.ts
```diff
@@ -26,4 +27,6 @@ export function registerCosmosDBCommands(): void {
     registerCommand('cosmosDB.executeNoSqlQuery', executeNoSqlQuery);
     registerCommand('cosmosDB.getNoSqlQueryPlan', getNoSqlQueryPlan);
     // #endregion
+
+    registerCommand('cosmosDB.ai.deployInstructionFiles', deployLLMInstructionsFiles);
```
bk201-: Please move it to the command folder, this file will be deleted as soon as we understand that no one uses the scrapbook

languy: Moved it to `registerCommands.ts` where all the commands are registered.

PR #2835, opened 2025-11-18
Improved item tooltips

### LINKED ISSUE #2798: Improve Items Tooltips
Currently we show this tooltip when hovering an Item inside the Items node:

Image

I don't think that most of this information is useful, especially that it can not be copied and just looking at it does not provide much value.
What is really useful in this scenario though is the ID (especially if the extension is configured to show a different field) and the partition keys. Optionally we could list the first level values (if there are not too many, we could limit it maybe).

So what I'd expect especially from DevEng perspective:
* **ID: {id}**
* _{Partiton Key 1}: {value}_
* _{Partiton Key 2}: {value}_
* _{Partiton Key x}: {value}_
* other values

Note the formatting and order, ID first **bold**, followed by PKs _italic_ (special values such as `null`/`undefined` could be greyed out if formatting allows), followed by all other item values (limit to some sane number, 10 or so?)

mkrueger:
Improved the item tooltip layout.

This pull request refactors how tooltips are generated for Cosmos DB items in the tree view. The main change is replacing the previous verbose, multi-section tooltip logic with a new compact, Markdown-formatted tooltip that presents key information more concisely and clearly. The new tooltip combines item ID, partition key(s), and selected metadata fields, and introduces improved formatting and value handling.

image

**Tooltip generation improvements:**

* Replaced the previous `generateDocumentTooltip` and `generatePartitionKeyTooltip` methods with a new `generateCompactTooltip` method that produces a single, compact Markdown tooltip combining ID, partition key(s), and metadata fields.
* The new tooltip displays the item ID, supports both single and multiple partition key paths with formatted values, and presents system metadata fields (`_rid`, `_etag`, `_ts`) in a simple Markdown table.
* Added helper methods `getPartitionKeyValuesArray` and `formatValue` to extract and format partition key values, including truncating long strings/objects and handling null/undefined values gracefully.
* Updated the tooltip assignment in the tree item to use `generateCompactTooltip` instead of combining separate document and partition key tooltips.
Fixes #2798


sevoku:
Some notes:
* `Item Value` header is misaligned (left compared to other headers) probably because it's a table
* Do we need headers at all? I proposed to remove them in #2798 to save space
   * This would also allow us to use just one single markdown table for all data
 * Why do we need the system metadata items? The only useful one is rid that could be used for REST calls, but it's still not useful in that context (we could have a "Copy RID" context menu item)
 * Either all items should have a colon or none, right now ID and PKs have one, but items don't.
 * Also note my formatting proposal in #2798

mkrueger:
it formats like that if there are more pk - which aren't in my example. 

IDK what meta data is good for but I thought that fells under "other values". 

I would've removed them - it's just internal meta data but what does "other values" mean then? The tooltip format is markdown so advanced styling is difficult.

src/tree/cosmosdb/CosmosDBItemResourceItem.ts
```diff
@@ -60,33 +58,85 @@ export abstract class CosmosDBItemResourceItem
...
+                }
+                return '`' + json + '`';
+            } catch {
+                return '`[object]`';
+            }
+        }
+        if (typeof value === 'string') {
+            // Truncate long strings
+            if (value.length > 100) {
+                return `"${value.substring(0, 97)}…"`;
```
bk201-: Here you don't use 3 symbols for ellipsis, but only one.

src/tree/cosmosdb/CosmosDBItemResourceItem.ts
```diff
@@ -60,33 +58,85 @@ export abstract class CosmosDBItemResourceItem
...
+            }
+            return `"${value}"`;
```
bk201-: I don't remember how it works in JS, but I'm 95% sure that here the user will see "value", but in fields above it will be \`value\`

bk201- (dismissed):

src/tree/cosmosdb/CosmosDBItemResourceItem.ts
sevoku: @bk201- we should always prefer using the real ellipsis character `…` rather than 3 dots `...`. On Windows they look the same, but not on macOS, Apple HIG requires it in fact. See https://gist.github.com/vancura/27fad9793840ceabae14 for good typography coding rules :)

bk201-: I meant that substring has to be till 99

sevoku: lol, sorry, I totally missed that, sigh

Issue #2446, opened 2024-11-20
Control Plane operations are not supported with RBAC


sevoku:
Control Plane operations are not supported with the data plane SDK and we need to switch to control plane when using RBAC: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/security/how-to-grant-control-plane-role-based-access?tabs=built-in-definition%2Ccsharp&pivots=azure-interface-portal

**Repro steps:**

1. Navigate to a NoSQL account with `localAuth` disabled or after enabling the `azureDatabases.useCosmosOAuth` setting
2. Select the "Create Database..." command from the accounts context menu or run `cosmosDB.createDocDBDatabase` from the command palette.
3. Give the database a name and hit Enter

**Expected:** database gets created

**Actual:** Error is logged (see below).

Action: cosmosDB.createDocDBDatabase
Error type: 403
Error Message: Request blocked by Auth redacted:username-test-nosql : The given request [POST /dbs] cannot be authorized by AAD redacted:key data plane. Learn more: redacted:url
ActivityId: redacted:id, redacted:url

Version: 0.24.1-alpha
OS: darwin
OS Release: 24.1.0
Product: Visual Studio Code - Insiders
Product Version: 1.96.0-insider
Language: en

<details>
<summary>Call Stack</summary>

```
httpRequest @azure/RequestHandler.js:95:1
@azure/retryUtility.js:44:1
addDignosticChild @azure/diagnostics.js:44:1
addDignosticChild @azure/diagnostics.js:44:1
ClientContext.create @azure/ClientContext.js:236:1
Databases.createInternal @azure/Databases.js:90:1
withDiagnostics @azure/diagnostics.js:100:1
Proxy.createChildImpl DocDBAccountTreeItemBase.ts:99:44
```

</details>

PR #2856, opened 2025-12-02
feat: Allow using only Cosmos DB Account Reader Role

JMonsorno:
- Check for PrimaryReadonlyMasterKey using `readonlykeys` action if `listkeys` action returns 403 forbidden

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,26 @@
 
 ## 0.31.0
 
+### Added
+
+- Added "Create Database" and "Create Container" nodes to NoSQL accounts and Databases (#2837).
+- Added Database/Container selection to the Query Editor toolbar and updated the Query Editor Icon (#2701).
+- Added missing filter and sort buttons to Items, Procedures, Triggers and attached Accounts (#2855).
+- Enabled Emulator support on Apple Silicon (darwin-arm64) (#2825).
+- Updated localization for 13 languages (#2860).
+
 ### Fixed
 
+- Fixed an issue where pasting text into the Query Editor was broken (#2843).
+- Fixed an issue where creating containers failed with default settings for serverless accounts (#1931).
 - Fixed an issue where the Extension Settings were missing in the Extensions Settings section (#2826).
+- Updated several dependencies to address security issues.
+
+### Changed
+
+- Updated to React 19 (#2760).
+- Improved Tooltips for single items and added tooltips showing partition keys for containers (#2835).
+- Renamed the "Create Database Server" wizard and removed MongoDB and PostgreSQL support, please use the specialized extensions instead (#2863).
 
 ## 0.30.0
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-cosmosdb",
-  "version": "0.31.0",
+  "version": "0.31.1",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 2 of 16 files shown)


src/cosmosdb/CosmosDBCredential.ts
```diff
@@ -148,28 +145,65 @@ async function getKeyCredentialWithARM(
...
-            context.telemetry.properties.receivedKeyCreds = 'true';
-        } else {
+        if (localAuthDisabled) {
             throw new Error(l10n.t('Local auth is disabled'));
         }
-    } catch {
-        context.telemetry.properties.receivedKeyCreds = 'false';
-        logLocalAuthDisabledWarning(accountName);
+        const key: string | undefined =
+            (await getPrimaryMasterKeyWithARM(client, resourceGroup, accountName)) ||
```
sevoku: Looks great, but maybe we should unify these two functions into one with a `readOnlyKey: bool` param to reduce code duplication.

sevoku (changes requested):
Thanks for submitting this improvement! See my comment below. This reminds me of https://github.com/microsoft/vscode-cosmosdb/issues/2446, maybe we should have a readOnly flag for accounts which disables all write operations with a warning, but that's a different topic. I'll file an issue for that.

src/cosmosdb/CosmosDBCredential.ts
JMonsorno: I was torn originally on doing that as well as making the read only one not have exception/error handling so it'd just pass up and landed somewhere in between. It less code duplication but a hair harder to read IMO with the negative case first, not sure how you feel about something that's an abstraction away from read-only to make the positive use case first like `writeEnabled: boolean`

src/cosmosdb/CosmosDBCredential.ts
```diff
@@ -148,28 +145,54 @@ async function getKeyCredentialWithARM(
...
+        const key: string | undefined =
+            (await getPrimaryKeyWithARM(client, resourceGroup, accountName, false)) ||
+            (await getPrimaryKeyWithARM(client, resourceGroup, accountName, true));
+        if (key) {
+            keyCred = {
+                type: AuthenticationMethod.accountKey,
+                key: key,
+            };
+        }
+    } finally {
```
sevoku, line 161: I just realised this will break the automatic auth selection flow if getting the keys fails with any error other than 403 (when getPrimaryKeyWithARM throws). The initial code caught all exceptions and always showed the local auth disabled warning, so that the caller could move on and skip key auth. Now all exceptions propagate and auth will fail even with AuthenticationMethod.auto.

Also... there is a flaw in old code, we don't log the initial error. Maybe logLocalAuthDisabledWarning should have a `reason` param which would be logged in case of unexpected errors.

src/cosmosdb/CosmosDBCredential.ts
sevoku: I don't have any strong feelings about the one or the other, at the end it's just a helper func :)

src/cosmosdb/CosmosDBCredential.ts
JMonsorno: I think I got the error handling but not comfortable with the internationalization requirements to add a new warning.

sevoku: Ok thanks!

bk201- (approved):

PR #2968, opened 2026-03-27
feat: Add Monaco autocompletion for CosmosDB NoSQL query language

bk201-:
- Register custom nosql Monaco language with Monarch tokenizer for CosmosDB NoSQL-specific syntax highlighting (keywords, built-in functions, operators, comments, strings, quoted identifiers)

- Implement context-aware CompletionItemProvider with: SQL keyword completions, 80+ built-in function completions with signatures, schema-driven property suggestions after dot notation, nested property resolution, bracket notation for special property names, and lightweight FROM-clause alias tracking

- Pipe container schema from extension host to webview: add containerSchema field to QueryEditorState, send schema on ready/generation/merge/wipe, listen for schemaUpdated event in QueryEditorContextProvider

- Enable fixedOverflowWidgets so autocomplete popup floats above Allotment split panes


sevoku (commented):
First Pass: looks great!

Couple of thoughts:
* "Delete current schema" should maybe have a confirmation? Or maybe let's call it "Remove current schema" which sounds less dangerous?
   * Do we need this feature at all? Why would one want to remove it at this point anyways?
* "Generate schema based on queries" opens Settings, can't we turn it into a Checkmark/box? We could keep the setting to be the default value for freshly opened query editors.

sevoku (approved):

PR #2998, opened 2026-04-24
Improve AI query advisor

languy:
Improvements to the AI assisted query advisor:
* Disable non-copilot models for now: add a centralized flag to enable/disable.
* New telemetry flags when closing QueryGenerateComponent (prompt was entered, query was generated)
* Fix schema sample allow/deny source field in telemetry
* Disable explain query when no text in query editor
* Explain query uses selected query in the editor if applicable.
* Add support for inline query in chat participant "explainQuery" command (fixes #2969)
  * For example "@cosmosdb /explainQuery select * from c where c._ts > 1234567890". The intent-based works as well: "@cosmosdb explain this query select * from c where c._ts > 1234567890"). 
  * With an inline query, don't require an open query editor/connection
  * For "explainQuery", the logic to pick which query to explain now prioritizes: first inline query if specified, otherwise selected query if exists, otherwise entire query editor content.
  * The "Query: <the query being explained>" section in the output always clarifies which query is being explained.
* Remove redundant logic to select which query to use
* Remove unused `ExplainQueryButton` component
* Move the default "select * from c" query out of query editor webview so we can modify the default for update query avoid flickering (ie show "select * from c" briefly, then update with updated query)
* Translated outputChannel info messages and some remaining error messages and AI stream output.


src/panels/trpc/routers/queryEditorRouter.ts
```diff
@@ -642,24 +649,30 @@ export const queryEditorRouterDef = queryEditorRouter({
...
-        ctx.state.pendingConfirmResolve?.(false);
-        ctx.state.pendingConfirmResolve = undefined;
-        ctx.state.generateQueryCancellation?.cancel();
-        ctx.state.generateQueryCancellation?.dispose();
-        ctx.state.generateQueryCancellation = undefined;
-    }),
+    closeGenerateInput: queryEditorProcedure
+        .input(z.object({ hadEnteredPrompt: z.boolean(), hadExecutedGenerateQuery: z.boolean() }).optional())
+        .mutation(async ({ input, ctx }) => {
+            ext.outputChannel.info('[Generate Query] Generate query input closed by user.');
```
bk201-: l10n?

src/panels/trpc/appRouter.ts
```diff
@@ -93,10 +94,8 @@ export type DocumentRouterContext = BaseRouterContext & {
  * Telemetry middleware is already baked into each procedure, so individual
  * `.use(trpcToTelemetry)` calls are not needed here.
  */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+
```
bk201-: Redundant new line

languy: Fixed

src/panels/trpc/routers/queryEditorRouter.ts
languy: Yes. Fixed other non-translated instances.

bk201- (dismissed):

bk201- (dismissed):

bk201- (approved):

PR #3016, opened 2026-05-06
fix(rbac): Refactor Cosmos DB Control Plane Implementation

### LINKED ISSUE #2990: Creating database fails with native RBAC
To create databases we use the [Cosmos DB SDK directly](https://github.com/microsoft/vscode-cosmosdb/blob/main/src/commands/createDatabase/CosmosDBExecuteStep.ts#L26), which works fine with account keys and old accounts using [native RBAC](https://aka.ms/cosmos-native-rbac) and a data plane role with "Microsoft.DocumentDB/databaseAccounts/sqlDatabases/*" data action.

New accounts however don't allow any database operations using the data plane SDK, all database operations must go through ARM and the principal needs to have at least the `Cosmos DB Operator` role assigned to it.

We need to rewrite all database related commands to use ARM instead of CosmosClient. Also need to check the same for containers.

sevoku:
## Summary

Refactors Azure Cosmos DB (NoSQL) **control-plane** operations behind a new `CosmosDBControlPlane` abstraction with two implementations: ARM (preferred for Azure-signed-in accounts) and the Cosmos DB SDK (`CosmosClient`, used for the local emulator and workspace-attached connection-string accounts). Selecting ARM for Azure-signed-in accounts fixes resource creation/deletion failing on accounts configured with strict native data-plane RBAC.

Fixes #2990.

## Why

Cosmos DB accounts that disable data-plane writes (native data-plane RBAC) reject control-plane operations issued via the `CosmosClient` SDK (create database, create container, delete operations, throughput reads). The previous `withClaimsChallengeHandling` retry loop could not recover from these because the operations are not authorized regardless of token claims. Routing them through ARM resolves the failure for Azure-signed-in accounts.

### Control-plane abstraction

- **New** `CosmosDBControlPlane` interface defining a unified surface for **databases, containers, and throughput**, plus a `ThroughputResource` DTO so callers do not need to know which implementation produced the value.
  - **Error contract is explicit**: all methods throw on failure (`ErrorResponse` from `@azure/cosmos` or `RestError` from `@azure/core-rest-pipeline`); successful return implies completion. Translating those errors into user-visible notifications remains the job of the surrounding `callWithTelemetryAndErrorHandling` wrapper used by command handlers.
  - **Stored procedures, triggers, and UDFs are intentionally not part of this abstraction**: they are data-plane resources and the standard data-plane RBAC role grants their CRUD. Routing them through ARM blurred the control/data-plane boundary without functional benefit. Their call sites continue to use the data-plane `CosmosClient` directly via `withClaimsChallengeHandling`.
- **New** `ArmCosmosDBControlPlane` implementation using `@azure/arm-cosmosdb`.
- **New** `CosmosDBSdkControlPlane` (renamed from the original `DataPlaneCosmosDBControlPlane` per review feedback — the name now describes the transport rather than overloading "data plane"). Wraps `withClaimsChallengeHandling` over `CosmosClient`. Its constructor accepts either `AccountInfo` or `NoSqlQueryConnection`.
- **New** `getControlPlane(accountInfo)` selector: returns the ARM implementation when `!isEmulator && subscription && resourceGroup`, otherwise the SDK implementation.
- **New** `getControlPlaneForConnection(connection)` selector: same logic against `NoSqlQueryConnection` (also requires `accountName`); used by the query editor.

### Refactored callsites (now go through the control plane)

- Tree loaders: `CosmosDBAccountResourceItem`, `CosmosDBDatabaseResourceItem`.
- Resource create wizards: `createDatabase/CosmosDBExecuteStep`, `createContainer/CosmosDBExecuteStep`.
- Resource delete commands: `deleteDatabase`, `deleteContainer` (the redundant `success`-flag dance was removed; we now rely on the documented throws-on-failure contract).
- Throughput viewing: `viewOffer`.
- Query editor connection picker: `queryEditorRouter.getConnections` now uses `getControlPlaneForConnection().listDatabases()` + `listContainers()` instead of a direct `CosmosClient` call. Same partial-failure semantics and telemetry are preserved.

### Connection / account context

- **`AccountInfo`** extended with optional `subscription?: AzureSubscription` and `resourceGroup?: string`. Populated for Azure-signed-in accounts only.
- **`NoSqlQueryConnection`** extended with optional `accountName?: string`, `subscription?: AzureSubscription`, `resourceGroup?: string`. Populated by `createNoSqlQueryConnection` and the duplicate `getConnectionFromNode` factory in `openNoSqlQueryEditor.ts` from the source `AccountInfo`. Existing consumers are unaffected because all three fields are optional.

### Bug fixes

- **Partition key kind selection** in both `ArmCosmosDBControlPlane.createContainer` and `CosmosDBSdkControlPlane.createContainer`. The expression `(definition.partitionKey?.kind ?? partitionKeyPaths.length > 1) ? MultiHash : Hash` would short-circuit on a truthy `"Hash"` string and incorrectly produce `MultiHash` whenever `kind` was explicitly set to `Hash`. Replaced with an explicit `=== PartitionKeyKind.MultiHash` comparison that only falls back to the path-count heuristic when `kind` is undefined.
- **`minimumThroughput` parsing** now preserves a value of `0` (previously coerced to `undefined`).

### Tests

- New `getControlPlane.test.ts` covering the selection matrix for both `getControlPlane` and `getControlPlaneForConnection`: Azure-signed-in, emulator, workspace-attached, and the partial-Azure-context fallback cases (11 tests total).

## Behavior

| Account kind | Before | After |
| --- | --- | --- |
| Azure-signed-in, standard | SDK | ARM |
| Azure-signed-in, native data-plane RBAC | SDK (failed) | ARM (works) |
| Local emulator | SDK | SDK |
| Workspace-attached (connection string) | SDK | SDK |

## Known limitation (documented in code)

Workspace-attached connection-string accounts that point at a real Azure Cosmos DB account configured with strict native data-plane RBAC will still fall into the SDK branch and have control-plane operations rejected, because the connection string does not carry Azure subscription or resource-group context. Resolving this requires either prompting the user to associate the attached account with an Azure subscription or extending the connection-string import flow to capture that context. A `TODO` is left on `getControlPlane` describing the gap.

## Out of scope (intentionally deferred)

- `resolveConnectionState` in `queryEditorRouter` continues to read the container definition for partition key via `withClaimsChallengeHandling`. Partition key is consumed by data-plane query execution and the existing path already handles native RBAC.
- Schema generation (`generateSchema`) is a SQL query and stays on the data plane.

## Validation

- `npm run build` — passes.
- `npx vitest run src/cosmosdb/controlPlane` — 11/11 pass.
- `npm run prettier-fix` and `npm run lint` — clean.
- Manual smoke testing on real Azure-signed-in (standard + native RBAC), emulator, and workspace-attached accounts is recommended before merge.

### DIFF
--- a/src/commands/deleteContainer/deleteContainer.ts
+++ b/src/commands/deleteContainer/deleteContainer.ts
@@ -53,11 +53,8 @@ export async function cosmosDBDeleteContainer(
     }
 
     try {
-        const success = await deleteContainer(node);
-
-        if (success) {
-            showConfirmationAsInSettings(successMessage);
-        }
+        await deleteContainer(node);
+        showConfirmationAsInSettings(successMessage);
     } finally {
         const lastSlashIndex = node.id.lastIndexOf('/');
         let parentId = node.id;
@@ -68,13 +65,9 @@ export async function cosmosDBDeleteContainer(
     }
 }
 
-async function deleteContainer(node: CosmosDBContainerResourceItem): Promise<boolean> {
-    let success = false;
+async function deleteContainer(node: CosmosDBContainerResourceItem): Promise<void> {
     await ext.state.showDeleting(node.id, async () => {
         const controlPlane = getControlPlane(node.model.accountInfo);
         await controlPlane.deleteContainer(node.model.database.id, node.model.container.id);
-        success = true;
     });
-
-    return success;
 }
--- a/src/commands/deleteContainer/deleteContainer.ts
+++ b/src/commands/deleteContainer/deleteContainer.ts
@@ -7,7 +7,7 @@ import { type IActionContext } from '@microsoft/vscode-azext-utils';
 import { AzExtResourceType } from '@microsoft/vscode-azureresources-api';
 import * as l10n from '@vscode/l10n';
 import { API } from '../../AzureDBExperiences';
-import { withClaimsChallengeHandling } from '../../cosmosdb/withClaimsChallengeHandling';
+import { getControlPlane } from '../../cosmosdb/controlPlane';
 import { ext } from '../../extensionVariables';
 import { type CosmosDBContainerResourceItem } from '../../tree/cosmosdb/CosmosDBContainerResourceItem';
 import { getConfirmationAsInSettings } from '../../utils/dialogs/getConfirmation';
@@ -71,13 +71,9 @@ export async function cosmosDBDeleteContainer(
 async function deleteContainer(node: CosmosDBContainerResourceItem): Promise<boolean> {
     let success = false;
     await ext.state.showDeleting(node.id, async () => {
-        await withClaimsChallengeHandling(node.model.accountInfo, async (cosmosClient) => {
-            const response = await cosmosClient
-                .database(node.model.database.id)
-                .container(node.model.container.id)
-                .delete();
-            success = response.statusCode === 204;
-        });
+        const controlPlane = getControlPlane(node.model.accountInfo);
+        await controlPlane.deleteContainer(node.model.database.id, node.model.container.id);
+        success = true;
     });
 
     return success;
--- a/src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
+++ b/src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
@@ -432,12 +432,20 @@ function toThroughputResource(response: ThroughputSettingsGetResults): Throughpu
     return {
         throughput: resource.throughput,
         autoscaleMaxThroughput: resource.autoscaleSettings?.maxThroughput,
-        minimumThroughput:
-            typeof resource.minimumThroughput === 'string'
-                ? Number.parseInt(resource.minimumThroughput, 10) || undefined
-                : (resource.minimumThroughput as number | undefined),
+        minimumThroughput: parseMinimumThroughput(resource.minimumThroughput),
         raw: response,
     };
+}
+
+function parseMinimumThroughput(value: unknown): number | undefined {
+    if (typeof value === 'number') {
+        return value;
+    }
+    if (typeof value === 'string') {
+        const parsed = Number.parseInt(value, 10);
+        return Number.isNaN(parsed) ? undefined : parsed;
+    }
+    return undefined;
 }
 
 function isNotFound(err: unknown): boolean {
--- a/src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
+++ b/src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
@@ -8,26 +8,13 @@ import {
     type SqlContainerCreateUpdateParameters,
     type SqlContainerGetResults,
     type SqlDatabaseGetResults,
-    type SqlStoredProcedureGetResults,
-    type SqlTriggerGetResults,
     type ThroughputSettingsGetResults,
 } from '@azure/arm-cosmosdb';
-import {
-    PartitionKeyDefinitionVersion,
-    PartitionKeyKind,
-    type ContainerDefinition,
-    type StoredProcedureDefinition,
-    type TriggerDefinition,
-} from '@azure/cosmos';
+import { PartitionKeyDefinitionVersion, PartitionKeyKind, type ContainerDefinition } from '@azure/cosmos';
 import { callWithTelemetryAndErrorHandling, type IActionContext } from '@microsoft/vscode-azext-utils';
 import { type AzureSubscription } from '@microsoft/vscode-azureresources-api';
 import * as l10n from '@vscode/l10n';
-import {
-    type ContainerResource,
-    type DatabaseResource,
-    type StoredProcedureResource,
-    type TriggerResource,
-} from '../../tree/cosmosdb/models/CosmosDBTypes';
+import { type ContainerResource, type DatabaseResource } from '../../tree/cosmosdb/models/CosmosDBTypes';
 import { createCosmosDBManagementClient } from '../../utils/azureClients';
 import { type CosmosDBControlPlane, type ThroughputResource } from './CosmosDBControlPlane';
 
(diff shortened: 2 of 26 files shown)


src/cosmosdb/controlPlane/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ *   with native data-plane RBAC, where data-plane control operations are
+ *   rejected by the service (see issue #2990).
+ * - The local emulator and workspace-attached connection-string accounts use
+ *   the data-plane `CosmosClient` because ARM is not reachable for them.
+ */
+export function getControlPlane(accountInfo: AccountInfo): CosmosDBControlPlane {
+    if (!accountInfo.isEmulator && accountInfo.subscription && accountInfo.resourceGroup) {
+        return new ArmCosmosDBControlPlane(accountInfo.subscription, accountInfo.resourceGroup, accountInfo.name);
+    }
+    return new DataPlaneCosmosDBControlPlane(accountInfo);
```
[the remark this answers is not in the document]
sevoku: done: [cae528b](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/cae528b4c25a44531fc8eb24d8e89a1cf246751f)

src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
```diff
@@ -0,0 +1,448 @@
...
+        definition: ContainerDefinition,
+        throughput?: number,
+    ): Promise<ContainerResource> {
+        const client = await this.getArmClient();
+        const containerId = definition.id!;
+        const partitionKeyPaths = definition.partitionKey?.paths ?? [];
+        const kind =
+            (definition.partitionKey?.kind ?? partitionKeyPaths.length > 1)
+                ? PartitionKeyKind.MultiHash
+                : PartitionKeyKind.Hash;
```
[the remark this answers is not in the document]
sevoku: fixed: [e184221](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/e184221964294c92f9653e150282e349efae5d3f)

src/cosmosdb/controlPlane/DataPlaneCosmosDBControlPlane.ts
```diff
@@ -0,0 +1,246 @@
...
+        const options: RequestOptions = {};
+        if (throughput && throughput !== 0) {
+            options.offerThroughput = throughput;
+        }
+
+        const partitionKeyPaths = definition.partitionKey?.paths ?? [];
+        const partitionKeyDefinition = {
+            paths: partitionKeyPaths,
+            kind:
+                (definition.partitionKey?.kind ?? partitionKeyPaths.length > 1)
```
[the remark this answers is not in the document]
sevoku: Fixed: [e184221](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/e184221964294c92f9653e150282e349efae5d3f)

src/cosmosdb/controlPlane/ArmCosmosDBControlPlane.ts
```diff
@@ -0,0 +1,449 @@
...
+    if (!resource) {
+        return { raw: response };
+    }
+    return {
+        throughput: resource.throughput,
+        autoscaleMaxThroughput: resource.autoscaleSettings?.maxThroughput,
+        minimumThroughput:
+            typeof resource.minimumThroughput === 'string'
+                ? Number.parseInt(resource.minimumThroughput, 10) || undefined
+                : (resource.minimumThroughput as number | undefined),
```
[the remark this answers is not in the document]
sevoku: Even though 0 is unlikely here, still valid, fixed [82804e5](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/82804e55c2b56f0e65fe94d17cd467cc3a7dd0ad)

mkrueger (dismissed):

src/cosmosdb/controlPlane/DataPlaneCosmosDBControlPlane.ts
```diff
@@ -0,0 +1,258 @@
...
+import { type NoSqlQueryConnection } from '../NoSqlQueryConnection';
+import { withClaimsChallengeHandling } from '../withClaimsChallengeHandling';
+import { type CosmosDBControlPlane, type ThroughputResource } from './CosmosDBControlPlane';
+
+/**
+ * Control-plane implementation that routes every operation through the
+ * data-plane `CosmosClient`. Used for the local emulator and for
+ * workspace-attached connection-string accounts, where ARM is not reachable.
+ */
+export class DataPlaneCosmosDBControlPlane implements CosmosDBControlPlane {
```
bk201-: Class name is confusing, since it does not related with Data Plane.

I would like to call them based on purpose or privileges: Arm/SDK, RBAC/without, High/Low privileges.

src/commands/deleteContainer/deleteContainer.ts
```diff
@@ -71,13 +71,9 @@ export async function cosmosDBDeleteContainer(
...
     await ext.state.showDeleting(node.id, async () => {
-        await withClaimsChallengeHandling(node.model.accountInfo, async (cosmosClient) => {
-            const response = await cosmosClient
-                .database(node.model.database.id)
-                .container(node.model.container.id)
-                .delete();
-            success = response.statusCode === 204;
-        });
+        const controlPlane = getControlPlane(node.model.accountInfo);
+        await controlPlane.deleteContainer(node.model.database.id, node.model.container.id);
```
bk201-, line 71: Even if container wasn't deleted we will get success == true, since now one in control plane implementation does not check response.

src/cosmosdb/controlPlane/CosmosDBControlPlane.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Abstraction over Azure Cosmos DB control-plane operations for the SQL
+ * (NoSQL) API. Implementations either route through the ARM management plane
+ * (preferred for Azure-signed-in accounts) or through the data-plane
+ * `CosmosClient` (used for the local emulator and workspace-attached
+ * connection-string accounts where ARM is not available).
+ */
+export interface CosmosDBControlPlane {
```
bk201-, line 46: Using this class means that all error handling should be implemented inside it, because this class returns values or void but user loses the additional info - response headers and statuses. It will be OK it all methods detected not valid state, but in implementation classes I don't see this.

src/commands/deleteContainer/deleteContainer.ts
sevoku: Yeah, it's not really consistent. The SDK call would throw in most cases anyway, so the statusCode check is mostly superflous. But won't hurt to surface unexpected errors that are not caught by SDK and don't indicate success. Fixed with 74ed0cb5.

src/cosmosdb/controlPlane/DataPlaneCosmosDBControlPlane.ts
sevoku: Yup, renamed to CosmosDBSdkControlPlane in [4b71b38](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/4b71b3897c2fbda6d290e6eb07003bea99d2712a)

src/cosmosdb/controlPlane/CosmosDBControlPlane.ts
sevoku: SDK calls are supposed to throw anyway and besides the deleteContainer/deleteDatabase calls no other call site was checking the statusCode or anything else, the errors were just propagated, hence there is no real need for this unless we need to treat Cosmos/ARM errors/exceptions differently.

sevoku: In fact based on SDK code analysis not even the 204 statusCode is needed. The SDK will throw on any other code. The only difference is even worse... we checked the result and the only thing we did with it is showing a notification if it was successful ([throwing away any other status code that was not throwing!](https://github.com/microsoft/vscode-cosmosdb/blob/44617d61545879c1afb087f6d990dab8e9cfccd4/src/commands/deleteContainer/deleteContainer.ts#L58-L60)). Added a comment and removed the checks in fc3ffd31.

src/commands/deleteContainer/deleteContainer.ts
sevoku: Actually that was even worse, see https://github.com/microsoft/vscode-cosmosdb/pull/3016#discussion_r3221676497. Removed the statusCode checks altogether.

src/cosmosdb/controlPlane/CosmosDBControlPlane.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Abstraction over Azure Cosmos DB control-plane operations for the SQL
+ * (NoSQL) API: databases, containers, and throughput. Implementations either
+ * route through the ARM management plane (preferred for Azure-signed-in
+ * accounts) or through the data-plane `CosmosClient` (used for the local
+ * emulator and workspace-attached connection-string accounts where ARM is
+ * not available).
+ *
+ * Stored procedures, triggers, and user-defined functions are intentionally
+ * not included here: they are data-plane resources and callers should access
+ * them through the data-plane `CosmosClient` (see `withClaimsChallengeHandling`).
```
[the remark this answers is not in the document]
sevoku: updated description.

src/cosmosdb/controlPlane/index.ts
```diff
@@ -0,0 +1,51 @@
...
+ * rejected by the service. Resolving this requires either prompting the user
+ * to associate the attached account with an Azure subscription or extending
+ * the connection-string import flow to capture that context.
+ */
+export function getControlPlane(accountInfo: AccountInfo): CosmosDBControlPlane {
+    if (!accountInfo.isEmulator && accountInfo.subscription && accountInfo.resourceGroup) {
+        return new ArmCosmosDBControlPlane(accountInfo.subscription, accountInfo.resourceGroup, accountInfo.name);
+    }
+    return new CosmosDBSdkControlPlane(accountInfo);
+}
```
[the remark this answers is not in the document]
sevoku: This is a rare operation and caching would be an overkill here. Once we have fill control plane support in SDKs with latest native RBAC we'll need ARM even less. Not worth addressing at the moment.

src/cosmosdb/fs/TriggerFileDescriptor.ts
```diff
@@ -48,16 +48,24 @@ export class TriggerFileDescriptor implements EditableFileSystemItem {
...
+        const existing = await withClaimsChallengeHandling(this.model.accountInfo, async (client) => {
+            try {
+                const response = await client
+                    .database(this.model.database.id)
+                    .container(this.model.container.id)
+                    .scripts.trigger(this.model.trigger.id)
+                    .read();
+                return response.resource;
+            } catch (err) {
+                if (isNotFound(err)) {
```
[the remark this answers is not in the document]
sevoku: good point, fixed [7abfe58](https://github.com/microsoft/vscode-cosmosdb/pull/3016/commits/7abfe5893cefde38b6e816fa8ce37f9481dc1862)

bk201- (approved):

PR #3022, opened 2026-05-06
refactor: Formating multiple markdown files

bk201-:


sevoku (approved):
Most of the files were already touched by assistants, should we maybe have an additional instruction in https://github.com/microsoft/vscode-cosmosdb/blob/main/.github/copilot-instructions.md to apply some of these rules?

PR #3042, opened 2026-05-19
Don't publish Cosmos DB Shell MCP server when prerequisites are missing

mkrueger:
## Problem

Whenever Copilot (or any other MCP consumer) loaded, users saw notifications prompting them to install Cosmos DB Shell or to enable `cosmosDB.shell.MCP.enabled`, even though they had never opted into Cosmos DB MCP.

Root cause: `provideMcpServerDefinitions` always published the Cosmos DB Shell MCP server definition. VS Code/Copilot then invoked `resolveMcpServerDefinition` during background tool discovery, which displayed `showMcpSettingsNotification(...)` in the "not installed" and "MCP disabled" branches.

## Fix

- **`provideMcpServerDefinitions`**: return `[]` when prerequisites aren't met (`cosmosDB.shell.MCP.enabled` is off, or `isCosmosDBShellInstalled()` is false). The existing `didChangeEmitter` already re-fires when those settings or `cosmosDB.shell.path` change, so the server is automatically republished once the user actually opts in.
- **`resolveMcpServer`**: demote the "not installed" and "MCP disabled" branches to `ext.outputChannel.appendLine(...)` and keep the `throw` so the MCP runtime reports a clean failure. This acts as a safety net for cached definitions; no user-facing notification fires for either branch anymore.
- Port-conflict and "shell already running without MCP" notifications are **unchanged** — those are still actionable user-facing conditions tied to deliberate user action.
- Rename the misleading `isCosmosDBShellSupportEnabled()` to `isCosmosDBShellInstalled()` (it only checks whether the shell binary is installed; the previous name suggested a feature flag).

## Result

Users who haven't installed Cosmos DB Shell or haven't enabled the MCP setting no longer see Cosmos DB install prompts triggered by Copilot's background MCP discovery. Users who do enable the setting / install the shell get the server published automatically via the existing `didChangeEmitter` wiring.

(Supersedes #3041, which was auto-closed when its head branch was renamed.)


sevoku (approved):

mkrueger:
Superseded by #3043 (recreated from the microsoft/vscode-cosmosdb branch instead of the fork).

PR #3063, opened 2026-05-21
feat(query-editor): Priority Level support

### LINKED ISSUE #3007: Enable Priority-based execution for queries run from the VS Code extension when the account feature is enabled
When users run Azure Cosmos DB queries from the VS Code extension, query requests should honor Priority-based execution if the Cosmos account has the feature enabled.

Today, extension-executed queries may not explicitly set request priority, which can prevent customers from using High/Low priority behavior to protect critical workloads during RU saturation.

**Background:**
Azure Cosmos DB supports Priority-based execution, where Low priority requests are throttled before High priority requests once provisioned throughput is fully consumed. Customers enabling this feature expect all query entry points, including developer tools, to respect priority settings.

**Reference:**
[https://learn.microsoft.com/en-us/azure/cosmos-db/priority-based-execution?tabs=net-v3](vscode-file://vscode-app/c:/Users/sasinnat/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

**Problem statement:**

Queries executed via the VS Code extension do not currently provide a clear way to apply request priority.
Customers using the extension for operational or troubleshooting queries cannot align with their workload priority strategy.
This can lead to unexpected throttling behavior and reduced predictability under load.
Expected behavior:

If Priority-based execution is enabled on the customer account, queries from the extension should send priority on request options.

- Users should be able to choose query priority (for example: High or Low), with a sensible default.
- If the account does not support the feature, the extension should fail gracefully or fall back without breaking query execution.
- Proposed implementation:

- Add a priority selector for query execution in the extension UI/command flow.
- Pass the selected priority in query request options/header for all extension query paths.
- Gate behavior based on feature availability/capability detection where possible.
- Add logging/telemetry for priority selection and fallback outcomes.

**Acceptance criteria:**

Query executions from the extension include priority when selected.
Priority behavior is verified against an account with the feature enabled.
No regression for accounts without the feature.
Documentation/help text in the extension explains priority behavior and prerequisites.

**Why this matters:**
This brings extension behavior in line with Cosmos DB service capabilities and gives customers predictable query throttling behavior that matches production workload priorities.

bk201-:
### 1. Query Editor — Priority Level support (main feature)

Adds Cosmos DB **Priority-based execution** support to the Query Editor.

-   New `Priority Level` submenu inside the Run button, with `High` / `Low` radio options.
-   Submenu is shown only when the connected account has priority-based execution enabled at the ARM resource level (`isPriorityLevelEnabled` capability) and disabled for the local emulator.
-   The selected level is persisted per account; on first open the picker is seeded from the previously-saved value, with the existing in-session choice preserved against late-arriving capability fetches.
-   The header is wired through tRPC (`queryEditorRouter.runQuery`) into the SDK's request options, and intentionally **not** sent for emulator connections (where the header is meaningless).

Touched files:

-   `src/panels/trpc/routers/queryEditorRouter.ts`
-   `src/webviews/cosmosdb/QueryEditor/QueryPanel/RunQueryButton.tsx`
-   `src/webviews/cosmosdb/QueryEditor/state/QueryEditorContextProvider.tsx`

Also fixed a pre-existing UI bug in the Run-button menu: when both `Throughput Bucket` and `Priority Level` submenus were visible, two `<MenuDivider />`s were rendered back-to-back. Now a single divider is shown when **at least one** of the config submenus is visible.

### 2. Webview cold-open performance

Opening a webview panel in **dev mode** previously took ~5 s before content appeared. A DevTools trace pinned the cost to two issues:

-   ~1.6 s — Vite dev server cold-transforming ~230 source modules on demand as separate HTTP requests (waterfall).
-   ~2.4 s — V8 evaluating a single 16.7 MB pre-bundled dep chunk that merged `@griffel/core`, `stylis`, `rtl-css-js`, `@emotion/hash`, Fluent UI runtime, etc.

Fixes applied:

-   **`src/webviews/WebviewRegistry.ts`** — entries are now `() => import(...)` factories. Each view (`Document`, `MigrationAssistant`, `QueryEditor`) is dynamically imported, so opening one panel no longer pays the cost of loading the other two. In prod each view becomes its own Rollup chunk — e.g. opening the Document panel no longer downloads Monaco.
-   **`src/webviews/index.tsx`** — `render()` awaits the dynamic import before mounting React.
-   **`vite.config.views.mjs`** — added `server.warmup` for `src/webviews/**` and `packages/*/src/**` so the dev server pre-transforms files at startup instead of on first request.
-   **`vite.config.views.mjs`** — added `optimizeDeps.include` for `@griffel/core`, `@griffel/react`, `stylis`, `rtl-css-js`, `@emotion/hash`, `@fluentui/react-components`, `@fluentui/react-icons` so Rolldown emits a separate dep pre-bundle entry for each instead of merging them into one monolithic shared chunk that V8 has to parse sequentially on the main thread.

### 3. Initial loader spinner

`src/panels/BaseTab.ts` — added a CSS-only loader (spinner + localized "Loading…" label) rendered inside `#root` directly in the HTML template. It appears instantly when the panel opens — before any module is fetched or React hydrates — and is automatically replaced when `createRoot().render()` mounts the React tree.

Accessibility:

-   `role="status"` + `aria-live="polite"` on the container so screen readers announce it.
-   `aria-hidden="true"` on the rotating ring (decorative).
-   Uses VS Code theme variables (`--vscode-foreground`, `--vscode-progressBar-background`, `--vscode-font-family`, …) so it matches the active theme.
-   Honors `prefers-reduced-motion`.

### 4. Maintenance

`npm run lint-fix` autofix pass on 5 unrelated files (dropped redundant `return await` and unused `eslint-disable` directives), followed by `npm run prettier-fix`.

Fixes: #3007


src/cosmosdb/CosmosDBCredential.ts
```diff
@@ -45,33 +46,38 @@ export function getCosmosDBEntraIdCredential(credentials: CosmosDBCredential[]):
...
+    // Basic account information, required for all credential types
     accountName: string;
     documentEndpoint: string;
     isEmulator: boolean;
 
     // ARM-specific parameters
-    armClient?: CosmosDBManagementClient;
-    resourceGroup?: string;
-    databaseAccount?: DatabaseAccountGetResults;
+    arm?: AzureResourceMetadata;
```
languy, line 55: nit: I would choose a more explicit name like `azureResourceMetadata` to avoid confusion with ARM (Azure resource manager)

src/webviews/cosmosdb/Migration/MigrationAssistant.tsx
```diff
@@ -1679,14 +1680,14 @@ function MigrationAssistantInner({ channel }: { channel: MigrationChannel }) {
...
                                             {state.assessmentProgress && (
-                                                <Text
-                                                    role="status"
-                                                    aria-live="polite"
-                                                    size={200}
-                                                    style={{ color: 'var(--vscode-descriptionForeground)' }}
-                                                >
-                                                    {state.assessmentProgress}
-                                                </Text>
+                                                <output aria-live="polite">
```
languy, line 1683: It's great to fix it, but next time, let's keep the change separate.

src/webviews/WebviewRegistry.ts
```diff
@@ -3,12 +3,19 @@
...
+ *   no longer downloads Monaco.
+ */
 export const WebviewRegistry = {
-    cosmosDbDocument: Document,
-    cosmosDbMigration: MigrationAssistant,
-    cosmosDbQuery: QueryEditor,
-} as const;
+    cosmosDbDocument: () => import('./cosmosdb/Document/Document').then((m) => m.Document),
+    cosmosDbMigration: () => import('./cosmosdb/Migration/MigrationAssistant').then((m) => m.MigrationAssistant),
+    cosmosDbQuery: () => import('./cosmosdb/QueryEditor/QueryEditor').then((m) => m.QueryEditor),
```
languy, line 20: Nice!

languy (approved):

src/cosmosdb/CosmosDBCredential.ts
bk201-: it was made intentionally :) I even called class to be similar with ARM

src/webviews/cosmosdb/Migration/MigrationAssistant.tsx
bk201-: Agree in this case a new version of oxlint complains at a11y rules (new version fixes lint phantom error, but adds also new rules). This code was created with old oxlint

PR #3130, opened 2026-06-05
test: migrate to Vitest + add Playwright e2e suite (VS Code + emulator)

bk201-:
This PR layers two test-infrastructure deliveries on the same branch.
The Vitest migration is the bigger structural change; the Playwright
e2e suite reuses the resulting test framework and adds a real
extension-driving layer on top.

---

# Part 1 — Migrate integration tests from Mocha to Vitest

Unifies the test framework: both unit and integration tests now use
`import { describe, it, expect } from 'vitest'`. The integration test
host invokes `@vitest/runner.startTests()` directly inside the VS Code
Extension Host, sidestepping the worker-pool / Vite-server machinery of
`startVitest()` (and the `vscode`-module-not-in-worker problem that
would have come with it).

## Architecture

- `test/index.ts` (~140 LOC): custom test entry. Implements a minimal
  `VitestRunner` (basically just `importFile`) and globs compiled
  `out/test/**/*.test.js`. `@vitest/runner` calls our `importFile`
  via dynamic `import()` in the same process, so the test code sees
  the real `vscode` module.
- `scripts/run-integration-tests.mjs` (~80 LOC): replaces
  `@vscode/test-cli`. Wraps `@vscode/test-electron` to download VS
  Code, install `ms-azuretools.vscode-azureresourcegroups`, then launch
  the Extension Host with our compiled entry.
- `tsconfig.test.json`: now emits NodeNext ESM. `pretest` writes
  `out/package.json` with `{type: 'module'}` so Node treats the
  compiled tree as ESM. Includes only `test/**/*.ts`: integration
  tests must depend only on the public extension API (`vscode`,
  registered commands, contributed configuration), not on `src/`.
- `test/extensionActivation.test.ts`: first real integration test —
  proves we are inside the real Extension Host (not the unit mock),
  activates the cosmosdb extension, and asserts a contributed command
  is registered.

Dependencies dropped: `mocha`, `@types/mocha`, `mocha-junit-reporter`,
`mocha-multi-reporters`, `eslint-plugin-mocha`, `@vscode/test-cli` and
the mocha override block in `overrides` (52 packages removed in total).

## Windows-specific note

`glob` returns absolute paths with canonical uppercase drive letters
(`C:\`) while VS Code can launch the extension host with a lowercase
drive letter (`c:\`). In ESM, modules are cached by URL, so the two
casings would load `@vitest/runner` twice and break collector state.
The runner globs with relative paths and resolves through
`here = path.dirname(fileURLToPath(import.meta.url))` to keep one
consistent casing.

---

# Part 2 — Playwright e2e suite against real VS Code + Cosmos DB emulator

End-to-end test infrastructure that launches a **real downloaded VS
Code** via `@vscode/test-electron` and drives the React webviews inside
it. Pairs with a dedicated Cosmos DB emulator (separate compose
project, ports **8082/1235**) so the suite never collides with the
developer's local emulator on 8081.

## Why a real VS Code

The webview runtime VS Code provides is **not just a browser** — it
includes `acquireVsCodeApi()`, a custom CSP, `--vscode-*` CSS variables,
`l10n_bundle` injection, the postMessage transport, and the Electron
version that ships with the editor. Faking all of that in plain
Chromium is brittle and never quite right; launching the real thing
avoids the entire class of "works in tests, breaks in VS Code" bugs.

## What's in `test/e2e/`

| Path | Purpose |
| --- | --- |
| `setup/globalSetup.ts` | Auto-rebuilds `dist/` when stale (mtime check on `package.json` + `src/**`), downloads VS Code, installs dependent extensions, brings up the emulator, seeds it. Skip flags: `COSMOSDB_E2E_SKIP_BUILD=1`, `COSMOSDB_E2E_SKIP_EMULATOR=1`. |
| `setup/globalTeardown.ts` | Tears the emulator down (only if globalSetup actually started it — tracked via an ownership marker file), removes the run-scoped temp dir. |
| `setup/emulator.ts` | `docker compose -p cosmosdb-e2e up/down`, readiness probe via `getDatabaseAccount()` (with `enableEndpointDiscovery: false` to pin the SDK to 8082 instead of the emulator-advertised `127.0.0.1:8081`), seed import wrapper. |
| `setup/activation.ts` | Pre-test handshake: reveals the Azure sidebar and waits for the "Cosmos DB Accounts" tree item — proves both Azure Resources **and** our extension finished activating before any spec runs. |
| `fixtures/vscode.ts` | Worker-scoped `vscodeApp` + `vscodeWindow` fixtures. Copies the workspace fixture, seeds `settings.json`, pushes emulator env vars into the launched VS Code, dismisses native dialogs. |
| `fixtures/webviewHelpers.ts` | `runCommand`, `getWebviewByPredicate`, `closeAllEditorTabs`. |
| `fixtures/webviews.ts` | `openMigrationAssistant`, `openQueryEditor`, `openDocument`, `attachEmulator`. |
| `fixtures/workspace/` | Source-of-truth workspace folder copied into each worker's temp dir (`sample.nosql`, `.vscode/settings.json`). |
| `specs/smoke.spec.ts` | Each of the three React webviews mounts. |
| `specs/emulator-connected.spec.ts` | `attachEmulator` → open Query Editor → run default `SELECT * FROM c` → assert `prod-00000` appears in the results grid. End-to-end proof: **workspace storage → connection �� CosmosClient → emulator → tRPC → grid**. |

## Test-only extension commands

Gated by `process.env.COSMOSDB_E2E_TEST === '1'` (set by the Playwright
fixture) and a matching `cosmosDB.e2eTestMode` context key on the
palette entries. Production users never see them — without the flag
the commands are not registered at all.

- `cosmosDB.e2e.openQueryEditor` / `cosmosDB.e2e.openDocument` — render
  the panel with a real `NoSqlQueryConnection` built from the emulator
  env vars (fallback to disconnected when env is absent).
- `cosmosDB.e2e.attachEmulator` — pushes an attached-account entry into
  workspace storage via `StorageService`. Uses the same hash id
  (`getEmulatorItemUniqueId`) the production wizard uses, so re-running
  is idempotent.

## Docker compose for e2e

`docker-compose.e2e.yml` — **no volumes**, every test run starts from a
pristine emulator. Project name `cosmosdb-e2e`, so
`docker compose -p cosmosdb-e2e down` only ever touches our container
even if the developer is iterating on the local emulator on 8081.

## CI workflow

`.github/workflows/e2e.yml` runs the suite on Linux:

1. `npm ci`
2. `npx playwright install-deps chromium` — installs the GTK/NSS shared
   libs the downloaded VS Code Electron build needs (no actual chromium
   browser, just the apt packages Electron shares with it).
3. `npm run vite-prod` — explicit build so failures surface separately
   from test failures. `COSMOSDB_E2E_SKIP_BUILD=1` is then set so
   `globalSetup` skips the freshness check.
4. `xvfb-run -a npm run e2e` — mandatory wrapper because Electron has
   no real headless mode and the runner has no display.

On failure, three artifacts are uploaded for post-mortem:
HTML report, raw traces / videos / `error-context.md`,
and `docker logs` of the emulator container.

Also: emoji prefixes on every job display name across `.github/workflows`.

---

## Validation

- `npm run build`: 3/3 tsc projects green
- `npm run vitest`: 1092 passed, 1 skipped
- `npm test`: 3 integration tests passed inside real VS Code
- `npm run e2e`: smoke + emulator-connected specs pass locally
- `npm run lint`: 0 errors
- `npm run l10n`: no diffs


bk201-:
Superseded by #3136 — same change, squashed into a single commit on a fresh branch. Closing this one because the Azure Pipelines `Build` required status got stuck queued on it.

Issue #3144, opened 2026-06-11
Streaming import of large JSON/JSONL files with buffered bulk upload


bk201-:
## Summary
Today the document import (`importDocuments.ts`) reads the **entire file into memory** (`fs.readFile` + full parse) and materializes all documents into a single array before inserting. This does not scale to large files (multi-GB → OOM). We want a **streaming import** that reads documents incrementally and uploads them in bulk via a buffer.
This builds on the idea introduced in #2692 (now superseded for the in-memory NoSQL path and **can be closed**). The buffering concept from that PR is valuable — but it belongs in a *streaming* scenario, not the current in-memory one.
## Background: JSONL / NDJSON
- **JSONL / NDJSON** (Newline-Delimited JSON): one complete JSON document per line:
  ```
  {"id":"1","name":"a"}
  {"id":"2","name":"b"}
  ```
- It is the de-facto standard for large data exports (mongoexport, BigQuery, logs, ML datasets) precisely because it streams: one line = one document, without loading the whole file.
- Cost overhead is negligible (1 `\n` per document).
- A single large JSON **array** (`[ {...}, {...} ]`) cannot be read line-by-line; it requires an incremental/streaming parser.
## Proposed approach
1. **Adopt a modern streaming JSON parser.** Evaluate `@streamparser/json` (TS-native, works in browser/webview too) and `stream-json` (`StreamArray`). Hand-rolling a brace-counting parser is error-prone (braces inside strings/escapes) — use a real tokenizer.
2. **Support JSONL/NDJSON as a first-class import format** (alongside the existing JSON array).
3. **Pipeline:** `fs.createReadStream` → streaming parser → **buffer** → `executeBulkOperations` (chunked at 100), with **backpressure** (pause reading while a bulk insert is in flight).
## Buffer responsibility (important design constraint)
The **buffer is purely a buffer** — it accumulates documents and reports its state (size/count). It does **NOT** perform the bulk insert. Insertion is **our** orchestration logic:
- We push documents into the buffer.
- We check whether the buffer is "full" and, if so, trigger a bulk insert.
- If a number of documents fail to insert (e.g. `429` throttling, or bulk operation unsupported on the emulator) and accumulate beyond a threshold, we **stop the upload**.
This keeps the buffer simple and reusable (the over-engineered auto-flush from #2692 is intentionally avoided).
## Future: Upload Manager (likely a separate task)
Ideally an **Upload Manager** that:
- shows live upload progress,
[...]

PR #3145, opened 2026-06-11
feat: bulk document import via executeBulkOperations

bk201-:
## Summary
Replaces the per-document `items.create` loop in `importDocuments` with chunked bulk insertion via **`executeBulkOperations`** (the modern, non-deprecated `@azure/cosmos` API — same pattern used in `DocumentSession.processBulkDeleteBatch`).
## Changes
- **Bulk insert** documents in chunks of 100 via `items.executeBulkOperations`.
- **Throttling-aware**: throttled (`429`) documents are retried up to 10 times, honoring `x-ms-retry-after-ms` with backoff.
- **Phase-2 fallback**: any document that fails the bulk path (incl. emulators that do not support `executeBulkOperations`) is retried one-by-one.
- **Auto-generates `id`** (UUID) for documents missing one.
- **Cancellable** progress (`cancellable: true` + `CancellationToken`).
- Correct partition key extraction via `extractPartitionKey`.
## Why not #2692
This supersedes #2692 for the CosmosDB NoSQL path and that PR **can be closed**:
- #2692 used the **deprecated** `items.bulk()` (limit 100, no NoSQL throttling handling).
- Its generic auto-flush buffer was over-engineered for the in-memory case (all documents are already materialized into a single array before insert, so simple chunking is enough).
- It was based on an outdated import architecture (`getCosmosClient` + Mongo branch), whereas import is now CosmosDB-only and uses `withClaimsChallengeHandling`.
## Follow-up
The buffering idea from #2692 is valuable for a **streaming** import scenario (large JSONL/NDJSON files that should not be loaded fully into memory). Tracked separately in #3144.


src/commands/importDocuments/importDocuments.ts
```diff
@@ -110,33 +119,13 @@ export async function importDocumentsWithProgress(
...
-                            error: result.error,
-                        }),
-                    );
-                    ext.outputChannel.show();
-                    hasErrors = true;
-                } else {
-                    count++;
-                }
+            if (bulkResult.errors.length > 0) {
+                bulkResult.errors.forEach((error) => ext.outputChannel.appendLog(error));
```
sevoku, line 126: shouldn't this be outputChannel.error()?

sevoku (approved):
We should also add some telemetry for this to at least measure success (excl. single document errors, and mask them), but that's probably out of scope here.

PR #3149, opened 2026-06-11
docs(nosql): document LOG/LOG10 zero-argument limitation (#310)

### LINKED ISSUE #310: No 'Open' action displays for a graph/document
Repro Steps:
1. Choose a Cosmos DB account and expand it.
2. Create a graph/document under one database node.
3. Right click the new graph/document and check the context menu.

Expect:
There is 'Open' action in the context menu.

Actual:
There is no 'Open' action in the context menu. --Please see the below attachment:
noopen

bk201-:
## Summary
Documents the `LOG()` / `LOG10()` zero-argument behaviour reported in #310.
`LOG(0)`, `LOG10(0)` (and any negative argument) evaluate to `-Infinity`, which cannot be represented in JSON (RFC 8259). Azure Cosmos DB therefore rejects the result with **HTTP 400, error code 4001** — confirmed on **both production and the emulator**.
## Why no diagnostic?
This is **data-dependent**: a query like `SELECT LOG(c.value) FROM c` is perfectly valid and only fails for rows where `c.value <= 0`. A static diagnostic would fire on every `LOG(...)` call and produce false positives, so we document the behaviour instead.
## Changes
- **Hover docs** (`src/docs/index.ts`): added a warning to `LOG` and `LOG10` explaining the `-Infinity` -> 4001 behaviour and a guard workaround (`WHERE c.value > 0` or `c.value > 0 ? LOG(c.value) : null`).
- **Fixtures** (`select-functions.ts`): reclassified `M-07` / `M-13` `knownLimitation` from emulator-only to a Cosmos DB language limitation.
- **test-suite.md**: moved `M-07` / `M-13` into a new *Cosmos DB language limitations (not emulator-specific)* section.
Closes #310


sevoku (dismissed):
Referenced the wrong issue in PR description?

sevoku (approved):

PR #3153, opened 2026-06-16
fix: remove unnecessary locales from learn.microsoft.com urls

languy:
Remove locale ('en-us') from these URL's, they are automatically added when user visits the page.
Fix one broken link from review.microsoft.com to learn.microsoft.com.


bk201- (approved):

sajeetharan (approved):
LGTM

sevoku (approved):
We should consider using ama.ms links, so we can change the urls easily if needed

PR #3198, opened 2026-07-10
Gate Throughput Bucket selector on preview registration

### LINKED ISSUE #3189: Throughput Bucket selector shown even when Throughput Buckets feature is not enabled on the account
### Describe the bug

In the Query Editor, the **Run → Throughput Bucket** submenu always shows all 5 buckets (Bucket 1–5) for any non-emulator Cosmos DB account. However, **Throughput Buckets** is a preview feature that must be explicitly registered/enabled at the account level (`Microsoft.DocumentDB` → *Throughput Buckets* preview feature).

When the feature is **not** enabled on the account, the extension still renders the full bucket selector and lets the user pick a bucket (e.g. "Bucket 2"). This is misleading — it implies throughput buckets are configured and functional when they are not, and selecting a bucket has no meaningful effect (the account has no bucket configuration behind it).

### Steps to reproduce

1. Use a Cosmos DB (NoSQL) account where the **Throughput Buckets** preview feature is **not** registered (default state — see the *Preview Features* blade under `Microsoft.DocumentDB`).
2. Open a container in the Query Editor.
3. Click the dropdown next to **Run** → **Throughput Bucket**.
4. Observe that Bucket 1–5 are all listed and selectable.

### Expected behavior

The **Throughput Bucket** option should only be shown when the *Throughput Buckets* feature is actually enabled on the account. If the feature is not registered at the account level, the submenu should be hidden entirely (same as the existing emulator behavior), so users are not misled into thinking buckets are configured.

### Actual behavior

All 5 buckets are shown by default for any non-emulator account, regardless of whether the account-level *Throughput Buckets* preview feature is enabled.

### Root cause / pointer

The gating currently only excludes the emulator and hard-codes 5 enabled buckets:

`src/panels/trpc/routers/queryEditorRouter.ts`
```ts
// hide the option entirely when the active connection points at an emulator.
const supportsThroughputBuckets = !!ctx.state.connection && !ctx.state.connection.isEmulator;
...
throughputBuckets: supportsThroughputBuckets ? [true, true, true, true, true] : undefined,
```

This should additionally check whether the *Throughput Buckets* feature is enabled on the account (e.g. by reading the account's registered preview feature / bucket configuration) before exposing the selector, rather than defaulting to all 5 buckets.

### Environment

- Extension: Azure Databases / Cosmos DB for VS Code
- Feature: Query Editor → Throughput Bucket selector

mkrueger:
Fixes #3189.

## Problem
The Throughput Bucket selector in the Query Editor was shown for every non-emulator account, even when the Throughput Buckets preview is not registered on the subscription. Throughput Buckets is a subscription-scoped preview, so the account model alone doesn't indicate whether the feature is usable — resulting in a control that has no effect.

## Fix
- Query the subscription feature registration for `Microsoft.DocumentDB/ThroughputBuckets` and only show the selector when the feature state is `Registered`.
- Fail closed (hide the selector) for:
  - Emulator connections
  - Connections without Azure subscription metadata
  - ARM unavailability or authorization errors
- Default the webview state to hidden until capability detection resolves, avoiding a flash of the selector on startup.

## Implementation notes
- Added `@azure/arm-features` and a `createFeatureClient` helper (masks the subscription ID).
- Split the registration check into a pure predicate (`throughputBucketsFeature.ts`) so it can be unit-tested without VS Code host dependencies.
- Added 7 unit tests covering registered / case-insensitive / inactive states / unrelated features.

## Files
- `src/cosmosdb/throughputBuckets.ts` — connection-aware capability detection
- `src/cosmosdb/throughputBucketsFeature.ts` — pure registration predicate
- `src/cosmosdb/throughputBuckets.test.ts` — unit tests
- `src/panels/trpc/routers/queryEditorRouter.ts` — gate wired into Query Editor init
- `src/utils/azureClients.ts` — `createFeatureClient`
- `src/webviews/cosmosdb/QueryEditor/state/QueryEditorState.tsx` — hidden default state
- `package.json` / `package-lock.json` — new dependency

## Validation
- Focused unit tests: 7/7 passing
- ESLint + Oxlint on changed files: clean
- Editor diagnostics on changed files: none


bk201- (dismissed):

bk201- (approved):

bk201-:
@microsoft-github-policy-service rerun

PR #3269, opened 2026-08-17
Fix document editor ETag concurrency handling

### LINKED ISSUE #3263: Document editor can silently overwrite concurrent changes without an ETag precondition
## Type
Bug / data integrity / possible regression

## Extension version
Azure Cosmos DB for VS Code 0.36.0

## Description
Saving an existing item from the Document Editor does not send the loaded item `_etag` as an `If-Match` access condition. A stale editor tab can therefore replace a document that another client changed after the tab was opened, silently discarding the newer changes.

This appears related to or potentially a regression from #1287, which records that document mutations previously encountered ETag precondition failures and discusses prompting the user to overwrite, cancel, or compare.

## Reproduction
Use a disposable container with partition key `/pk` and two clients: the VS Code extension and Azure Portal Data Explorer.

1. Create this item:

```json
{
  "id": "concurrency-proof",
  "pk": "test",
  "version": 1,
  "portalChange": "original",
  "vscodeChange": "original"
}
```

2. Open the item with **Edit Item** in VS Code and leave that tab open. It now holds the original document and its initial `_etag`.
3. In Azure Portal Data Explorer, update and save:

```json
"version": 2,
"portalChange": "saved from portal"
```

4. Without refreshing the stale VS Code tab, change only:

```json
"vscodeChange": "saved from stale tab"
```

5. Save from VS Code.
6. Re-read the item or run:

```sql
SELECT
    c.id,
    c.version,
    c.portalChange,
    c.vscodeChange,
    c._etag
FROM c
WHERE c.id = "concurrency-proof"
```

## Actual behavior
The stale VS Code save is accepted. Values from the stale tab replace the newer values written through the portal, so `version` returns to `1` and `portalChange` returns to `original` while `vscodeChange` is saved.

No concurrency warning or HTTP 412 conflict is shown.

## Expected behavior
The extension should condition replacement on the `_etag` loaded into the editor. If the item changed in the meantime, Cosmos DB should return HTTP 412 and the extension should preserve the newer server version while offering appropriate choices such as refresh, compare, or explicitly overwrite.

## Source evidence
In `src/cosmosdb/session/DocumentSession.ts`, `replaceDocument` currently calls:

```ts
container.item(documentId.id!, documentId.partitionKey).replace(document, {
    abortSignal: signal,
    priorityLevel,
});
```

The request options contain no ETag-based `If-Match` access condition. The item `_etag` present in the loaded document is not used to guard the replacement.

## Suggested direction
- Retain the ETag corresponding to the version loaded in the editor.
- Pass it as an `If-Match` access condition during `replace`.
- Handle HTTP 412 by preserving both versions and offering refresh/compare/explicit-overwrite actions.
- Add an integration test with two writes where the stale write must not succeed by default.

jibin7jose:
Fixes #3263

### Summary
* Add `If-Match` precondition handling to document editor saves using the loaded `_etag`.
* Detect Cosmos DB `412` precondition failures and surface a conflict flow instead of silently overwriting.
* Add **Refresh**, **Compare**, and **Overwrite** actions for concurrent edit conflicts.
* Keep the existing `_etag` guard coverage with unit tests.

### DIFF
--- a/src/cosmosdb/session/DocumentSession.test.ts
+++ b/src/cosmosdb/session/DocumentSession.test.ts
@@ -3,15 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { type ItemDefinition } from '@azure/cosmos';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { type NoSqlQueryConnection } from '../NoSqlQueryConnection';
-import {
-    buildIfMatchRequestOptions,
-    deleteDocument,
-    isPreconditionFailedError,
-    replaceDocument,
-} from './DocumentSession';
 
 const cosmosMocks = vi.hoisted(() => ({
     delete: vi.fn(),
@@ -42,20 +35,7 @@ vi.mock('../withClaimsChallengeHandling', () => ({
     ),
 }));
 
-vi.mock('vscode', () => ({
-    window: {},
-    workspace: {},
-    commands: {},
-    Uri: { parse: vi.fn(), file: vi.fn() },
-}));
-
-vi.mock('@microsoft/vscode-azureresources-api', () => ({
-    AzExtResourceType: { AzureCosmosDb: 'AzureCosmosDb' },
-}));
-
-vi.mock('@microsoft/vscode-azext-azureauth', () => ({
-    getSessionFromVSCode: vi.fn(),
-}));
+import { deleteDocument, replaceDocument } from './DocumentSession';
 
 const connection = {
     containerId: 'container',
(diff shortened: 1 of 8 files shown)


bk201- (dismissed):

jibin7jose:
Thanks for the heads-up. I’ll fix the conflicts and check the build failure.

sevoku:
Hi @jibin7jose, thank you for working on this and for the thorough implementation and tests.

I merged the core concurrency fix through [#3266](https://github.com/microsoft/vscode-cosmosdb/pull/3266), since it overlapped with the partition-key move data-loss flow addressed there.

The merged implementation stores the loaded ETag as trusted extension-host state, keeps Overwrite conditional against the latest server version, and protects partition-key moves through create-before-delete and recoverable cleanup.

The Compare feature remains unique to this PR and would still be valuable. Could you update this branch from main and narrow the PR to adding Compare on top of the merged conflict handling?

PR #3342, opened 2026-09-04
Fix large JSON query result freezes

### LINKED ISSUE #3192: Crash/freeze when switching from Table View to JSON View with ~100 large documents
## Summary
When a query returns ~100 items, switching from **Table View** to **JSON View** can sometimes cause VS Code to crash. This is especially noticeable when each Cosmos DB document is relatively large, and appears to be caused by memory or rendering overload while building the JSON view.

## Repro
1. Run a query that returns ~100 relatively large documents.
2. View the results in **Table View**.
3. Switch to **JSON View**.
4. VS Code becomes unresponsive / crashes.

## Expected
Switching between Table View and JSON View should remain stable regardless of result size or document size.

## Suggestions
- Virtualize / paginate the JSON rendering so only visible content is materialized.
- Lazy-render large documents and/or cap the initial rendered payload with a "load more" affordance.
- Investigate memory growth when re-rendering the full result set on view switch.

## Environment
- Extension: vscode-cosmosdb
- API: Cosmos DB for NoSQL
- OS: Windows

bk201-:
## Summary
- warn before opening JSON query results larger than 5 MiB
- offer a 512 KiB preview with an accessible tooltip
- stream JSON into Monaco in bounded chunks and disable expensive editor features
- avoid retaining a complete JSON string in result view state
- add an optional binary payload mode to the large test-data generator
- gracefully degradation for Monaco services (disable: minimap, curve parenthesis coloring, partially syntax highlighting, etc)

Warning message bar
image

Only preview
image

Gradually loading
image

The end of data
image

After full loading of 200MiB the memory size around 900Mb, but due to chunked loading the UI is responsible
image

Fixes #3192


languy (dismissed):

src/webviews/cosmosdb/QueryEditor/ResultPanel/ResultTabViewJson.tsx
```diff
@@ -3,21 +3,320 @@
...
+    loading: {
+        display: 'flex',
+        flexDirection: 'column',
+        gap: '4px',
+    },
+});
+
+const encoder = new TextEncoder();
+
+const LARGE_JSON_EDITOR_OPTIONS: MonacoEditorType.editor.IStandaloneEditorConstructionOptions = {
```
sevoku: I see some useful usability features being disabled here, what makes total sense for large datasets, but for small/default use cases? Default results limit is 100 afair, that should not make any issues even if documents nearing 2MB limit. I think we should keep Monaco defaults and when the size warnings kicks in, reduce functionality. The warning could also say something like "Some editor features are be disabled".

sevoku: Also checking the `largeFileOptimizations: true` which kicks in at around 20Mi characters or 300lines, it already disables the same features such as:
* tokenization/highlighting
* wordWrap
* folding
* matchBrackets

Shouldn't we rely more on Monaco default behaviour here? It might be good to measure the impact of `largeFileOptimizations` vs. all the other explicit props.

bk201-: Fixed

sevoku (approved):

PR #3346, opened 2026-09-10
Fix copying and exporting a single selected query result

### LINKED ISSUE #3328: Copying or exporting one selected query result includes the entire page
## Description

In the Query Editor result table, copying or exporting exactly one selected row operates on every result in the current page instead of the selected row.

Selecting two or more rows works as expected.

## Steps to reproduce

1. Run a query that returns at least two documents.
2. Select exactly one row in the results table.
3. Copy or export the results as CSV or JSON.
4. Inspect the copied or exported content.

## Expected behavior

Only the selected row is copied or exported.

## Actual behavior

Every result on the current page is copied or exported.

## Technical details

Both of these components treat a selection as present only when more than one row is selected:

- `src/webviews/cosmosdb/QueryEditor/ResultPanel/CopyToClipboardButton.tsx`
- `src/webviews/cosmosdb/QueryEditor/ResultPanel/ExportButton.tsx`

```ts
const hasSelection = state.selectedRows.length > 1;
```

A regular table click explicitly creates a one-row selection. Because `hasSelection` remains false, the components pass `undefined` to the converters, which means all current-page rows.

A focused regression test should cover zero, one, and multiple selected rows for both copy and export.

sylvesterkaczmarek:
Fixes #3328.

Treat any non-empty row selection as a selection in both copy and export actions. Previously, selecting exactly one row passed no selection to the converters, so the whole result page was copied or exported.

Adds 18 component regressions covering zero, one, and multiple selected rows for CSV, JSON, and the registered JSON hotkey handlers. The six single-row cases fail before the fix.

Validation on macOS arm64:
- Full Vitest suite: 2,136 passed, one existing skip.
- Repository formatter, complete lint suite, and all three TypeScript build targets passed.
- No user-facing strings or public APIs changed.

Live Cosmos DB and VS Code extension-host integration tests were not run locally.

### DIFF
--- a/src/webviews/cosmosdb/QueryEditor/ResultPanel/CopyToClipboardButton.tsx
+++ b/src/webviews/cosmosdb/QueryEditor/ResultPanel/CopyToClipboardButton.tsx
@@ -18,7 +18,7 @@ export const CopyToClipboardButton = (props: ToolbarOverflowItemProps<HTMLButton
     const dispatcher = useQueryEditorDispatcher();
 
     const { ref, selectedTab, type } = props;
-    const hasSelection = state.selectedRows.length > 1; // If one document selected, it's not a selection
+    const hasSelection = state.selectedRows.length > 0;
     const tooltipClipboardContent = hasSelection
         ? l10n.t('Copy selected items to clipboard')
         : l10n.t('Copy all results from the current page to clipboard');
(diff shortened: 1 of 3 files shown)


sevoku:
Is there a way to undo the single selection? On macOS I can do that by holding Command/Alt and selecting that row again, but I'm not sure if it works the same way on Windows/Linux. An alternative would be Cmd/Alt+A, but that seems to be broken.

sylvesterkaczmarek:
Yes. The current table handler toggles a row with Command-click on macOS and Ctrl-click on Windows/Linux; Space also toggles the focused row, so a single selection can be cleared on all three platforms. Alt-click and Cmd/Ctrl+A are not wired for this.

sevoku (approved):

PR #3360, opened 2026-09-23
Add Cosmos DB Shell chat skill

mkrueger:
Register a scoped chat skill that guides Cosmos DB Shell workflows through the extension's MCP server: selecting the right integration, verifying the connected account and explicit database/container target, keeping reads bounded, and respecting confirmation gates for destructive operations.


skills/cosmosdb-shell/SKILL.md
```diff
@@ -0,0 +1,108 @@
+---
+name: cosmosdb-shell
+description: |
+  Use Cosmos DB Shell (cosmosdbshell) and its MCP tools to explore databases and
```
sevoku, line 4: Should we maybe also encourage test/validation workflows as part of DevLoop? I think it's super helpful when Copilot is able to run it's own validations using MCP on its own sample data.

